PR #2, opened 2020-09-28
refactor: remove Cache trait and module

aleksator:
Move SimpleCache as OfflineSolver to solver.rs.


mpizenberg:
I did a small doc change, but otherwise, that looks good to me!

PR #3, opened 2020-09-29
refactor: reduce Term's method visibility

aleksator:


mpizenberg:
I'm not sure how I feel about the removal of `accept_optional_version`. It is not in use but is a fundamental method for a term, it's almost the meaning of a term. But same as `bump_major` for semantic versions which is not actually used in the code base but still fundamental. I think we could even go further and let it be the only `pub` method in addition to the `Term` enum and improve the doc to better explain what `Negative(range)` is since it's not just the complement of the range, which could be done by `range.negate` and returns another `Range`.

mpizenberg:
I guess we can re-add it later if we feel like it anyway

aleksator:
I was planning to put it back today but I guess I was too slow ;) Thanks for merging!

PR #5, opened 2020-09-29
ci: split ci into two files, for push or pr

mpizenberg:


mpizenberg:
I can only see if all ci changes are ok by merging this so here we go :)

mpizenberg:
Seems like it went well!

PR #8, opened 2020-10-03
refactor: change usize to u32 in SemanticVersion and NumberVersion

aleksator:


mpizenberg:
Yes I've no real reason for using `usize` over `u32` if that's better. Could you add a link here explaining what choice to make between usize vs u32 if such a link exist?

aleksator:
I couldn't find any official guidelines other than discussions such as this:
https://users.rust-lang.org/t/i32-vs-isize-u32-vs-usize/22657
https://www.reddit.com/r/rust/comments/4lql5l/beginner_question_should_i_use_usize_and_isize/

It makes sense to me though: if your logic is not architecture dependent, it shouldn't be made so.

mpizenberg:
Awesome thanks!

PR #9, opened 2020-10-04
Implement FromStr for SemanticVersion

aleksator:
Closes https://github.com/mpizenberg/pubgrub-rs/issues/7

### DIFF
--- a/src/version.rs
+++ b/src/version.rs
@@ -5,6 +5,8 @@
 //! Traits and implementations to create and compare versions.
 
 use std::fmt::{self, Debug, Display};
+use std::str::FromStr;
+use thiserror::Error;
 
 /// Versions have a minimal version (a "0" version)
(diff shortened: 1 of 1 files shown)


aleksator:
I wonder if TryFrom<&str> should be implemented alongside. Thoughts?

mpizenberg (commented):
This looks fantastic!

Regarding the TryFrom and FromStr, here is a relevant discussion: https://internals.rust-lang.org/t/pre-rfc-deprecate-fromstr-in-favor-of-tryfrom-str/12331.
It seems FromStr is not vanishing anytime soon since it has few advantages over TryFrom, but on the other hand TryFrom seems more generally understood (probably why I started with it). So I guess having both is useful

mpizenberg (approved):
Looks good to me!

PR #13, opened 2020-10-04
feat: make the algorithm deterministic

aleksator:
This changes underlying Map implementation to BTreeMap
to guarantee solution determinism.

Closes https://github.com/mpizenberg/pubgrub-rs/issues/1, https://github.com/mpizenberg/pubgrub-rs/issues/12

### DIFF
--- a/src/internal/incompatibility.rs
+++ b/src/internal/incompatibility.rs
@@ -12,8 +12,8 @@ use crate::package::Package;
 use crate::range::Range;
 use crate::report::{DefaultStringReporter, DerivationTree, Derived, External};
 use crate::term::{self, Term};
+use crate::type_aliases::Map;
 use crate::version::Version;
-use crate::Map;
 
 /// An incompatibility is a set of terms for different packages
 /// that should never be satisfied all together.
--- a/src/internal/incompatibility.rs
+++ b/src/internal/incompatibility.rs
@@ -5,7 +5,6 @@
 //! An incompatibility is a set of terms for different packages
 //! that should never be satisfied all together.
 
-use std::collections::HashMap as Map;
 use std::collections::HashSet as Set;
 use std::fmt;
 
@@ -14,6 +13,7 @@ use crate::range::Range;
 use crate::report::{DefaultStringReporter, DerivationTree, Derived, External};
 use crate::term::{self, Term};
 use crate::version::Version;
+use crate::Map;
 
 /// An incompatibility is a set of terms for different packages
 /// that should never be satisfied all together.
@@ -67,7 +67,7 @@ pub enum Relation<P: Package, V: Version> {
 impl<P: Package, V: Version> Incompatibility<P, V> {
     /// Create the initial "not Root" incompatibility.
     pub fn not_root(id: usize, package: P, version: V) -> Self {
-        let mut package_terms = Map::with_capacity(1);
+        let mut package_terms = Map::with_capacity_and_hasher(1, Default::default());
         package_terms.insert(
             package.clone(),
             Term::Negative(Range::exact(version.clone())),
@@ -86,7 +86,7 @@ impl<P: Package, V: Version> Incompatibility<P, V> {
             Term::Positive(r) => r.clone(),
             Term::Negative(_) => panic!("No version should have a positive term"),
         };
-        let mut package_terms = Map::with_capacity(1);
+        let mut package_terms = Map::with_capacity_and_hasher(1, Default::default());
         package_terms.insert(package.clone(), term);
         Self {
             id,
@@ -100,7 +100,7 @@ impl<P: Package, V: Version> Incompatibility<P, V> {
     /// because its list of dependencies is unavailable.
     pub fn unavailable_dependencies(id: usize, package: P, version: V) -> Self {
         let range = Range::exact(version);
-        let mut package_terms = Map::with_capacity(1);
+        let mut package_terms = Map::with_capacity_and_hasher(1, Default::default());
         package_terms.insert(package.clone(), Term::Positive(range.clone()));
         Self {
             id,
@@ -126,7 +126,7 @@ impl<P: Package, V: Version> Incompatibility<P, V> {
 
     /// Build an incompatibility from a given dependency.
     fn from_dependency(id: usize, package: P, version: V, dep: (&P, &Range<V>)) -> Self {
-        let mut package_terms = Map::with_capacity(2);
+        let mut package_terms = Map::with_capacity_and_hasher(2, Default::default());
         let range1 = Range::exact(version);
         package_terms.insert(package.clone(), Term::Positive(range1.clone()));
         let (p2, range2) = dep;
@@ -163,14 +163,14 @@ impl<P: Package, V: Version> Incompatibility<P, V> {
     /// If the result is None, remove that key from the merged map,
     /// otherwise add the content of the Some(_).
     fn merge<T: Clone, F: Fn(&T, &T) -> Option<T>>(
-        hashmap_1: &Map<P, T>,
-        hashmap_2: &Map<P, T>,
+        map_1: &Map<P, T>,
+        map_2: &Map<P, T>,
         f: F,
     ) -> Map<P, T> {
-        let mut merged_map = hashmap_1.clone();
-        merged_map.reserve(hashmap_2.len());
+        let mut merged_map = map_1.clone();
+        merged_map.reserve(map_2.len());
         let mut to_delete = Vec::new();
-        for (key, val_2) in hashmap_2.iter() {
+        for (key, val_2) in map_2.iter() {
             match merged_map.get_mut(key) {
                 None => {
                     merged_map.insert(key.clone(), val_2.clone());
@@ -266,7 +266,7 @@ impl<P: Package, V: Version> Incompatibility<P, V> {
     }
 
     /// Iterate over packages.
-    pub fn iter(&self) -> std::collections::hash_map::Iter<P, Term<V>> {
+    pub fn iter(&self) -> impl Iterator<Item = (&P, &Term<V>)> {
         self.package_terms.iter()
     }
 
@@ -357,15 +357,15 @@ pub mod tests {
         ///    { p1: t1, p3: t3 }
         #[test]
         fn rule_of_resolution(t1 in term_strat(), t2 in term_strat(), t3 in term_strat()) {
-            let mut i1 = Map::new();
+            let mut i1 = Map::default();
             i1.insert("p1", t1.clone());
             i1.insert("p2", t2.negate());
 
-            let mut i2 = Map::new();
+            let mut i2 = Map::default();
             i2.insert("p2", t2.clone());
             i2.insert("p3", t3.clone());
 
-            let mut i3 = Map::new();
+            let mut i3 = Map::default();
             i3.insert("p1", t1);
             i3.insert("p3", t3);
 
--- a/src/internal/mod.rs
+++ b/src/internal/mod.rs
@@ -4,8 +4,9 @@
 
 //! Non exposed modules.
 
-pub mod assignment;
-pub mod core;
-pub mod incompatibility;
-pub mod memory;
-pub mod partial_solution;
+mod assignment;
+pub(crate) mod core;
+pub(crate) mod incompatibility;
+mod memory;
+pub(crate) mod partial_solution;
+pub(crate) mod types;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -74,9 +74,9 @@
 //! ```
 //! # use pubgrub::solver::DependencyProvider;
 //! # use pubgrub::version::SemanticVersion;
-//! # use std::collections::HashMap;
 //! # use std::error::Error;
 //! # use pubgrub::range::Range;
+//! # use pubgrub::Map;
 //! #
 //! # struct MyDependencyProvider;
 //! #
@@ -92,7 +92,7 @@
 //!         &self,
 //!         package: &String,
 //!         version: &SemanticVersion,
-//!     ) -> Result<Option<HashMap<String, Range<SemanticVersion>>>, Box<dyn Error>> {
+//!     ) -> Result<Option<Map<String, Range<SemanticVersion>>>, Box<dyn Error>> {
 //!         unimplemented!()
 //!     }
 //! }
@@ -113,7 +113,7 @@
 //!
 //! When everything goes well, the algorithm finds and returns the complete
 //! set of direct and indirect dependencies satisfying all the constraints.
-//! The packages and versions selected are returned in a `HashMap<P, V>`.
+//! The packages and versions selected are returned in a [Map<P, V>](Map).
 //! But sometimes there is no solution because dependencies are incompatible.
 //! In such cases, [resolve(...)](solver::resolve) returns a
 //! [PubGrubError::NoSolution(derivation_tree)](error::PubGrubError::NoSolution),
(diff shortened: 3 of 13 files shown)


aleksator:
I submitted a change using `BTreeMap` and a test by @Eh2406 and @mpizenberg.
I found that changing `HashSet` to `BTreeSet` at places where it is used was not needed for the test to pass.

I'm not familiar with benchmarking techniques in Rust to compare this with "fixed seed hash map" and "indexmap" other proposed solutions.
Would it be interesting to compare those? If so, any tips on a proper way to do that?

mpizenberg:
I like [criterion](https://github.com/bheisler/criterion.rs) to benchmark things, but we should wait for @Eh2406 to have code able to generate large registries first before benchmarking.

Eh2406:
I am hoping to generate a larger index we can use for benchmarking as the proptest work continues. I don't have one yet available. When setting up benchmarks in the Rust community there are 2 options. The built in benchmarking support, nightly, bare bones, and only really acceptable for longish running examples. And criterion which can be used on stable, requires a dependency and designed for micro benchmarks.  

Looks like @mpizenberg beat me to writhing this up. :-)

aleksator:
Got it, the benchmarking is blocked for now. Does the benchmarking block us for picking 0-2 options discussed previously though? Probably not, as it is more an issue of desired properties.

@mpizenberg expressed his preference for using fixed hash solution since it provides fewer guarantees. I actually don't know how preserving insertion order can be depended upon externally. From an outside perspective it should be the same as random guaranteed order of fixed hashing, shouldn't it?

BTree's sorted behavior is observable though and adds an additional `Ord` requirement to `Package` trait.

aleksator:
There are 2 implementations now in this branch: `BTreeMap` and `IndexMap` if anyone's curious to compare them.

As for using fixed hash, I haven't found any possible solutions for that.

mpizenberg:
> I actually don't know how preserving insertion order can be depended upon externally

The `add_dependencies` method of the `Solver` trait takes an iterator as input and stores the incompatibilities generated by the dependencies in a structured way in the `HashMap` of the `Memory` type. This type then is used to pick which package will be next in `PartialSolution::pick_package()`. So having the insertion order kept in the `Memory` type will keep the order of the dependencies iterator provided by the user of the `Solver` trait. That's something that could be exploited to steer the behavior of the algorithm in a certain way I think.

aleksator:
> The `add_dependencies` method of the `Solver` trait takes an iterator as input and stores the incompatibilities generated by the dependencies in a structured way in the `HashMap` of the `Memory` type. This type then is used to pick which package will be next in `PartialSolution::pick_package()`. So having the insertion order kept in the `Memory` type will keep the order of the dependencies iterator provided by the user of the `Solver` trait. That's something that could be exploited to steer the behavior of the algorithm in a certain way I think.

If you think this is a concern... I can attempt to make a solution with fixed seed hashing algorithm + HashMap.
@Eh2406 suggested `fxhash`. I found https://crates.io/crates/twox-hash which seems more recent.
Would you prefer that or maybe something else? @mpizenberg

mpizenberg:
That would be awesome. As for which hashing is more adequate, I don't know, I've never looked at this kind of things. I trust you on this

src/internal/incompatibility.rs
```diff
@@ -266,7 +266,7 @@ impl<P: Package, V: Version> Incompatibility<P, V> {
     }
 
     /// Iterate over packages.
-    pub fn iter(&self) -> std::collections::hash_map::Iter<P, Term<V>> {
+    pub fn iter(&self) -> indexmap::map::Iter<P, Term<V>> {
```
harrysarson:
```suggestion
    pub fn iter(&self) -> impl Iterator<Item=(P, Term<V>)> {
```

This way we do not commit to the Map type in the API. I dunno though if we commit to the Map type somewhere else.

harrysarson:
If I may chime in, the standard hash map allows you to [configure the hasher](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.with_hasher) which sounds like exactly what we want. We also only need to fix the hasher to be deterministic when we later iterate over the hashmap. Therefore, we do not need to change all the hashmap's in rust pubgrub; only those we iterate over.

src/internal/incompatibility.rs
aleksator: Great suggestion! I remember there were a few similar places I had to change, so I'll include those when doing a solution with another hashing algorithm.

aleksator:
> If I may chime in, the standard hash map allows you to [configure the hasher](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.with_hasher) which sounds like exactly what we want. We also only need to fix the hasher to be deterministic when we later iterate over the hashmap. Therefore, we do not need to change all the hashmap's in rust pubgrub; only those we iterate over.

Yep, looks like that's what I'll have to do in the third implementation. Do you have any suggestions for the particular hashing functions? Currently I'm considering `twox-hash` and `fxhash`, but I don't know if I should prefer one over the other.
I'll probably boil down to the benchmarking of various options.

harrysarson:
> Currently I'm considering twox-hash and fxhash, but I don't know if I should prefer one over the other.

Ah sorry! I assumed twox-hash and fxhash were hash map implementations! I now understand you; they are hash _function_ implementations! I have no knowledge about which is better.

harrysarson:
This may be a crazy and infeasible idea but...

Maybe the types in rust pubgrub could be generic over a `S:  BuildHasher` parameter. Then users can pick the best option for them. We could always pick a safe (and I guess deterministic) default.

aleksator:
> Ah sorry! I assumed twox-hash and fxhash were hash map implementations! I now understand you; they are hash function implementations! I have no knowledge about which is better.

Actually `fxhash` provides a convenience of creating HashMap (from std) with hash function from the crate and exports that as `fxhash::FxHashMap`, so you can say it's both. Haven't looked if `twox-hash` does that yet, but it would make sense to do that as well.
We can of course use `HashMap::with_hasher()` directly if we don't want to publicly expose types such as `FxHashMap`.

mpizenberg:
> Maybe the types in rust pubgrub could be generic over a `S: BuildHasher` parameter.

That sounds like a possible extension `run_with_hasher::<hasher_here>(...)` or whatever actually correct way of saying this XD. Some people may be very serious about using cryptographic hashes or don't care about reproducibility. But maybe later I think, let's first use a good default.

aleksator:
I've added implementations with 2 custom hashers: `fxhash` and `twox-hash`.
I expected `twox-hash` to be faster on the basis of it being based on an established algorithm and being maintained, but it wasn't the case.

| solution | performance |
|-|-|
| BTreeMap |  38,800,753 ns/iter (+/- 2,824,742) |
| fxhash | 87,965,741 ns/iter (+/- 8,027,249) |
| HashMap  | 90,629,276 ns/iter (+/- 77,732,588) |
| IndexMap | 147,237,421 ns/iter (+/- 3,569,217) |
| twox-hash | 193,003,848 ns/iter (+/- 4,558,645) |

HashMap here is the default one (non-deterministic).

I myself prefer `BTreeMap` on the basis of it being available in std and being the easiest to use (especially compared to custom hashers).

Regarding performance numbers, we discussed them briefly with @Eh2406 and he made a point they may be mute: what makes or breaks the algorithm is what package it picks first for processing, and hashing may just be unlucky in providing it with a non-optimal first package.

I decided to post results anyway; the work had already been done after all.

mpizenberg:
Did you do the hashing bench with "real world" types to be hashed like string? Because the example provided by @Eh2406 uses numbers as package identifiers and the profile of hashing these VS strings for example might be very different I guess?

aleksator:
To reproduce the benchmarks, checkout commit with whatever implementation interests you and run
```
cargo +nightly bench --features=serde
```

> Did you do the hashing bench with "real world" types of hashes like string? Because the example provided by @Eh2406 uses numbers as package identifiers and the profile of hashing these VS strings for example might be very different I guess?

I cherry picked his benchmark into this branch and run that. Regarding strings, it's a valid point, but
> what makes or breaks the algorithm is what package it picks first for processing

so it doesn't matter until we implement picking heuristic anyway.

mpizenberg:
Ok so I propose this:
Do you think of a valid type of package ids, that we could want to use, and that do not implement `Ord`?
- If yes, we choose fx or twox and call it a day before other optimizations land, then revise
- If not, we choose BTreeMap and call it a day before other optimizations land, then revise

aleksator:
Hm, I don't think I'm the best person to answer this actually. Only floating point numbers come to my mind from `std` (because of `NaN` fiesta).
Could someone want to use those as package identifiers? In the context of package managers certainly not. In the context of using the library to solve more general problem? Maybe. I can't imagine one at the moment. 

@Eh2406 Do you have an opinion on having an `Ord` bound? Do you maybe have a preference for any of the implementations?

Eh2406:
The Trie used by [Cargo](https://github.com/rust-lang/cargo/blob/master/src/cargo/core/resolver/conflict_cache.rs) for storing `conflicts` (a less powerful equivalent of ` incompatibilities`) relies on an Ord bound. So it may be convenient to have one laying around. However the Trie can be constructed by using the order we first see a name inserted, so it is not critical.

I can't think of a package ids that do not impl ord. So let's go with BTree for now. Wen we are sure the algorithm is correct, and we have a user that complains we can rethink it.

mpizenberg:
Alright, let's go with BTree :+1:

aleksator:
Guys, I managed to solve the major gripe I had with using a custom hash function: having to add a third type parameter everywhere and proliferation of that function in type signatures.

If we define our Map type at crate level like this:
```
pub type Map<K, V> = HashMap<K, V, BuildHasherDefault<FxHasher64>>;
```
we can continue using `Map<K, V>` everywhere.
It also allows us to replace the said function easily in one place! It shouldn't even require any changes from the library users if we do so (right?).

The only downside is you can't use
`Map::with_capacity(1);` anymore, you have to do this:
`Map::with_capacity_and_hasher(1, Default::default());`
This is a bit annoying still because it's verbose and these methods would have to be changed back to `Map::with_capacity(1);` if we are to do the change to BTreeMap, for example.
But anyway, this is still relatively minor compared to what we had before with my initial implementation that used custom hashers  (compare with `feat: export Map type` commit).

As a result, I'm now on board with using a custom hash function for the reasons @mpizenberg originally pushed for it if you agree to exporting such type.

aleksator:
> It also allows us to replace the said function easily in one place!

To demonstrate, a commit that changes the function to `rustc-hash` (which is a spiritual? successor to `fxhash` currently used by rustc) took me a couple lines of code.

Here's the benchmark run:
`test large_case ... bench:  91,197,228 ns/iter (+/- 6,310,492)`

aleksator:
I've rebased and squashed the branch, so that we could pull the trigger on this at any moment ;)

All other implementations are available for checking and benchmarking in `deterministic_ordering_multiple_implementations` branch.

src/internal/incompatibility.rs
aleksator: I've applied the change, thank you again. The other one I wanted to do it was
`type IntoIter = std::collections::hash_map::IntoIter<P, Term<V>>;`
Doing that in `type` definitions turned out to be unstable :(

src/lib.rs
```diff
@@ -189,6 +189,9 @@
 
 #![warn(missing_docs)]
 
+/// Map implementation used by the library
+pub type Map<K, V> = rustc_hash::FxHashMap<K, V>;
```
mpizenberg: This may be subtle, but since I've been coding in Elm, I always avoid dependency cycles in modules. Putting the type alias here in `lib.rs` and importing it in submodules create a dependency cycle. Rust is smart enough to solve it but if possible I'd like to avoid it.

Eh2406: could we make a `mod type_aliases` that has this one line, and is imported by all the places that use it?

mpizenberg: that would be my preference indeed

aleksator: I thought of putting it in a module but for the reasons of `lib.rs` containing only module declarations and looking clean. I didn't think of the dependency cycle.

Would you like it being exported from the root as it is now? I think `pubgrub::Map` would be more convenient for users of the library than `pubgrub::type_aliases::Map`.
If we export, the module shouldn't be made `pub` and instead placed in `internal` I think.

And about the naming:
1) `type_aliases`
2) `types`
3) `exports`
4) something else?

aleksator: I moved it to `internal::types` and exported that from root. Naming is subject for discussion. Please take a look.

src/lib.rs
```diff
@@ -189,8 +189,7 @@
 
 #![warn(missing_docs)]
 
-/// Map implementation used by the library
-pub type Map<K, V> = rustc_hash::FxHashMap<K, V>;
+pub use internal::types::Map;
```
mpizenberg: Actually I think you should change the `use` in other in modules that expose the `Map` type in their type signature, and just put the `types.rs` module as a `pub` one in `lib.rs`.

aleksator:
By the way, does `squash merge` perform rebase as well? Looks like #23 might be coming faster, so I wonder if I should rebase manually after that to keep the history linear.

src/internal/mod.rs
```diff
@@ -4,8 +4,9 @@
 
 //! Non exposed modules.
 
-pub mod assignment;
-pub mod core;
-pub mod incompatibility;
-pub mod memory;
-pub mod partial_solution;
+mod assignment;
```
mpizenberg: no need to change anything here, those are effectively already `pub(crate)` since `internal` module is `mod` and not `pub mod`

mpizenberg: If we are to make code reorganization at some point, let leave it for another PR.

mpizenberg:
> By the way, does `squash merge` perform rebase as well? Looks like #23 might be coming faster, so I wonder if I should rebase manually after that to keep the history linear.

Squash merge put all your changes into one diff, and see if it can apply that diff on the other branch. It's not really a rebase but it's linear on `dev` just like a rebase, except there is only 1 commit

mpizenberg:
type_aliases might be more adequate than types but I don't think it matter too much

src/lib.rs
aleksator: Does it mean you prefer users of the library to import
`pubgrub::types::Map` instead of `pubgrub::Map`?

mpizenberg: For now yes. They already import all the rest from their respective modules, so let's do it like that for the time being. Once we have more usage, we will see what things are the most often use, and we can make a `prelude` module that reexport those for a one stop shop `use pubgrub::prelude::*` or `prelude::{...}`

aleksator:
> Squash merge put all your changes into one diff, and see if it can apply that diff on the other branch. It's not really a rebase but it's linear on dev just like a rebase, except there is only 1 commit

It might have created merge commit in case the branch wasn't already on top, that's why I was asking. Thank you for the clarification.

aleksator:
Map is now in `type_aliases` and is publicly accessible as `pubgrub::type_aliases::Map`.

src/lib.rs
aleksator: Got it, makes sense to keep it consistent. Even `resolve` function is not exported from `solver` module after all, and _that_ one is supposed to be the easiest to find.

mpizenberg (commented):
All good for me! I'll let @Eh2406 squash merge if they(he?/she?) also approve those changes

aleksator:
> All good for me! I'll let @Eh2406 squash merge if they(he?/she?) also approve those changes

Jacob prefers "he" ;)

Eh2406:
"He", but thanks for asking! The PR looks great!

mpizenberg:
> "He", but thanks for asking!

alright, same for me

PR #14, opened 2020-10-05
Benchmark

Eh2406:
This adds a large case as a benchmark. It was generated by a proptest `solution.len() < 50`, disabling shrinking, and converting by hand. I do want to have easier ways to go between proptest<->example tests<->benchmarks. But prioritized this as it sounds like we are waiting on a benchmark for #13.

This benchmark takes over 1.5sec per iteration on my laptop. That is much too slow to run regularly. I could make the example smaller, or pubgrub faster. So I did some profiling and a lot of the time was spent cloning `incompatibilities`, The last commit adds an `Rc` to make that clone cheaper. Bringing the benchmark to .4sec on my laptop. (I am sure there are other solutions, this one had a small diff.) 

As criterion likes microbenchmarks, I left this with nightly benchmarking for now. It can be run with `cargo +nightly bench --features=serde` and takes ~2 min to run on my laptop.


Eh2406:
Pushed another optimization that brings the runtime to .2sec on my laptop.

mpizenberg:
For the cloning of the incompatibility inside the unit propagation, that's actually something that we should think of algorithm-wise. I believe I did that because inside the unit propagation loop, we may run into a conflict and run `self.conflict_resolution()` which mutate `self.incompatibilities`. But there might be a better way to say that we want to iterate over the incompatibilities existing at the beginning of unit propagation. Maybe using an [immutable data structure](https://github.com/bodil/im-rs) is actually the right choice? don't know. Just my thought on this

mpizenberg:
By the way, that's super awesome!

mpizenberg:
Also another remark regarding benchmarking of this thing. The reporter is quite poorly written, performance wise, so it's wise to not include the reporting part in the benchmarks of the solver part. That will need it's dedicated thing.

mpizenberg:
Also, the vec of all incompatibilities grows quite a lot because I don't do any of the regrouping tactics mentioned in the doc comments of `Incompatibility::merge_into()`.

One thing that is doable with the current design where all incompatibilities are accumulated in the `incompatibility_store` anyway (which is important for error reporting) would be to collapse the `NoVersion` incompatibilities as I do in the reporting code. That makes it so that `{ a, 1 <= b < 2}` union `{2<= b < 3}` becomes `{a, 1 <= b < 3}` and we could remove the old from the current `incompatibilities`.

The other doable thing is as pub does is collapsing the depending part : `{ a1, not b}` union `{a2, not b}` would be collapsed into `{a1 union a2, not b}` and the old ones removed from the current incompatibilities.

One annoying thing here is that removing things in a vec is not super optimized, but it might help prevent that vec growing too much. I guess it depends how "normal" is the package registry used for that test.

mpizenberg:
But as you suggested, more complex improvements are better suited for another time, specifically on performance improvement.

Related to this PR though, I'm wondering, is there a way to prevent huge files from getting into git history? I mean a practical one. There is the git submodule option that moves the size problem on the submodule, which helps a bit. I've heard of git lfs but never used it. Any other option?

Eh2406:
> cloning of the incompatibility

Yes, redesigning the loop so that we push to incompatibility after the loop is one of the "other solutions" I had in mind. Switching to a more dedicated immutable data structure is another. 

Once we have added the regrouping tactics, I would not be surprised if we can do algorithmically better then a `Vec<Map<...>>`, we have a kind of Trie for doing something similar in [Cargo](https://github.com/rust-lang/cargo/blob/master/src/cargo/core/resolver/conflict_cache.rs). But I don't want to make big algorithm changes in this PR.

About large files. Before this merges I should squash the benchmark commits. And I can try to instrument the solver to find out which deps are used. One question is how small do you want me to keep the files? Git starts complaining at 50MB and this is still at 0.2MB.
If now or in the future we have files to big. We can set up a separate repo that has the benchmarks. Or we can use LFS, I have not used it yet myself. Yes I did just restate your suggestions. I was hoping that as I typed them out another suggestion would come to mind, but it didn't.

Eh2406:
I experimented with instrumenting the solver. I was able to remove some 1300 lines without affecting the runtime of the benchmark.

It is very clear that there is a lot of run-to-run variance in the runtime. Based on the instrumentation, it is do to the nondeterminism cc #13.

mpizenberg:
> Yes, redesigning the loop so that we push to incompatibility after the loop is one of the "other solutions" I had in mind. Switching to a more dedicated immutable data structure is another.

push incompatibilities after the loop brings a different behavior though for conflict resolution so I don't think that's correct. Conflict resolution should include up to date incompatibilities (except if those new ones don't have an impact but hum ... not sure)

mpizenberg:
> About large files

The issue with having this in the git tree is that for example in the PR where `run` is extracted from the `Solver` trait that becomes a `DependencyRetriever`, it will change all the lines of this file.

If we go with just changing the hashing function on just the hasmaps on which we iterate on the other PR for reproducibility, it should be fine and we should not rush this one I think.

What would be good, whether or not this use lfs / submodule / plain git, is having the data just represented as data. This benchmark could load the data from the corresponding file and generate the registry from it. What do you think of this? Do you have an idea of how to represent that data (json, toml, other) if you think that's a good idea?

aleksator:
> Do you have an idea of how to represent that data (json, toml, other) if you think that's a good idea?

That's a good idea that would need quite some work to implement.
It would require adding support for specifying dependencies in text format thus bringing https://lib.rs/crates/semver-parser, and creating mapping from their structs to ours.
Which is great and should probably happen at some point, but I'm not sure if we would want to block this PR on that feature.

aleksator:
As for the specific format... We could use relevant parts of Cargo.toml, which could make integration later easier.

We can also use something else and leave implementation of that layer for the Cargo itself.
I heard about RON (Rust Object Notation), which sounds like was created for this purpose. We could investigate if that would be a proper format.

mpizenberg:
> 'm not sure if we would want to block this PR on that feature.

In that case, we should wait for PR #15 (extraction of `run()`) which should be pretty soon. If that's ok with @Eh2406 too, let's go for merge as soon as that other pr is done. We can refine that later.

Eh2406:
I do like the idea of having the outputs as plane data files. It is part of what I had in mind with:
> I do want to have easier ways to go between proptest<->example tests<->benchmarks. 

Specifically a folder of plane data files can easily become:
- a test per file "for each package gets the same answer as an SAT solver"
- a benchmark per file "how fast can it resolve all the packages"

and a proptest can easily dump its data structure to a file when it finds a problem. And it may create a format for bug reports. And... Yes it is a good Idea. 
I wanted to get this out for discussion before doing all that work. Let's let #15 and mabe #13 land and re assess. Maybe I will have an inpl to show!

aleksator:
>  a test per file "for each package gets the same answer as an SAT solver"
    a benchmark per file "how fast can it resolve all the packages"

Good plan!
>Let's let #15 and mabe #13 land and re assess.

Sure, #15 is kinda independent of everything but is nice to merge first to avoid merge conflicts.
I intend to use the benchmark for #13 without committing it to help us pick the implementation. Even if we choose not to use, say, `BTreeMa`p since it introduces ordering, I am still curious how it performs compared to other options.

> Maybe I will have an inpl to show!

In that case there would be no need to commit this benchmark file, so let's wait and see.

Eh2406:
I have reorganized the history.
1. In the first commit the smallest version of the benchmark code. (Keeped in the history, as it can be used to regenerate the ron file if we change the encoding.)
2. In the second commit add serde and move the benchmark to a ron file. As suggested above. This also involved restructuring `OfflineSolver` to get the serialization I wanted.
3. In the third commit the incompatibilities optimization.
4. In the fourth commit the `all_terms` optimization.

I will rebase onto #15 when it lands. Then I am ready for merge. There are a lot of ways to improve this, but let's save it for other PR's.

aleksator:
Is the change in `OfflineSolver` to use `BTreeMap` intentional?

Eh2406:
Yes, but it is negotiable. I find it very helpful to have the packages in an example well organized. serde was serializing in the the `iter` order. Witch made the ron file unnavigable. If we don't want that change, then we are going to need some way to sort the data before putting it in a ron file. What would you prefer?

aleksator:
Yes, but it is negotiable. I find it very helpful to have the packages in an example well organized. serde was serializing in the the iter order. Witch made the ron file unnavigable. If we don't want that change, then we are going to need some way to sort the data before putting it in a ron file. What would you prefer?

Makes sense. I personally don't mind having a `BTreeMap` (that's the first implementation I did in #13).

aleksator:
I've added [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) labels to commit messages (no changes to the code).

aleksator:
By the way, whatever we end up doing in #13 will solve the ordering problem for RON files as well.

Eh2406:
I made one small change to the ron format. I will try to use that commit convention going forward.

Eh2406:
I rebased. If we are all on board, this is ready for a rebase merge.

mpizenberg:
This PR lands a BTree in the offline dependency provider, adding an `Ord` bound. Alex, when working on the reproducibility PR, could you check that there are no more `Ord` bounds around since we'll have it on the `Package` trait? Would you mind also checking that other changes in the offline dependency provider here are still coherent with your cached example?

Other than that, this is really great!

mpizenberg (approved):

Eh2406:
With your approval, I merged. We can clean Ord bounds in #13.

PR #15, opened 2020-10-05
feat: extract run() out of Solver trait

aleksator:
To prevent run() function being overridden in Solver impls
it was extracted from the Solver trait.
run() was renamed to resolve(), and Solver became DependencyProvider
as dependency fetching was separated from the algorithm execution.

Closes https://github.com/mpizenberg/pubgrub-rs/issues/11.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,13 +41,13 @@ For semantic versions, `bump` corresponds to an increment of the patch number.
 ## API
 
 ```rust
-solution = solver.run(package, version)?;
+let solution = resolve(&dependency_provider, package, version)?;
 ```
 
-Where `solver` provides the list of available packages and versions,
+Where `dependency_provider` supplies the list of available packages and versions,
 as well as the dependencies of every available package
-by implementing the `Solver` trait.
-The call to `run` for a given package at a given version
+by implementing the `DependencyProvider` trait.
+The call to `resolve` for a given package at a given version
 will compute the set of packages and versions needed
 to satisfy the dependencies of that package and version pair.
 If there is no solution, the reason will be provided as clear as possible.
--- a/examples/branching_error_reporting.rs
+++ b/examples/branching_error_reporting.rs
@@ -1,21 +1,21 @@
 use pubgrub::error::PubGrubError;
 use pubgrub::range::Range;
 use pubgrub::report::{DefaultStringReporter, Reporter};
-use pubgrub::solver::{OfflineSolver, Solver};
+use pubgrub::solver::{resolve, OfflineDependencyProvider};
 use pubgrub::version::SemanticVersion;
 
 // https://github.com/dart-lang/pub/blob/master/doc/solver.md#branching-error-reporting
 fn main() {
-    let mut solver = OfflineSolver::<&str, SemanticVersion>::new();
+    let mut dependency_provider = OfflineDependencyProvider::<&str, SemanticVersion>::new();
     #[rustfmt::skip]
     // root 1.0.0 depends on foo ^1.0.0
-    solver.add_dependencies(
+        dependency_provider.add_dependencies(
         "root", (1, 0, 0),
         vec![("foo", Range::between((1, 0, 0), (2, 0, 0)))],
     );
     #[rustfmt::skip]
     // foo 1.0.0 depends on a ^1.0.0 and b ^1.0.0
-    solver.add_dependencies(
+        dependency_provider.add_dependencies(
         "foo", (1, 0, 0),
         vec![
             ("a", Range::between((1, 0, 0), (2, 0, 0))),
@@ -24,7 +24,7 @@ fn main() {
     );
     #[rustfmt::skip]
     // foo 1.1.0 depends on x ^1.0.0 and y ^1.0.0
-    solver.add_dependencies(
+        dependency_provider.add_dependencies(
         "foo", (1, 1, 0),
         vec![
             ("x", Range::between((1, 0, 0), (2, 0, 0))),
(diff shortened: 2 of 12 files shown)


aleksator:
The major work was updating all the documentation. I also made doc tests runnable (removed `ignore`) to verify they won't go out of sync when doing such changes.

While changing all the `Solver` references to `DependencyProvider`, I made those clickable in the generated documentation and point to the trait in question.
I'd like to do this for other documentation items as well, but it was out of scope for this PR.

aleksator:
@Eh2406 @mpizenberg 
Feel free to bikeshed the naming by the way :wink:

Eh2406:
Gave it a quick skim and it looks good. :-)

aleksator:
DependencySupplier is another naming to consider.

mpizenberg:
I like `DependencyProvider`, this looks good! especially it's nice to have doc example compile. One thing I'm not so certain about is the change from `&mut` to `&` for the required methods. If the user of the resolve function want to be able to update their cache for their registry, they need to either have a `&mut ref` cache in the method arguments, or have `DependencyProvider` itself being able to mutate its internal cache. The latter seems cleaner than to have a cache that we do not use, or that we could use but adds constraints about how to use a cache where really it's not our concern.

aleksator:
_That's_ what it's for! I thought it was just an oversight.

The implementors can actually store a cache separately though. DependencyProvider would update the mut cache struct (on their side) and query that before doing an http call.

I can make an example to demonstrate if you'd like.

mpizenberg:
The implement can store a cache separatly, but I don't see how they can know how to modify it (except with a closure maybe?) if since they don't know when those functions are called by the runner? I'm glad if you can provide an example. You can look at this code for a concrete example of what I'm saying: https://github.com/mpizenberg/elm-test-rs/blob/31596f874a6e9340350de3e7c783a5cde9ca396b/src/solver.rs#L407

mpizenberg:
This is probably more relevant line for an example of how I update the cache while running this: https://github.com/mpizenberg/elm-test-rs/blob/31596f874a6e9340350de3e7c783a5cde9ca396b/src/solver.rs#L474

aleksator:
@mpizenberg Here's a relevant part of documentation. I believe we are hitting one of the exact intended usages of RefCell:
> Occasionally it may be desirable not to expose in an API that there is mutation happening "under the hood". This may be because logically the operation is immutable, but e.g., caching forces the implementation to perform mutation

https://doc.rust-lang.org/std/cell/#implementation-details-of-logically-immutable-methods

Eh2406:
This may be a different issue, but is there a way to avoid the clones and allocation in https://github.com/mpizenberg/elm-test-rs/blob/31596f874a6e9340350de3e7c783a5cde9ca396b/src/solver.rs#L439? In Cargo we use `Rc<Vec<Summary>>` https://github.com/rust-lang/cargo/blob/07162dbaa84aaad97d6f3be5069265e784179bcd/src/cargo/core/resolver/dep_cache.rs#L35, so that a cache hit is  and Rc clone instead of a Vec clone. I wonder if there is a trate bound that works for the algorithm and allows returning `Vec<>`, `Rc<Vec<_>>`, `im::Vector<_>`, or even a leaked `&'static [_]`?

aleksator:
> This may be a different issue, but is there a way to avoid the clones and allocation in https://github.com/mpizenberg/elm-test-rs/blob/31596f874a6e9340350de3e7c783a5cde9ca396b/src/solver.rs#L439? In Cargo we use `Rc<Vec<Summary>>` https://github.com/rust-lang/cargo/blob/07162dbaa84aaad97d6f3be5069265e784179bcd/src/cargo/core/resolver/dep_cache.rs#L35, so that a cache hit is and Rc clone instead of a Vec clone. I wonder if there is a trate bound that works for the algorithm and allows returning `Vec<>`, `Rc<Vec<_>>`, `im::Vector<_>`, or even a leaked `&'static [_]`?

Sounds like it's worth investigating. I wouldn't want to do such changes in this PR though to keep it atomic. Could you please create a new issue?

aleksator:
I've added a `CachingDependencyProvider` as an example how it could be implemented when we accept immutable &self in our API.
I didn't come up with a dependency tree that would actually query the same package twice; feel free to suggest one if you have it in mind.

@mpizenberg 
1. Are you on board with changing `&mut self` to `&self` in our API?
2. Would you like this `CachingDependencyProvider` implementation exposed alongside with `OfflineDependencyProvider` or would you prefer it to be kept in an example code?

mpizenberg:
> I've added a `CachingDependencyProvider` as an example how it could be implemented when we accept immutable &self in our API.

Thanks for the example! I didn't know this was possible without going the unsafe road. One thing about the example if it is to be useful for something else than showcasing interior mutability, is that the call to `list_available_versions` should cache the retrieved versions. For that I suppose the `OfflineDependencyProvider` must expose it's function to acknowledge existence of versions. Otherwise, one call to `get_dependencies` will cache one version of a given package. And the next time the list of all packages existing is asked, it will go the wrong branch because one version was added to cache.

> Would you like this CachingDependencyProvider implementation exposed alongside with OfflineDependencyProvider or would you prefer it to be kept in an example code?

Example is better I think.

Other than that, if @Eh2406 is also on board with the removal of `mut` let's go for it :)

Eh2406:
We can always improve or move the example.

I think it is a close call on the `mut` issue. On one hand it is great to show that run will not affect the DependencyProvider. On the other hand it is odd to have the intended usage to use refcell. I am leaning toward removing the `mut`. So let's go for it! 🚀

mpizenberg:
Yes, we should mention the example in the documentation though. I consider myself not a beginner and was not aware of that functionality (maybe because the functional programmer in me try to avoid anything related to mutations XD)

aleksator:
> Yes, we should mention the example in the documentation though. I consider myself not a beginner and was not aware of that functionality (maybe because the functional programmer in me try to avoid anything related to mutations XD)

Don't worry, I myself was just aware that it was possible _in theory_, but had to actually learn _how to_ for creating this example :)

> We can always improve or move the example.

We can, but there's a bug in the example is pointed out by Matthieu:
> one call to get_dependencies will cache one version of a given package. And the next time the list of all packages existing is asked, it will go the wrong branch because one version was added to cache.

I should fix that before we can merge it I think. The question is how to detect that `list_all_versions()` was called and not just 1 dependency was added by `get_dependencies()`. Maybe I can save packages for which `list_all_versions()` was called into a HashSet .

Another alternative would be to get rid of caching at all in `list_all_versions()` to keep the example simple. Since we are not exposing this provider as an actual type, it might be preferable.

mpizenberg:
> The question is how to detect that `list_all_versions()` was called and not just 1 dependency was added by `get_dependencies()`.

That's not an issue, we only ask for dependencies of a version we've picked, meaning we've asked available versions first.

mpizenberg:
> Another alternative would be to get rid of caching at all in `list_all_versions()` to keep the example simple. Since we are not exposing this provider as an actual type, it might be preferable.

or that yeah. Possible too

aleksator:
I've removed caching in `list_all_versions()`.
Please let me squash the last 2 commits before merging (if everything else is okay of course).

Eh2406:
Looks good to me! I think https://github.com/mpizenberg/pubgrub-rs/blob/b47781d71e55ca0cb0229dd68e93da467f87d04b/src/lib.rs#L93 may be a good paragraph to menchen the example and refcel. What do you think?

aleksator:
> Looks good to me! I think
> 
> https://github.com/mpizenberg/pubgrub-rs/blob/b47781d71e55ca0cb0229dd68e93da467f87d04b/src/lib.rs#L93
> may be a good paragraph to menchen the example and refcel. What do you think?

Do you know if it's possible to link to example in rustdoc?

Eh2406:
I don't know.

aleksator:
Probably not, since documentation isn't generated for examples. The best I could do was link to a file like this:
`[CachingDependencyProvider](../../../examples/caching_dependency_provider.rs)`

I'm not sure hacking directory structure like this will work when documentation is generated during publishing and if it's worth doing in the first place.

Eh2406:
That looks great! When you are happy with your history, lets merge.

aleksator:
Cleaned up a history a bit to leave only 2 commits (change itself + adding an example) that I didn't want to squash.

I'll let Matthieu take a final look before merging.

mpizenberg:
This is great!

Regarding history, I thought I'd squash-merge all PRs into dev so the only thing left from your commit history will be the description when included in the squashed commit. Do you want me to rebase-merge instead to keep the two commit distinct?

aleksator:
I would prefer `rebase merge`: I tried to keep the commits atomic and wrote a description for each.
Although I couldn't resist fixing a typo and removing `unwrap` in the last one. Sigh... On the one hand atomicity is great, on the other, having a "typo fix" and "remove unwrap" commits as separate just creates noise in the history. How can I eat my cake and have it too? :cake:  /EndOfRant

aleksator:
Thank you for merging! :+1:

PR #23, opened 2020-10-08
Fix prior cause computation

mpizenberg:
PubGrub documentation describes the first step of prior cause computation as ["the union of the terms in incompatibility and the terms in satisfier's cause"](https://github.com/dart-lang/pub/blob/master/doc/solver.md#conflict-resolution). As a result, I was aggregating terms of both incompatibilities, and for packages in common, using the union of the terms.

Instead, one should compute the intersection of terms of packages in common as done here:
https://github.com/dart-lang/pub/blob/23a59a380cdc88aca6a50f9b0147ecdd3b948919/lib/src/solver/incompatibility.dart#L68

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -138,7 +138,11 @@ impl<P: Package, V: Version> State<P, V> {
                         );
                         return Ok(current_incompat);
                     }
-                    Derivation { cause, .. } => {
+                    Derivation {
+                        cause,
+                        term,
+                        package,
+                    } => {
                         if previous_satisfier_level != satisfier_level {
                             self.backtrack(
                                 current_incompat.clone(),
@@ -148,8 +152,13 @@ impl<P: Package, V: Version> State<P, V> {
                             return Ok(current_incompat);
                         } else {
                             let id = self.incompatibility_store.len();
-                            let prior_cause =
-                                Incompatibility::prior_cause(id, &current_incompat, &cause);
+                            let prior_cause = Incompatibility::prior_cause(
+                                id,
+                                &current_incompat,
+                                &cause,
+                                package,
+                                term,
+                            );
                             // eprintln!("\ncause 1: {}", &current_incompat);
                             // eprintln!("cause 2: {}", &cause);
                             // eprintln!("prior cause: {}\n", &prior_cause);
--- a/src/internal/incompatibility.rs
+++ b/src/internal/incompatibility.rs
@@ -138,22 +138,9 @@ impl<P: Package, V: Version> Incompatibility<P, V> {
         }
     }
 
-    /// Perform the union of two incompatibilities.
-    /// Terms that are always satisfied are removed from the union.
-    fn union(id: usize, i1: &Map<P, Term<V>>, i2: &Map<P, Term<V>>, kind: Kind<P, V>) -> Self {
-        let package_terms = Self::merge(i1, i2, |t1, t2| {
-            let term_union = t1.union(t2);
-            if term_union == Term::any() {
-                None
-            } else {
-                Some(term_union)
-            }
-        });
-        Self {
-            id,
-            package_terms,
-            kind,
-        }
+    /// Perform the intersection of terms in two incompatibilities.
+    fn intersection(i1: &Map<P, Term<V>>, i2: &Map<P, Term<V>>) -> Map<P, Term<V>> {
+        Self::merge(i1, i2, |t1, t2| Some(t1.intersection(t2)))
     }
 
     /// Merge two hash maps.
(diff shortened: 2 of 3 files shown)


aleksator:
Might be worth submitting a PR to fix the original algorithm description. If it's not accepted, we can fork that file and fix the description as we go.

I remember you hit another bug with the guide (https://github.com/dart-lang/pub/issues/2538) and that one still isn't commented on, so there are already at least 2.

Eh2406:
For previous discussion see: https://rust-lang.zulipchat.com/#narrow/stream/260232-t-cargo.2FPubGrub/topic/There.20is.20a.20bug!.20https.3A.2F.2Fgithub.2Ecom.2Fmpizenberg.2Fpubgrub-rs.2F.2E.2E.2E

I can't say as I understand how this works or why it is justified by the description. But it passes the existing tests, and the proptests that found the original case. If there is a way to replace the `rule_of_resolution` it would be nice, but overall I am +1.

mpizenberg:
> I remember you hit another bug with the guide ([dart-lang/pub#2538](https://github.com/dart-lang/pub/issues/2538)) and that one still isn't commented on, so there are already at least 2.

nah that bug was an indirect consequence of me doing that same mistake in the elm port

aleksator:
I would like to find a test case that fails with determinism PR and without this patch. Otherwise we don't have a way to detect this bug if it ever reappears in the future, since the current test always passes with the current determinism approach.
Theoretically both 1) this bug reappearing 2) changes to determinism could result in the test that's in the branch being able to catch it, but it's unlikely.

I could start trying to find such test case tomorrow.
Other than that, I am happy with the fix and we can merge if we think it's not worth spending time on finding the test case.

mpizenberg:
> Might be worth submitting a PR to fix the original algorithm description

Yeah the description is not wrong, we are doing the union of terms in both incompatibilities. But it does not say how to deal with terms in both incompatibilities for the same package (those have to be intersected).

mpizenberg:
> I would like to find a test case that fails with determinism PR and without this patch

That would require being able to fuzz-generate "compatible" registries by construction, and asserting that the result of `resolve` is `Ok(...)`

Eh2406:
I can't think of a way to find a test case. And even if we do any change, a new heuristic or different hashing, will brake it again.
Eventually I will port over the test that compares `resolver.is_ok() == SATlibrary.is_ok()`, so I dont think it is worth spending time on now.

aleksator:
+1 on merging then. I agree that `rule_of_resolution()` was kind of nice though.

mpizenberg:
> +1 on merging then. I agree that `rule_of_resolution()` was kind of nice though.

I'll see if I can re-add it

aleksator (approved):

Eh2406:
I ran the benchmark, and the difference from this PR was not statistically significant. :-)

mpizenberg:
Just squashed the 3 commits together and fixed merge conflict

Eh2406:
Looks good to me! @aleksator want to do the honours?

mpizenberg:
Reflecting a bit on this, I believe one way of framing prior cause in a general setting is the following.

----------------------

Let an incompatibility be a set of package terms that must not be all satisfied together. Saying that versions in the range `r1` of package `a` are incompatible with package `b` in the range `r2` can be expressed by the following association of package terms:

```txt
{a: t1, b: t2}
```

where `t1` is the term `Positive(r1)` and `t2` is the term `Positive(r2)`. Similarly, using `{a: Positive(r1), b: Negative(r2)}` would mean instead that versions in range `r1` of package `a` depend on a version in range `r2` in package `b`. So a dependency can be expressed as an incompatibility.

Of course, incompatibilities can contain more than two package terms. The incompatibility `{a: ta, b: tb, c: tc, ...}` means that the packages described here with their associated terms are incompatible as a whole, they cannot all be satisfied at the same time.

More generally, we could note incompatibilities as sets of the form `{ (p_i, t_i) | p_i ∈ P }` where `P` is the set of all available packages, `t_i` is the term associated with package `p_i` in that incompatibility. Imagine that `P` is `{a, b, c}`, then all incompatibilities can be expressed in the form `{ a: ta, b: tb, c: tc }` where I use the colon notation `:` instead of pairs just for readability.

Indeed even `{a: ta, b: tb}` is strictly equivalent to `{a: ta, b: tb, c: Negative(∅)}` where `∅` is the empty range, because `Negative(∅)` is always satisfied. It is satisfied if no version of `c` is picked or if any version of `c` is picked.

In that context, for any `p_j ∈ P`, I think that prior cause computation can be expressed as:

```txt
    { (p_j, t1_j) } ∪ { (p_i, t1_i) | p_i ∈ P, p_i ≠ p_j }
and { (p_j, t2_j) } ∪ { (p_i, t2_i) | p_i ∈ P, p_i ≠ p_j }

implies

    { (p_j, t1_j ∪ t2_j) } ∪ { (p_i, t1_i ∩ t2_i) | p_i ∈ P, p_i ≠ p_j }
```

This works when a package does not appear in the other incompatibility since in that case it implicitly is `Negative(∅)`. Let's take the common concrete example of two dependencies: `{a: t1, b: t2}` and `{b: t3, c: t4}`.

If we consider that `b` is our `p_j`, then prior cause expressed as above gives:

```txt
{ a: t1 ∩ Negative(∅), b: t2 ∪ t3, c: t4 ∩ Negative(∅)}
    == { a: t1, b: t2 ∪ t3, c: t4 }
```

This is the rule of resolution as explained in PubGrub documentation. When `t2` and `t3` are complementary, for example `t3 = not t2`, we have `t2 ∪ t3 = Negative(∅)` so it can be safely removed from the incompatibility.

If we consider that `a` is our `p_j` instead, then prior cause expressed as above gives:

```txt
{ a: t1 ∪ Negative(∅), b: t2 ∩ t3, c: t4 ∩ Negative(∅)}
    == { a: Negative(∅), b: t2 ∩ t3, c: t4 }
    == { b: t2 ∩ t3, c: t4 }
```

In the case where `t2` and `t3` are complementary, we have `b: Positive(∅)` which is the opposite of `Negative(∅)` and can never be satisfied. So that incompatibility is quite safe, its terms cannot be satisfied all together since `b` can literally never be satisfied. It also brings as no information so not very useful.
But in the case where `t2` and `t3` are not complementary, say for example `t2 = Negative(1)` and `t3 = Positive(2)`. This means that `a: t1` depends on `b` at version 1, and that b at version 2 is incompatible with `c: t4`. Then we have

```txt
t2 ∩ t3
    == Negative(1) ∩ Positive(2)
    == Positive(2)
```

so the prior cause expressed as above results in

```txt
{ b: t2 ∩ t3, c: t4 }
    == { b: Positive(2), c: t4 }
```
Meaning `b` at version 2 is incompatible with `c: t4`. That's litterally what we had at the beginning (`{b: t3, c: t4}`) so hey no new information, but at least the general expression is correct!

aleksator:
That's a thorough write up. Might be worth adding it to the docs!

mpizenberg:
> That's a thorough write up. Might be worth adding it to the docs!

Done in #45

PR #25, opened 2020-10-09
feat: use SPDX licence identifier

### LINKED ISSUE #24: SPDX license identifiers
Currently our source files contain full notice recommended by Mozilla in the beginning:
```
// This Source Code Form is subject to the terms of the Mozilla Public
// License, v. 2.0. If a copy of the MPL was not distributed with this
// file, You can obtain one at http://mozilla.org/MPL/2.0/.
```

I propose to change the headers to use SPDX license identifier instead, which looks like this:
```
// SPDX-License-Identifier: MPL-2.0
```
This has advantages of:
1. Following an established standard (https://spdx.dev/).
1. Being human and machine readable.
2. Being shorter.

I can implement the change if there are no objections from the team.

aleksator:
Replaces standard Mozilla header and adds an identifier to files that were missing one.

Closes #24

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,5 @@
+# SPDX-License-Identifier: MPL-2.0
+
 [package]
 name = "pubgrub"
 version = "0.1.0"
--- a/benches/large_case.rs
+++ b/benches/large_case.rs
@@ -1,3 +1,5 @@
+// SPDX-License-Identifier: MPL-2.0
+
 #![feature(test)]
 extern crate test;
 use test::Bencher;
@@ -12,7 +14,8 @@ use pubgrub::version::NumberVersion;
 /// It has not bean minimized. There are meny `add_dependencies` that have no impact on the runtime or output.
 fn large_case(b: &mut Bencher) {
     let s = std::fs::read_to_string("test-examples/large_case_u16_NumberVersion.ron").unwrap();
-    let dependency_provider: OfflineDependencyProvider<u16, NumberVersion> = ron::de::from_str(&s).unwrap();
+    let dependency_provider: OfflineDependencyProvider<u16, NumberVersion> =
+        ron::de::from_str(&s).unwrap();
 
     // bench
     b.iter(|| {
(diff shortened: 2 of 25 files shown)


mpizenberg:
I overall agree with this with one nitpick. Git stores the whole file (in snapshots) when files change in commits. It's not as bad as it seems since things are compressed in git database, but maybe we can avoid small changes on big data (`.ron`) files? I'm not even sure adding a license to this makes much sense. What do you think?

aleksator:
I thought storage of text files is optimized well in git, but I haven't measured.

We can add the header to the commit that introduced the change and force push `dev` branch. If we do that, I would also add (#number) references to PRs that "rebase and merge" strategy didn't add to commit headers.

mpizenberg:
> I thought storage of text files is optimized well in git, but I haven't measured.

As well as typical text compression.

> We can add the header to the commit that introduced the change and force push dev branch. If we do that, I would also add (#number) references to PRs that "rebase and merge" strategy didn't add to commit headers.

Nah better not rewriting history already merged in dev. It will be fine I think. I mostly wanted to draw attention on those little details that are often disregarded. But git is like entropy, it only grows so better be careful ^^

aleksator:
> Nah better not rewriting history already merged in dev.

Well, that would be one advantage of having a development branch separate from master.
Especially when doing tiny changes that won't lead to any conflicts.

> Nah better not rewriting history already merged in dev. It will be fine I think. I mostly wanted to draw attention on those little details that are often disregarded. But git is like entropy, it only grows so better be careful ^^

I'll remove the header from RON file, no problem)

mpizenberg:
> Well, that would be one advantage of having a development branch separate from master.

True

aleksator:
I couldn't create a separate commit since RON file didn't have an empty line in the end as well, and text editors I tried inserted it every time I tried to just remove the license header.
I had to create a separate branch from dev and cherry pick all changes from the old branch except that RON file. :fearful: 
Hence I force pushed, but the commit is identical in everything besides RON changes.

mpizenberg (approved):
Looks good to me!

PR #28, opened 2020-10-09
Proptest index

Eh2406:
This is mostly porting proptests from Cargo. It is mostly ready for a first PR. 

Things that have been intentionally left for a follow up:
- The good tooling around reporting the part of the `dependency_provider` that was used.
- The validation that a `Ok(_)` output has all its dependencies satisfied.
- The  validation that a `Err(_)` output happens when an SAT solver also can't find a solution.

This is going to be unpleasant unless we:
- ~~Impl a time out. So inputs that leave us in an infinite loop (or effectively infinite) just come out as a panic.~~ CC [discussion on zulip](https://rust-lang.zulipchat.com/#narrow/stream/260232-t-cargo.2FPubGrub/topic/How.20to.20add.20a.20timeout.3F)
- ~~Some way to make sure the slow job of shrinking does not happen on CI. (not hard but I need to do it.)~~

Just wanted to share for others to take a look.

### DIFF
--- a/tests/proptest.rs
+++ b/tests/proptest.rs
@@ -71,58 +71,57 @@ pub fn registry_strategy(
     )
         .prop_map(
             move |(crate_vers_by_name, raw_dependencies, reverse_alphabetical)| {
-                let list_of_pkgid: Vec<((String, NumberVersion), bool)> = crate_vers_by_name
+                let mut list_of_pkgid: Vec<(
+                    (String, NumberVersion),
+                    Option<Vec<(String, Range<NumberVersion>)>>,
+                )> = crate_vers_by_name
                     .iter()
                     .flat_map(|(name, vers)| {
-                        vers.iter()
-                            .map(move |x| ((name.clone(), NumberVersion::from(x.0)), x.1))
+                        vers.iter().map(move |x| {
+                            (
+                                (name.clone(), NumberVersion::from(x.0)),
+                                if x.1 { Some(vec![]) } else { None },
+                            )
+                        })
                     })
                     .collect();
                 let len_all_pkgid = list_of_pkgid.len();
-                let mut dependency_by_pkgid: Vec<Vec<(String, Range<NumberVersion>)>> =
-                    vec![vec![]; len_all_pkgid];
                 for (a, b, (c, d)) in raw_dependencies {
                     let (a, b) = order_index(a, b, len_all_pkgid);
                     let (a, b) = if reverse_alphabetical { (b, a) } else { (a, b) };
-                    let ((dep_name, _), _) = &list_of_pkgid[a];
-                    if &(list_of_pkgid[b].0).0 == dep_name {
+                    let ((dep_name, _), _) = list_of_pkgid[a].to_owned();
+                    if &(list_of_pkgid[b].0).0 == &dep_name {
                         continue;
                     }
-                    let s = &crate_vers_by_name[dep_name];
+                    let s = &crate_vers_by_name[&dep_name];
                     let s_last_index = s.len() - 1;
                     let (c, d) = order_index(c, d, s.len());
 
-                    dependency_by_pkgid[b].push((
-                        dep_name.to_owned(),
-                        if c == 0 && d == s_last_index {
-                            Range::any()
-                        } else if c == 0 {
-                            Range::strictly_lower_than(s[d].0 + 1)
-                        } else if d == s_last_index {
-                            Range::higher_than(s[c].0)
-                        } else if c == d {
-                            Range::exact(s[c].0)
-                        } else {
-                            Range::between(s[c].0, s[d].0 + 1)
-                        },
-                    ))
+                    if let (_, Some(deps)) = &mut list_of_pkgid[b] {
+                        deps.push((
+                            dep_name,
+                            if c == 0 && d == s_last_index {
+                                Range::any()
+                            } else if c == 0 {
+                                Range::strictly_lower_than(s[d].0 + 1)
+                            } else if d == s_last_index {
+                                Range::higher_than(s[c].0)
+                            } else if c == d {
+                                Range::exact(s[c].0)
+                            } else {
+                                Range::between(s[c].0, s[d].0 + 1)
+                            },
+                        ))
+                    }
                 }
 
                 let mut out: Vec<_> = list_of_pkgid
                     .into_iter()
-                    .zip(dependency_by_pkgid.into_iter())
-                    .map(|(((name, ver), allow_deps), deps)| {
+                    .map(|((name, ver), deps)| {
                         (
                             name,
                             ver,
-                            if !allow_deps {
-                                vec![("bad".to_owned(), Range::any())]
-                            } else {
-                                let mut deps = deps;
-                                deps.sort_by_key(|(ref d, _)| d.clone());
-                                deps.dedup_by_key(|(ref d, _)| d.clone());
-                                deps
-                            },
+                            deps.unwrap_or_else(|| vec![("bad".to_owned(), Range::any())]),
                         )
                     })
                     .collect();
--- a/tests/proptest.rs
+++ b/tests/proptest.rs
@@ -3,6 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 use pubgrub::range::Range;
+use pubgrub::solver::OfflineDependencyProvider;
 use pubgrub::version::NumberVersion;
 
 use proptest::collection::{btree_map, vec};
@@ -17,13 +18,7 @@ pub fn registry_strategy(
     max_crates: usize,
     max_versions: usize,
     shrinkage: usize,
-) -> impl Strategy<
-    Value = Vec<(
-        String,
-        NumberVersion,
-        Vec<(String, pubgrub::range::Range<NumberVersion>)>,
-    )>,
-> {
+) -> impl Strategy<Value = OfflineDependencyProvider<String, NumberVersion>> {
     let name = string_regex("[A-Za-z][A-Za-z0-9_-]{0,5}")
         .unwrap()
         .prop_filter("reseved names", |n| {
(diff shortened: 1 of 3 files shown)


tests/proptest.rs
```diff
@@ -0,0 +1,436 @@
...
+
+use proptest::collection::{btree_map, vec};
+use proptest::prelude::*;
+use proptest::sample::Index;
+use proptest::string::string_regex;
+
+/// The same as DP but it prefers the opposite versions.
+/// If DP returns versions from newest to oldest, this returns them from oldest to newest.
+#[derive(Clone)]
+struct MinimalDependencyProvider<DP>(DP);
```
mpizenberg: Maybe using something like `ReversedDependencyProvider` or something else closer to the meaning of this.

tests/proptest.rs
```diff
@@ -0,0 +1,436 @@
...
+        self.call_count.set(calls + 1);
+        Ok(())
+    }
+}
+
+#[test]
+#[should_panic]
+fn callback_can_panic() {
+    let mut dependency_provider = OfflineDependencyProvider::<_, NumberVersion>::new();
+    dependency_provider.add_dependencies(0, 0, vec![(666, Range::any())]);
```
mpizenberg, line 82: ahahah good one

tests/proptest.rs
```diff
@@ -0,0 +1,436 @@
...
+        &TimeoutDependencyProvider::new(dependency_provider, 1),
+        0,
+        0,
+    );
+}
+
+fn string_names() -> impl Strategy<Value = String> {
+    string_regex("[A-Za-z][A-Za-z0-9_-]{0,5}")
+        .unwrap()
+        .prop_filter("reseved names", |n| {
```
mpizenberg: reserved

tests/proptest.rs
```diff
@@ -0,0 +1,436 @@
...
+                            } else if c == d {
+                                Range::exact(s[c].0)
+                            } else {
+                                Range::between(s[c].0, s[d].0 + 1)
+                            },
+                        ))
+                    }
+                }
+
+                let mut solver = OfflineDependencyProvider::<N, NumberVersion>::new();
```
mpizenberg: Still old `solver` terminology

tests/proptest.rs
```diff
@@ -0,0 +1,436 @@
...
+        } else {
+            // but that local builds will give a small clear test case.
+            2048 // u32::MAX
+        },
+        result_cache: prop::test_runner::basic_result_cache,
+        .. ProptestConfig::default()
+    })]
+
+    #[test]
+    /// This test is mostly for profiling
```
mpizenberg: Because it uses strings for package ids right?

tests/proptest.rs
```diff
@@ -0,0 +1,436 @@
...
+
+    #[test]
+    /// This tests wheter the allgorithm is still deterministic
+    fn prop_same_on_repeated_runs(
+        (dependency_provider, cases) in registry_strategy(0u16..665, 666)
+    )  {
+        for (name, ver) in cases {
+            let one = resolve(&TimeoutDependencyProvider::new(dependency_provider.clone(), 50_000), name, ver);
+            for _ in 0..3 {
+                match (&one, &resolve(&TimeoutDependencyProvider::new(dependency_provider.clone(), 50_000), name, ver)) {
```
mpizenberg, line 308: Should we check that the difference is not due to the timeout?

tests/proptest.rs
```diff
@@ -0,0 +1,436 @@
...
+    fn prop_passes_int(
+        (dependency_provider, cases) in registry_strategy(0u16..665, 666)
+    )  {
+        for (name, ver) in cases {
+            let _ = resolve(&TimeoutDependencyProvider::new(dependency_provider.clone(), 50_000), name, ver);
+        }
+    }
+
+    #[test]
+    /// This tests wheter the allgorithm is still deterministic
```
mpizenberg: "whether" and "algorithm"

tests/proptest.rs
```diff
@@ -0,0 +1,436 @@
...
+                    _ => panic!("not the same result")
+                }
+            }
+        }
+    }
+
+    #[test]
+    /// MinimalDependencyProvider change what order the candidates
+    /// are tried but not the existence of a solution
+    fn prop_minimum_version_errors_the_same(
```
mpizenberg: Use "reversed" terminology (or similar) instead of minimal maybe

tests/proptest.rs
```diff
@@ -0,0 +1,436 @@
...
+                    **package,
+                    *version,
+                    dependencys.into_iter().filter(|x| x.0 != dependency)
+                )
+            }
+        }
+        for (name, ver) in cases {
+            if resolve(&dependency_provider, name, ver).is_ok() {
+                prop_assert!(
+                    resolve(&TimeoutDependencyProvider::new(dependency_provider.clone(), 50_000), name, ver).is_ok(),
```
mpizenberg: Should removing dependencies also ensure shorter runs?

mpizenberg (commented):
I haven't properly read everything in details yet but this looks amazing!

tests/proptest.rs
Eh2406: Yes. It was one of the first tests I wrote after getting `registry_strategy` working. It may no longer be pulling its weight. I wanted to compare a profile with expensive clones vs one with cheap clones. The other use the faster version ids to not waste CI time, but I want something to test that we are not cloning way way to much. I am open to suggestions.

tests/proptest.rs
Eh2406: Maybe, In Cargo we do not have fine grained control over the timeout. So I have not yet thought about how to take advantage of the better controls here.

tests/proptest.rs
Eh2406: That is a really good thought. That is a thing I did not think to test when working on Cargo as it is not flexible like that. Do you think we should work on it in this PR or in a follow up?

tests/proptest.rs
mpizenberg: It's good to still have a strings example in my opinion, to be closer to real usage when benchmarking things that depend on hash maps. If it's not needed on CI though, is there a way to deactivate a bunch of tests specifically when on CI?

tests/proptest.rs
mpizenberg: Nah, jut a thought I had but it's not correct. A missing dependency could generate an incompatibility that later reduces potential packages. But we should check that the difference, if there is one, is not due to timeouts I think.

tests/proptest.rs
Eh2406: There is always `if std::env::var("CI").is_ok() {return;}`. If it is still worth running then I am ok running on CI. When CI is a bottleneck we can revisit.

tests/proptest.rs
```diff
@@ -0,0 +1,441 @@
...
+                    **package,
+                    *version,
+                    dependencys.into_iter().filter(|x| x.0 != dependency)
+                )
+            }
+        }
+        for (name, ver) in cases {
+            if resolve(&TimeoutDependencyProvider::new(dependency_provider.clone(), 50_000), name, ver).is_ok() {
+                prop_assert!(
+                    resolve(&TimeoutDependencyProvider::new(removed_provider.clone(), 50_000), name, ver).is_ok(),
```
Eh2406, line 363: I generally don't want to `resolve` on a generated `dependency_provider` without some kind of timeout. It is frustrating when CI fails with "terminated do to no output in 10 min" with no info on what input is slow. (looks like that limitation may have been from my Travis days, Github Actions may not have that limitation, but a hang is also frustrating.

What behaviour do you think the test should have when one of these hangs?

mpizenberg: I think it's probable that those two won't take the same amount of time and it's not what this test tries to ensure, so if at least one of the two `resolve` timeouts the test should pass in my opinion. Passing or not passing that test will be much more correlated to the shape of index than to its size if there is an issue somewhere. In fact it would make sense to have different registry strategies, of different sizes and shapes, but that can be explored later.

Concretely, we can let it fail for now. Gather the examples when it fails to rerun them locally and figure if those are true failures. When a few of these happen we can see if it's worth continuing to fail or not.

Eh2406: We will definitely have to work on all of these tests, to ensure they maintain a good signal to noise ratio. If they turn out not to pull their weight, we can alter or remove!
Just to clarify are you ok with merging as is?

mpizenberg (approved):
Amazing work, thanks @Eh2406 !

The only thing I haven't spent time to look at closely is the proptest strategy to generate the index. I'll look at it later when I have time since this is obviously giving us good examples such as the ones we used for benchmarks.

Good to merge in my opinion, just waiting for @aleksator feedback

mpizenberg:
Oh I had one remark actually that I forgot. The name `callback` sounds like something that should be called only once, when things have finished. I have `hook` that come to mind but there are probably better options.

aleksator:
I added some proofreading commits and renamed `ReversedDependencyProvider` to `ReverseDependencyProvider`.
To see why that makes more sense try to read `TimedoutDependencyProvider` instead of `TimeoutDependencyProvider` as an analogy.

I've noticed there are a few unresolved discussions started by @mpizenberg (like `Should we check that the difference is not due to the timeout?`), but PR is already approved. Are we leaving those for later?

aleksator (approved):

Eh2406:
> Are we leaving those for later?

I interpret "Good to merge in my opinion, just waiting for @aleksator feedback" as a yes.

I like @aleksator's renames. I see the point about the name `callback`, can we have something more descriptive? Like `callback` and `hook` give no indication of why or when there called. Maybe `check_up`, `check_in`, `heartbeat`?

Also do we want that callback to take an argument in case we have diagnostics we can share about the progress?

mpizenberg:
We can improve the callback API later, for now let's just find a more descriptive name and merge this. What do you think of `checkpoint` or `loop_checkpoint`?

Eh2406:
That works!

aleksator:
Maybe `check_interrupt[ion]`?.. Hm.

aleksator:
`loop_checkpoint` says where we are going to use that now, but maybe we could add usages elsewhere, so I would rather not encode that in the name.

Some other suggestions:
`cancellation`, `should_cancel`, `should_interrupt`.

Maybe we should discuss this on Zulip :smile:

Eh2406:
https://rust-lang.zulipchat.com/#narrow/stream/260232-t-cargo.2FPubGrub/topic/How.20to.20add.20a.20timeout.3F

Eh2406:
rebased and updated the name to `should_cancel`.

mpizenberg:
Any remarks left @aleksator ?

aleksator (approved):

mpizenberg:
> aleksator merged commit 055fbec 

damn I wasn't fast enought XD. Thanks @aleksator  and amazing work @Eh2406 !

aleksator:
> damn I wasn't fast enought XD. Thanks @aleksator and amazing work @Eh2406 !

That's good, because I managed to edit commit messages :yum:

> Any remarks left @aleksator ?

Maybe we should have separated a timeout feature from proptest port. I realized that when I was adding a conventional commit label: it contained both `feat:` and `test:`. But oh well.

> amazing work @Eh2406 !

Agree! :+1:

PR #26, opened 2020-10-09
Clickable documentation links

aleksator:

PR #31, opened 2020-10-10
Check formatting and documentation on CI

aleksator:


mpizenberg:
Hi @aleksator can you give some explanations for the CI changes happening here. I'm not well versed in CI stuff, especially GitHub actions so a recap would be nice! By the way, why is nightly needed here?

One thing that would be great for CI is to avoid network access if it has already been able to run once, both because it is more resilient and it uses less of the rust servers (crates and all). In particular, it would be nice if we could temporarily add a config that prevent network access to make sure the setup is right. Another reason to cache things is to speed up the CI run. I'm saying this since I see you've split CI into different parts (which is good!) and I've no idea how that plays with caching.

aleksator:
> can you give some explanations for the CI changes happening here.

I've tried to keep commits separate for easier reviewing what is done where.
Basically, I've added checks for documentation and formatting with nightly, narrowed down directories for caching and made CI being done on PR to any branch with the latest stable. Currently it runs for pushes to `dev` as well, and I thought of adding `master` to the list as well. I saw you had it before and removed, so I wanted to ask you: why?

Another question I had: you added `restore-keys: ${{ runner.os }}` in a separate commit. I didn't find any explanations online what benefits it brings compared to not having it. Do you know?

> I'm not well versed in CI stuff, especially GitHub actions so a recap would be nice!

Me neither, it's a learning opportunity for me :smile: 

> By the way, why is nightly needed here?

Remember #26? Links aren't checked and some (like [String]) are not generated on stable.
Besides that, I think it's nice having the latest improvements from nightly for
- formatting
- documentation generation
- clippy (going to add it later)

We need the latest stable to test our library. If you want, we can test under nightly as well. I discussed whether it's worth it with @Eh2406, and we thought that maybe it's not worth increasing our CI time for that.

Another consideration we haven't discussed is if we want to test under Windows and Mac. I could set up our testing on those platforms in addition to `Ubuntu`. Some libraries like `ripgrep` do that.

> One thing that would be great for CI is to avoid network access if it has already been able to run once, both because it is more resilient and it uses less of the rust servers (crates and all).

Overall, I think it would be hard to achieve and would require quite a few sacrifices from our side. It would mean we wouldn't be able to test under neither the latest stable, nor nightly. Cache also needs to be downloaded by the runner I think.

As an alternative, we would have to create our own runner that has everything preinstalled. It would pin nightly though and would have to be updated from time to time. Let me know if you think I should look into it.

> Another reason to cache things is to speed up the CI run. I'm saying this since I see you've split CI into different parts (which is good!) and I've no idea how that plays with caching.

There's nothing in "deploy_documentation.yml" we could cache besides creating our own runner to prevent nightly pull. That workflow doesn't build our project.

As for general caching improvements, @Eh2406 suggested
- adding the output of `cargo +stable -V` to the key
- possibly generating `Cargo.lock` and putting that in the key instead of `Cargo.toml`. This handles the case of our dependencies being updated when we haven't `Cargo.toml`, thus leading to cache miss. It requires doing `cargo update`, which performs a network call to update the index - adding to the discussion above.

aleksator:
> possibly generating Cargo.lock and putting that in the key instead of Cargo.toml. This handles the case of our dependencies being updated when we haven't Cargo.toml, thus leading to cache miss. It requires doing cargo update, which performs a network call to update the index - adding to the discussion above.

Tested this suggestion. It adds about 15-20 seconds to every run, I don't think it's worth it.

mpizenberg:
> I thought of adding master to the list as well. I saw you had it before and removed, so I wanted to ask you: why?

Master will never receive any push that isn't a simple merge from dev so I didn't thought it was useful. I was not so sure about dev, that's why I had testing stuff there too in addition to gh-pages. You can add if you think that's worth it.

> you added restore-keys: ${{ runner.os }} in a separate commit

It would be a shame to restart everything from scratch if `Cargo.toml` changed. We can have multiple backup keys tried one after the other to start from. When a perfect cache hit is found, the cache is not updated at the end but when a backup key was used (or none found) a new cache with exact key is generated at the end. That's my understanding.

> Links aren't checked and some (like [String]) are not generated on stable

Good, keeping nightly for links!

> the latest improvements from nightly for formatting, documentation generation, clippy (going to add it later)

For formatting and clippy it would make more sense to me to not use nightly. The library is supposed to be usable on stable, so should its formatting. In addition clippy is a rather fast moving target, more so on nightly and I've already often beeing "annoyed" by it not passing CI and having to upgrade it locally to just make clippy changes. In addition that removes the nightly requirement for caching for anything except docs.

> If you want, we can test under nightly as well. I discussed whether it's worth it with @Eh2406, and we thought that maybe it's not worth increasing our CI time for that.

Not worth it in my opinion.

> Another consideration we haven't discussed is if we want to test under Windows and Mac. I could set up our testing on those platforms in addition to Ubuntu. Some libraries like ripgrep do that.

That would be amazing!

> Overall, I think it would be hard to achieve and would require quite a few sacrifices from our side. It would mean we wouldn't be able to test under neither the latest stable, nor nightly. Cache also needs to be downloaded by the runner I think.

I think cache is using local network or at least is not subject to the same potential down time of rust servers. So that would be a good thing to persue. We could change the base image to use one that has latest stable. This can be persued in a later PR though.

> There's nothing in "deploy_documentation.yml" we could cache besides creating our own runner to prevent nightly pull. That workflow doesn't build our project.

All the libraries have to be downloaded, and other stuff is cached too because `cargo doc` does way less work on my computer when only changes to the current crate doc are done. This is the output after `cargo clean`:

```txt
   Compiling proc-macro2 v1.0.23
   Compiling unicode-xid v0.2.1
   Compiling syn v1.0.42
    Checking rustc-hash v1.1.0
   Compiling quote v1.0.7
   Compiling thiserror-impl v1.0.20
    Checking thiserror v1.0.20
 Documenting pubgrub v0.1.0 (/home/matthieu/git/mpizenberg/pubgrub-rs)
    Finished dev [unoptimized + debuginfo] target(s) in 8.99s
```
and this is the one after just changing a thing in our doc comments
```txt
 Documenting pubgrub v0.1.0 (/home/matthieu/git/mpizenberg/pubgrub-rs)
    Finished dev [unoptimized + debuginfo] target(s) in 1.17s
```

> adding the output of cargo +stable -V to the key

Seems good, that could become something like

```yaml
        key: ${{ runner.os }}-${{ cargo version here }}-${{ hashFiles('Cargo.toml') }}
        restore-keys: |
          ${{ runner.os }}-${{ cargo version here }}
          ${{ runner.os }}
```

> possibly generating Cargo.lock and putting that in the key instead of Cargo.toml

I'm not a big fan of versioning `Cargo.lock` for a crate but that could change.

aleksator:
> For formatting and clippy it would make more sense to me to not use nightly. The library is supposed to be usable on stable, so should its formatting.

It's 2 independent properties:
1) being able to import our library on stable
2) implementation detail of developers (formatting).

1st would work even if we put all our code in 1 line, so we could use nightly for that.

Another thing to consider: if we put formatting in stable section, it will be performed 3 times: 1 for each OS (once I add that). This is a useless thing to do and just adds to CI time (~15 seconds).

If we want to add a separate job for formatting on stable, we would have to do one of:
1) Download a stable Rust version a second time (all jobs are run in separate VMs). It wouldn't add total CI time since it's done in parallel, but it's bit wasteful.
2) Use old Rust that comes with the image. We can get discrepancies between latest stable formatting on local machine and the one in CI. To fix that, it would require everyone to pin an old stable Rust locally and format with that. It makes for an unpleasant workflow, so I would like to avoid this if possible.

Just realized that we should take a `stable clippy` request into account, which also doesn't make sense to do every time for each OS. We could group formatting + clippy together under a separate job that downloads stable Rust.

---

To summarize, I think the choice is between checking formatting on 
1. nightly along with the documentation
2. stable, during testing (1 time per OS, so 3 times)
3. stable, separate job (1 time, but additional Rust download)

Let me know what you think @mpizenberg @Eh2406.

---

> In addition clippy is a rather fast moving target, more so on nightly and I've already often beeing "annoyed" by it not passing CI and having to upgrade it locally to just make clippy changes.

I heard you, I'll add _stable_ clippy in the future.

> In addition that removes the nightly requirement for caching for anything except docs.

Hm, I haven't thought of caching documentation files before. I'll see if I can add that.

> That would be amazing!

Alright, I will add it to this PR.

> I think cache is using local network or at least is not subject to the same potential down time of rust servers. So that would be a good thing to persue.

I'll keep that in mind when doing changes.

> We could change the base image to use one that has latest stable. This can be persued in a later PR though.

This is not guaranteed, since runners are provided by Github. Thus every Rust project I had a look at (e.g. ripgrep, rust-analyzer, this-error) downloaded a stable Rust as a first step in CI (although the exact methods for that had been different).
It's possible to create custom runners, but those come with maintenance cost.

> Seems good, that could become something like

        key: ${{ runner.os }}-${{ cargo version here }}-${{ hashFiles('Cargo.toml') }}
        restore-keys: |
          ${{ runner.os }}-${{ cargo version here }}
          ${{ runner.os }}

Yep, this is already done in one of the commits.

> I'm not a big fan of versioning Cargo.lock for a crate but that could change.

Not sure if we should do that either. It makes sense for applications for sure, but pinning dependencies for a library is meh when `semver` should do the trick. If something breaks in our dependencies we want to notice it as well.
What I checked was generating Cargo.lock on the fly during CI which took too long.

mpizenberg:
> 1st would work even if we put all our code in 1 line, so we could use nightly for that.

Indeed, though it would be annoying to do that. Same as it is annoying to ask a contributor to setup their IDE to format with nightly (and install the toolchain) when their PR is rejected by CI. But if you or Jacob think some features in rustfmt nightly are worth the add, then let's do it.

> 1. nightly along with the documentation
> 2. stable, during testing (1 time per OS, so 3 times)
> 3. stable, separate job (1 time, but additional Rust download)

I personally prefer (3) but I don't think I know enough to be sure that's the right choice.

> Alright, I will add it to this PR.

About the multi-OS setup, it might be better to do it in a separate (following) PR? but if you want to go for it, go ahead.

> This is not guaranteed, since runners are provided by Github. Thus every Rust project I had a look at (e.g. ripgrep, rust-analyzer, this-error) downloaded a stable Rust as a first step in CI (although the exact methods for that had been different).
It's possible to create custom runners, but those come with maintenance cost.

Yeah let's avoid the custom runner thing.

---------------

Regarding most of the points, it would be very beneficial to have the point of view of someone more experienced, even though none of this is critical so any choice would be an improvement. Maybe we could ask on zulip the point of view of some seasoned maintainers of relatively used crates?

aleksator:
> I personally prefer (3) but I don't think I know enough to be sure that's the right choice.

Alright, I can set that up.

> About the multi-OS setup, it might be better to do it in a separate (following) PR? but if you want to go for it, go ahead.

The change is just a few lines, so I'm not worried about doing it here.

> Regarding most of the points, it would be very beneficial to have the point of view of someone more experienced, even though none of this is critical so any choice would be an improvement. Maybe we could ask on zulip the point of view of some seasoned maintainers of relatively used crates?

That's basically why I reviewed CI files by the most prominent Rust projects. The code speaks better than the words.
I haven't seen granular caching like we've done here with generating keys, so either there's a good reason for that or we are improving in this area.

Our Zulip doesn't have a lot of people, maybe we can ask elsewhere to give some advice or review this PR/the whole setup directly.
I don't know any place connected to Rust where that would be a relevant question though.

Eh2406:
I think it would be reasonable to ask in https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra. And the rest of the conversation sounds eminently reasonable. I am lucky to be working with such thoughtful people!

aleksator:
Alright, this is ready for a review; easier to be done when looking at the commits individually.
Feel free to invite additional people to have a look, I don't have any open questions left myself.

I've removed running this on multiple OSes from this PR due to a bug with caching on MacOS. I have a workaround which I'll submit separately when the corrupt cache I got while testing expires in a week (there isn't a way to invalidate them manually).

aleksator:
I also introduced weekly testing which is up for a discussion.

mpizenberg:
I tried using docker images out of curiosity, here is an example usage: https://github.com/mpizenberg/pubgrub-rs/blob/e0639b7e1aa50107052e0195b618adeb24dc3f6c/.github/workflows/ci.yml#L15

.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
...
+      - name: Get Cargo version
+        id: cargo_version
+        run: echo "::set-output name=version::$(cargo -V | tr -d ' ')"
+        shell: bash
+
+      - name: Download cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/bin/
```
Eh2406: Won't the rust install have set this up? Why cache as well?

aleksator: I was following caching suggestions from The Cargo Book:
https://doc.rust-lang.org/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci

Which directories do you think are redundant?
We can change those in this PR and open an issue with The Cargo Book to unlist those.

Eh2406: `.cargo/bin/` I will asked on https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/why.20cache.20.2Ecargo.2Fbin.2F.20.3F

It it is recommended by the book then lets do it. If the book is wrong for our needs we can do a follow up PR.

aleksator: We can wait for the decision that comes out of that thread.
In our case I don't think caching that makes a difference since we don't use `cargo install` and the dir is going to be empty, so might as well remove it from caching.

mpizenberg:
Shall we merge? seems all good to me

aleksator:
> Shall we merge? seems all good to me

Yep looks like we won't get any advice regarding this. Merging 👍

Issue #20, opened 2020-10-06
Pick packages with a heuristic


Eh2406:
Cargo and the Dart implementation have an optimization where if it is not clear which package to make a decision on it picks the one with the fewest options. We do not yet have that implomented, it could go https://github.com/mpizenberg/pubgrub-rs/blob/f1608ecf2deee92550aa56693c1a97113a8584c9/src/internal/partial_solution.rs#L94. A good heuristics can make a huge difference here. It is not clear how to thread the data thure, but we should do something.

mpizenberg:
Yeah it might force us to precompute the intersection of all derivations in the partial solution, so it might also change a few other places where we have to compute that.

mpizenberg:
It also requires to have the actual list of potential packages, which is not necessary for now since we just pick the first item of the iterator, so it stops evaluation of potential packages when one is found.

Eh2406:
using the dumbest huristick:
```rust
    pub fn pick_package(
        &self,
        dependency_provider: &impl DependencyProvider<P, V>,
    ) -> Option<(P, Term<V>)> {
        self.memory
            .potential_packages()
            .min_by_key(|(p, _)| {
                dependency_provider
                    .list_available_versions(p)
                    .unwrap() // TODO: remove this!
                    .len() // TODO: should be len that matches
            })
            .map(|(p, all_terms)| (p.clone(), Term::intersect_all(all_terms.iter())))
    }
```
Brings the benchmark from .282 sec/iter to .081 sec/iter. So this is definitely worth pursuing.

mpizenberg:
ahah optimization season is open ^^

mpizenberg:
I've replaced back our deterministic hashmap by the default one just to have random order as a strategy. Then running many runs may provide good statistical insight. Having the standard deviation and min/max duration would provid very useful info about what an optimal or a bad strategy might look like performance-wise. Here I get more or less the following results:

```txt
test large_case ... bench:  56,593,558 ns/iter (+/- 43,373,689)
```

I'm not sure what the number in parenthesis is, but if it is the standard deviation, it means that there are indeed wild differences induced by the order of packages. It also gives a rough idea of the types of gains that a very good heuristic might provide. Being able to increase the number of runs could be very useful to get that statistical information. Maybe criterion provides that kind of settings?

Eh2406:
[...]

PR #32, opened 2020-10-10
perf: choose package with fewest versions matching (30x speedup)

Eh2406:
This adds the heuristic that is described in the comment. This is the heuristic asked for in #20, but there may be even better ones to try. Unfortunately, it is hard to tell with the existing benchmark as even this makes the benchmark fly. For me the benchmark 0.175 sec -> 0.006 sec!

### DIFF
--- a/src/internal/partial_solution.rs
+++ b/src/internal/partial_solution.rs
@@ -5,13 +5,19 @@
 //! The partial solution is the current state
 //! of the solution being built by the algorithm.
 
-use crate::internal::assignment::Assignment::{self, Decision, Derivation};
-use crate::internal::incompatibility::{Incompatibility, Relation};
 use crate::internal::memory::Memory;
 use crate::package::Package;
 use crate::term::Term;
 use crate::type_aliases::Map;
 use crate::version::Version;
+use crate::{
+    error::PubGrubError,
+    internal::incompatibility::{Incompatibility, Relation},
+};
+use crate::{
+    internal::assignment::Assignment::{self, Decision, Derivation},
+    solver::DependencyProvider,
+};
 
 /// The partial solution is the current state
 /// of the solution being built by the algorithm.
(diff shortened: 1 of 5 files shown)


src/term.rs
```diff
@@ -108,6 +108,17 @@ impl<V: Version> Term<V> {
     }
 }
 
+// Other useful functions.
+impl<V: Version> Term<V> {
+    /// Check if a term allows a given version.
+    pub fn contains(&self, version: &V) -> bool {
```
mpizenberg: The `term` module was not initially thought of as being part of the public API but making the `Term` type public was necessary for error reporting (no solution) so we made everything except the `Term` enum `pub(crate)`.

mpizenberg: PPS: this already exists as `accept_version`, which is obviously bad naming since you rewrote it. I like `contains` better I think

src/internal/partial_solution.rs
```diff
@@ -94,19 +100,39 @@ impl<P: Package, V: Version> PartialSolution<P, V> {
...
+        dependency_provider: &impl DependencyProvider<P, V>,
+    ) -> Result<Option<(P, Term<V>)>, PubGrubError<P, V>> {
+        let mut out: Option<(P, Term<V>)> = None;
+        let mut min_key = usize::MAX;
+        for (p, term) in self
+            .memory
             .potential_packages()
-            .next()
-            .map(|(p, all_terms)| (p.clone(), Term::intersect_all(all_terms.iter())))
+            .map(|(p, all_terms)| (p, Term::intersect_all(all_terms.iter())))
```
mpizenberg, line 116: This looks like a very pricey thing to do if we effectively iterate over all potential packages. In such case, it would probably be worth keeping in `memory` not a `Vec` of derivations but their intersections directly. I haven't reread the code but I think we may never need individual derivations for anything. Every usage we have of `Memory` could be as efficient or more if we actually had the intersection directly.

To be checked

mpizenberg:
> 0.175 sec -> 0.06 sec!

Did you mean 3x speed up then? XD

src/term.rs
Eh2406: Good catch. Fixed.

src/internal/partial_solution.rs
Eh2406: Yes, That is definitely worth looking into! But with this already being 30x it is hard to tell based on this benchmark. It seams to worth committing now and evaluating that again when we have a new benchmark.

Eh2406:
No, sorry typo. I ment 0.006. 😄

src/internal/partial_solution.rs
mpizenberg: Alright!

mpizenberg (approved):
All good to me! If that's ok with @aleksator too?, let's merge

aleksator:
I've added a small doc change to the `pick_package` function to improve the grammar and remove `todo` that asks to improve the function. We may never find a better heuristic, so it's unclear what would be the next best time to remove it if not now. And if we do improve, the change would be guided by benchmarking and `todo` wouldn't help finding it.

PR #34, opened 2020-10-11
Benchmark v2

Eh2406:
This adds two new synthetic benchmarks. They were generated by applying #32, #29, and #28 and setting the time out to 3sec, then running proptests with release until it fails. Then converting to ron files.


Eh2406:
They are both smaller then `large_case_u16_NumberVersion` but take longer to run. If they turn out to have related performance profiles, then it may makes sense to only keep one.

I am open to more realistic, or size efficient benchmarks. I do think the benchmarks should not depend on live data, (for example crates.io) nor on complicated generation code. As I want to be sure that the benchmarks are comparable over time.

aleksator:
I see the benchmarks only differ by input, but the body for all 3 is the same.
To avoid copy-pasting testing frameworks in other languages have `parametrized tests` that allow to share the body but add annotations on top of the function to specify which inputs this test should be run with.

Is there anything similar in the Rust land?

Quick searching says it's possible with macros, but I would need to research that further if there is any solution which is standard/idiomatic.

Eh2406:
That was defintly on my todo list. I have done it in the past with a homegrown macro.

Edit: https://github.com/frehberg/test-generator may be useful, also it may not be needed when we switch to criterion.

BTW: @mpizenberg sounded like he is not ready for merge https://github.com/mpizenberg/pubgrub-rs/issues/33#issuecomment-707204692

mpizenberg:
Yeah, it was convenient to have the first benchmark in repo, but adding new ones, I'd prefer we evaluate the LFS option. I planned to have a look at that next weekend.

aleksator:
> Yeah, it was convenient to have the first benchmark in repo, but adding new ones, I'd prefer we evaluate the LFS option. I planned to have a look at that next weekend.

I see. RON files in this PR are about ~100k each which I personally don't see as a problem, but we can evaluate LFS option.
I have no experience using it yet; it would be interesting to see how much complexity it adds.

Eh2406:
Also is there a file size small enough that would be ok?

I have been thinking about how to minimize them further. My thought is right a script that:

1. Calculates a (runtime)/(file size)
2. Deletes some random lines in the input
3. Get a new (runtime)/(file size) if it is lower than reset the file
4. go to 2

Eh2406:
I hacked something together, and have had some success making the files smaller with disproportionately smaller runtime changes.

114,568 -> 79,381 large_case_1_u16_NumberVersion.ron
125,430 -> 90,009 large_case_2_u16_NumberVersion.ron
111,299 -> 21,404 large_case_u16_NumberVersion.ron

Eh2406:
I have it set up to run in a loop working on `case_2`. Progress has slowed. However It is often finding that by removing one line it can make it take longer then 1 min per / iter. So that is definitely interesting. It all so often finds input that give `'We should always find a satisfier if called in the right context.'`. I don't know is that an interesting case worth making a minimal example of or is that a case of GIGO. What do you think?

mpizenberg:
> It all so often finds input that give 'We should always find a satisfier if called in the right context.'

Yes that's worth it. It means there is probably a subtle bug or edge case we're not handling correctly. (What does GIGO means by the way?)

Eh2406:
[Garbage in, garbage out](https://en.wikipedia.org/wiki/Garbage_in,_garbage_out). I am deleting blindly, I would not be surprised if some of the inputs are nonsensical.

Also, based on using the timeout functionality from #28 I now know that the `>1min/iter` is not calling `callback`. So even more interesting. Edit: profile shows it is in `find_satisfier_helper`.

I will endeavor to find a minimal example leading to that panic.

Eh2406:
I am closing. We have a real benchmark in the form of https://github.com/pubgrub-rs/pubgrub/issues/44#issuecomment-727289232. So synthetic benchmarks should only be added where they generate pathological performance profiles.

aleksator:
Should we replace our current synthetic benchmark with this one? This one was more demanding, wasn't it?

Eh2406:
I don't think there is reason to remove a benchmark. Whether it is in or out it will take up space in the git history.
These benchmarks where slower, but as I did performance work the %change on these where about the same as on the existing faster one. Similarly the profile of where the code was spending time was about the same for all three. So I don't think they were pulling their weight.

I am liking a structure where the code iterates over all the files and benchmarks the once you have, but we only commit some important or small ones. If there interesting to you, drop them in and you can use them.

aleksator:
@Eh2406 
> I don't think there is reason to remove a benchmark.

I didn't propose to remove the current benchmark, but asked if we should update it with this one.
I thought the one in this branch takes longer to complete, so might make noticing performance changes easier.

mpizenberg:
Jacob point was that all 3 synthetic benchmarks have the same profile and performance characteristics. And since the first one is "big enough" for benchmarks, no need updating it.

PR #35, opened 2020-10-13
Just an experiment (do not merge)

mpizenberg:


aleksator:
Hi @mpizenberg! I wonder what was the reason for `ci: remove usage of cargo version for cache`?
Should I remove it from #31?

aleksator:
By the way, instead of `ci: retry` you can rerun any pipeline from [Actions link on top](https://github.com/mpizenberg/pubgrub-rs/actions).

aleksator:
> Hi @mpizenberg! I wonder what was the reason for ci: remove usage of cargo version for cache?
Should I remove it from #31?

Ah, maybe it just wasn't working with the `ash` shell.

mpizenberg:
> Hi @mpizenberg! I wonder what was the reason for ci: remove usage of cargo version for cache?
Should I remove it from #31?

oh no don't worry, that's because I initially tried with the alpine image which is lighter but the command for that required bash

aleksator:
> oh no don't worry, that's because I initially tried with the alpine image which is lighter but the command for that required bash

I like the idea of using docker container that already has the latest Rust version installed. I knew it was possible in Gitlab, but haven't seen the option here.

Looking closer, it still has to be downloaded rather than being a base for the runner, huh. Might not provide any network savings that way at all...

I wonder if we could use someone else's image for `runs-on` (custom runner) that we wouldn't have to maintain ourselves. Haven't seen that in the wild yet though.

mpizenberg:
> Looking closer, it still has to be downloaded rather than being a base for the runner, huh. Might not provide any network savings that way at all...

Yes, I rerun it to see if it would have to download again the image and it appears it did. Too bad the image is not kept in the runner cache.

aleksator:
> Yes, I rerun it to see if it would have to download again the image and it appears it did. Too bad the image is not kept in the runner cache.

Yeah, a runner is a brand new VM for every job.

mpizenberg:
No need to keep that around. We can come back to it if needed in the closed PRs. It has served its experimental purpose.

PR #38, opened 2020-10-16
test: compare with a sat solver

Eh2406:
This is a follow up https://github.com/mpizenberg/pubgrub-rs/pull/28, it uses the [varisat](https://docs.rs/varisat/0.2.2/varisat/) library to make sure that we only reten `Err` when there is no solution. Ported from Cargo.
Edit: now also includes a check that the solution we found is acceptable to the sat solver.

Edit: Some reading on varisat:
https://jix.github.io/varisat/manual/0.2.0/lib/basic.html
https://jix.github.io/varisat/manual/0.2.0/lib/incremental.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,3 +23,4 @@ serde = { version = "1.0", features = ["derive"], optional = true }
 [dev-dependencies]
 proptest = "0.10.1"
 ron="0.6"
+varisat="0.2.2"
--- a/tests/proptest.rs
+++ b/tests/proptest.rs
@@ -15,6 +15,9 @@ use proptest::prelude::*;
 use proptest::sample::Index;
 use proptest::string::string_regex;
 
+mod sat_dependency_provider;
+use crate::sat_dependency_provider::SatResolve;
+
 /// The same as DP but takes versions from the opposite end:
 /// if DP returns versions from newest to oldest, this returns them from oldest to newest.
 #[derive(Clone)]
@@ -294,6 +297,20 @@ proptest! {
     )  {
         for (name, ver) in cases {
             let _ = resolve(&TimeoutDependencyProvider::new(dependency_provider.clone(), 50_000), name, ver);
+        }
+    }
+
+    #[test]
+    fn prop_sat_errors_the_same(
+        (dependency_provider, cases) in registry_strategy(0u16..665, 666)
+    )  {
+        let mut sat = SatResolve::new(&dependency_provider);
+        for (name, ver) in cases {
+            if let Ok(s) = resolve(&TimeoutDependencyProvider::new(dependency_provider.clone(), 50_000), name, ver) {
+                prop_assert!(sat.sat_is_valid_solution(&s));
+            } else {
+                prop_assert!(!sat.sat_resolve(&name, &ver));
+            }
         }
     }
 
--- a/tests/sat_dependency_provider.rs
+++ b/tests/sat_dependency_provider.rs
@@ -1,0 +1,145 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use pubgrub::package::Package;
+use pubgrub::solver::{DependencyProvider, OfflineDependencyProvider};
+use pubgrub::type_aliases::Map;
+use pubgrub::version::Version;
+use std::hash::Hash;
+use varisat::ExtendFormula;
+
+const fn num_bits<T>() -> usize {
+    std::mem::size_of::<T>() * 8
+}
+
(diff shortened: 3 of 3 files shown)


mpizenberg:
I'm currently learning how varisat works thanks to its guide to review that PR correctly. But I realized something rather independent but still related. If this, and the previous PR for property testing are heavily inspired by code in the cargo repository, we should check the licensing. We might have to change the licensing of some files in the pubgrub-rs repo. Do you mind checking what should be done @Eh2406 ?

aleksator:
> I'm currently learning how varisat works thanks to its guide to review that PR correctly. But I realized something rather independent but still related. If this, and the previous PR for property testing are heavily inspired by code in the cargo repository, we should check the licensing. We might have to change the licensing of some files in the pubgrub-rs repo. Do you mind checking what should be done @Eh2406 ?

Nothing is directly copied, so I would think no changes needed on our part. Moreover, if the original code is written by @Eh2406, he could just say he re-licensed it under MPL-2.0 for this repo (theoretically, I don't think it's needed here).

Eh2406:
Yes the code that was ported is code I originally wrote. So we don't have to deal with that question this time. But thank you for keeping an eye on it!

mpizenberg:
I'm continuing reading but I have a question regarding `sat_at_most_one`. What is the advantage of the binary encoding logic which is hard to think about over just a double loop with only increasing indices to keep just the upper half of the matrix and not duplicate clauses? Something like:

```rust
for (i, v1) in vars.iter().enumerate() {
    for v2 in &vars[i+1..].iter() {
        solver.add_clause(&[v1.negative(), v2.negative()]);
    }
}
```

mpizenberg:
Why using `solver.model().is_some()` in `sat_resolve()` and not in `sat_is_valid_solution()`? This question prompted me to dive a bit more in varisat and I have the impression we are not using it correctly with assumptions.

If my understanding is correct, without ever adding a "hard" unit clause to the sat solver, it will always be successful, and the option in the model will always be `Some()`. Though it may be `Some([])` which we do not want. So except if I'm misunderstanding something we should do something like

```rust
let solution = &self.solver.model();
solution.is_some() && solution != Some(vec![])
```

Or maybe
```rust
let failed_core = &self.solution.failed_core().is_none()
```

In the latter case, and in general with assumptions, we should probably clear all assumptions between two calls to `sat_resolve()`. And since it doesn't seem to be something doable in the API of varisat, it means we should probably clone the solver before every call to `solve()`. And if that is the case, then no need to make assumptions anymore, we can just add the equivalent unit clauses.


In `sat_is_valid_solution`, it is not enough to just solve I think. I believe we have to check that the `failed_core` is empty if we are using assumptions. And for that we need a way to clear previous assumptions before solving, so same remark as before?

Eh2406:
> What is the advantage of the binary encoding logic which is hard to think about over just a double loop?

I don't know if it makes a practical difference for our use case. The double loop adds `n*(n-1)/2` clauses. The binary encoding adds `n*log_bits(n)` clauses. The linked paper found the binary encoding to be much faster in there use cases.

Calling [assume](https://docs.rs/varisat/0.2.2/varisat/solver/struct.Solver.html#method.assume) replaces the current set of assumed literals. So I don't think we need to clone.

As such the difference between `self.solver.solve()` and `self.solver.model().is_some()`, is effectively nothing. Although the former is newer, faster, and clearer so let me fix that.

aleksator (commented):
Minor nit: conventional commits use "test:" rather than "tests:".

aleksator:
I'm yet to read the paper, but what you've done here is awesome.

> /// The SAT library does not optimize for the newer version,
/// so the selected packages may not match the real resolver.

We probably can't check that the solution contains exact set of packages either, right? As different versions can contain different sets of dependencies. Otherwise this test would be too good to be true. :laughing:

mpizenberg:
> We probably can't check that the solution contains exact set of packages either, right?

Yes, we only get a solution.

Thanks for the clarifications @Eh2406 . I had misread a lot of things in the varisat doc. Seems all good to me!

aleksator:
> Minor nit: conventional commits use "test:" rather than "tests:".

Fixed commit message in `dev` before anyone had a chance to pull it. Don't hit me. :smile:

Eh2406:
This test does have a great shape, in that it takes any set of dependencies and can return if there a correct answer. As such I would like to set up tests that runs this against the ron files to ensure that the benchmarks are not going fast by getting the wrong answer. CC #43.

PR #41, opened 2020-10-17
ci: enforce project conventions

aleksator:
Check that commit messages follow [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) specification.

It's still possible to make a mistake when accepting future PRs and changing a title during squash merge, but I don't know how to prevent that.
There is such thing as precommit hooks, but could they be set up for the Github? Needs investigating.

### DIFF
--- a/.commitlintrc.yml
+++ b/.commitlintrc.yml
@@ -1,0 +1,5 @@
+extends:
+  - '@commitlint/config-angular'
+
+rules:
+  body-max-line-length: [2, 'always', 72]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,3 +82,16 @@ jobs:
         env:
           RUSTDOCFLAGS: -D warnings
         run: cargo +nightly doc --no-deps --document-private-items
+
+  check_commit_conventions:
+    name: Check commit conventions
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Check commit conventions
+        uses: wagoid/commitlint-github-action@v2
+        with:
+          configFile: .commitlintrc.yml
+          failOnWarnings: true


aleksator:
Ready for review.

There is some sacred knowledge on contribution guidelines not documented anywhere, such as do PRs against `dev` branch, follow conventional commits, add SPDX license header to new files... I'll write those down in the follow up PR.

mpizenberg (approved):
> Check that commit messages follow Conventional Commit specification.

I can imagine that being a pain if forcing to rename commits that will eventually be squashed anyway. But I'm ok experimenting it.

Dylan-DPC-zz:
Generally not a fan of this tbh, it adds a barrier for newcomers and could put off contributors 

The same with branching - normally we just make people send prs against master and then we cut a release by  taking the latest master when it's time to make a release.

mpizenberg:
> The same with branching - normally we just make people send prs against master and then we cut a release by taking the latest master when it's time to make a release.

That's mostly due to the way I'm handling my projects and this one just being transferred out of my namespace yesterday. I tend to favor the user point of view, and I've often been put off by projects where what I see on github would not match how I can use a library. This happens quite often with major version changes being worked on.

That being said, if most of you think it would be better to have master instead of dev for the main WIP branch, I'm ok with it.

aleksator:
> Generally not a fan of this tbh, it adds a barrier for newcomers and could put off contributors

While this could be so, it sets up quality standards for the project. I personally want to see external contributions being as close as possible to the code that the core team would write.

As for conventional commits specifically, it has several benefits that can only be realized only when _everyone_ follows it.
Among those are automatic changelog generation and auto semver releases. I can see myself adding tooling for that in the future (for changelog for sure).

Also, filtering the history for e.g. `feat:` is rather neat even without special tooling support.

> That being said, if most of you think it would be better to have master instead of dev for the main WIP branch, I'm ok with it.

I like the current setup of published master and development experiments in `dev`.

aleksator:
> I can imagine that being a pain if forcing to rename commits that will eventually be squashed anyway. But I'm ok experimenting it.

Yeah let's at least try it. It a) adds discipline and b) allows rebase merges without edits since everything is already categorized.

mpizenberg:
> Among those are automatic changelog generation and auto semver releases

I stumbled upon this and thought I should mention it here: https://github.com/yoshuawuyts/changelog

Dylan-DPC-zz:
> I like the current setup of published master and development experiments in dev.

The general prob is :
1) people often end up sending prs to master even though if we mention it everywhere 
2) you often need to rebase in both directions which could be painful 
3) from past experiences it's a pain to set up CI for 2 branches and things like bors (not that we need it now ) often don't work for 2 branch-process

mpizenberg:
> people often end up sending prs to master even though if we mention it everywhere

That may happen indeed. We can edit the PR to change the branch to dev apparently so that's not much on an issue.

> you often need to rebase in both directions which could be painful

Could you give an example of this. I'm not very familiar with those situations. Since no commit will ever happen on master, just merge from dev, in which situations do we need to rebase in another situation?

> from past experiences it's a pain to set up CI for 2 branches and things like bors (not that we need it now ) often don't work for 2 branch-process

I've never use bors either and don't know its role. Could we just setup CI for dev? Master should never have its own commits anyway.

EDIT: In fact, I'd have like to setup `dev` as the "main" branch in the repo config, and setup `master` as the "home" branch but we can't dissociate the two. The branch displayed at repository home is the "main" branch, so we kept `master` being the "main" branch in the config. That would be nice if we could have a distinction but not possible.

aleksator:
> I stumbled upon this and thought I should mention it here: https://github.com/yoshuawuyts/changelog

Worth looking into when choosing changelog generator, but my guess is since it doesn't require any commit conventions we might get better mileage off one tailored for Conventional Commits specifically.

aleksator:
> people often end up sending prs to master even though if we mention it everywhere

A fun idea: setup CI for PRs to master that instantly fail with the message: "PRs go to `dev` branch" :smile:

mpizenberg:
In my mind there were two advantages of having a separate dev branch.
1. master as the home page reflects what the last release is and this is better suited from the user perspective. It comes with the downside that it's not as straightforward from the contributor perspective.
2. master can be used as a "release" branch (and should probably be renamed that way) such that only working commits live in master. But in fact if we use tags and GitHub releases appropriately there would be no need for that.

So if it's better for CI, I'm ok removing the master branch. We could keep `dev` as the main branch. That would also resolve the "master" naming convention issue.

> Worth looking into when choosing changelog generator, but my guess is since it doesn't require any commit conventions we might get better mileage off one tailored for Conventional Commits specifically.

Convention is fine, but humans make errors (like not realizing there is a breaking change), that's why rust is such a great improvement from C++. I'd have more trust in a tool doing code analysis than only using human metadata (commits names).

aleksator:
I think putting users forth over developer convenience is a worth tradeoff in many cases, including this. There are always many more users than there are developers after all and that's for whom the library is designed (although developer satisfaction is also important, since we are all here doing things for fun in our free time).

That's why I was on board with branch distinction that Matthieu offered: it doesn't complicate the setup much for the developers once you know the catch (literally 1 second branch switch during PR submission), but keeps the advantages described above for users.

aleksator:
> Convention is fine, but humans make errors (like not realizing there is a breaking change), that's why rust is such a great improvement from C++. I'd have more trust in a tool doing code analysis than only using human metadata (commits names).

And that makes perfect sense, so I would suggest trying out various options for whomever is going to be choosing a tooling for that. Before looking at the output produced by those tools it's impossible to say what works better. Manual changes might be needed in both cases.

PR #42, opened 2020-10-18
docs(readme): remove link to API docs of the dev branch

mpizenberg:


mpizenberg:
So the deployment of docs is working alright. The url is just a little bit redundant since there is the name of repo followed by the name of the crate, so it is located at https://pubgrub-rs.github.io/pubgrub/pubgrub/.

I've removed the link from the readme and updated the one from the repo description. Eventually, the repo description will link to the guide instead but we can add the link to the dev API docs in the contribution guidelines.

aleksator:
Ah, that's what I was missing yesterday: I could just add `/pubgrub/` to the root URL of https://pubgrub-rs.github.io/pubgrub/ which gave 404, and I tried to make root url work instead.

aleksator (approved):

aleksator:
Should we change the link to https://docs.rs/pubgrub instead?

I'm okay with either way; feel free to merge as is or change it.

mpizenberg:
The whole readme will receive improvements and fancy badges later when we prepare v0.2 so no need for now

PR #45, opened 2020-10-20
PubGrub guide

mpizenberg:
Before publishing the next version, I'd like to write guide, that could be a reference for people wanting to use or contribute to this crate. Mdbook seems like the right fit for this so I've just initialized a book for this WIP pull request. Here is what I currently see for the content of that guide. Let me know what you think.

1. Version solving
2. Using the pubgrub crate
    1. Basic example with OfflineDependencyProvider
    2. Writing your own dependency provider
    3. Caching dependencies in a DependencyProvider
    4. Version strategies in a DependencyProvider
    5. Solution and error reporting
    6. Writing your own error reporting logic
3. Internals of the PubGrub algorithm
    1. Overview of the algorithm
    2. Terms
    3. Incompatibilities
    4. Partial solution
    5. Conflict resolution
    6. Building a report tree
4. Testing and benchmarking
    1. Notes on reproducibility
    2. Property testing
    3. Benchmarking
5. How can I contribute? Here are some ideas


Eh2406:
I know this is wip but I just read it over and it looks grand so far! Thank you!

mpizenberg:
Thanks! don't hesitate to read when you want but indeed keep in mind it's wip. I'll let you guys know when that's ready for a review.

mpizenberg:
Alright, I've written a big part of the guide already and would love to have some feedback on what you think of the structure of the guide and its content. Beware that sections related to the dependency provider are still empty because I prefer to wait for the ongoing PRs that should land very soon. Also I didn't fill the testing and benchmarking sections. A deployed version of the guide is available at the following address for a more pleasant reading than source code ^^

https://5f9c4e71d7f5850c5e58c7b5--pubgrub-rs-guide.netlify.app/

Eh2406:
Just finished reading through it. It looks amazing! Thank you for writing this up!

The one place that sounded a little stilted was [Conflict resolution overview](https://5f9c4e71d7f5850c5e58c7b5--pubgrub-rs-guide.netlify.app/internals/overview.html#conflict-resolution-overview). Not sure I know how to make it clearer.

mpizenberg:
> The one place that sounded a little stilted was Conflict resolution overview. Not sure I know how to make it clearer.

I've not put the time yet to understand exactly why there is a loop there. Meaning, why exactly when the decision levels of the satisfier and previous satisfier are the same does it mean that this is not the root cause of the problem. Maybe when I take the time to understand that more deeply, I'll find the appropriate words to describe it.

mpizenberg:
@Eh2406 alright, I think I've cleared up a lot of things on conflict resolution. Let me know what you think.

I also realized that analysis will enable cleaning up some things in the code too so that's great!

Eh2406:
This is all looking really good so far! Thank you.

One note is that I think coz is mostly for multithreaded programs. I have not seen examples of it being used effectively on single threaded code. (Please correct me if I am uninformed.) So perhaps traditional profilers may be adequate for this project.

mpizenberg:
> One note is that I think coz is mostly for multithreaded programs.

Yes it appears so. I haven't used much profilers though so I don't know how useful that is compared to other profilers for monothreaded programs. Maybe the tooling to mark some parts of the code make the presentation of the analysis clearer, maybe not. To be honest, I was mostly interested in the "stabilizer" part of that talk, that randomizes code layout. But it appears that tool was custom made for an old version of LLVM and not usable with rust benchmarking options.

Eh2406:
On zulip you asked for some help with this, so looking over the missing parts of the guide:
- **Writing your own dependency provider** I agree we should weight for the API to stabilize.
- **Caching dependencies in a DependencyProvider** TLDR: depending on your use case you probably want a cache and `RefCel` may be needed for that. Is there more you want in that section?
- **Strategical decision making in a DependencyProvider** TLDR: why is fewest choices good? Is there more you want in that section?
- **Property testing: End-to-end** I can take a stab at.
- **Property testing: SAT solver** I can take a stab at.
- **Benchmarking: Performance analysis** What did you have in mind for this section?
- **Benchmarking: Profiling** TLDR: How to run our benchmarks, how to run them in a profiler. Is there more you want in that section?

I will try to start by wright up some on the **Property testing**.

Eh2406:
Well there is some words about SAT, and I clearly don't know how to get the LaTeX to work.

Eh2406:
Added some words on `registry_strategy`. Feel free to use, edit, or throw away my commits as it helps make the best guide we can.

Eh2406:
As usual great stuff!

> > Pub chooses the latest matching version of the package with the fewest versions

The paragraph that was bouncing around my head was something like, "let's imagine that one of the packages is limited to a range that does not have any versions in it. When we decide to look at that package we will need to backtrack, and throw away some work already done. So any work we do by deciding on a different packaged first is just more work to be thrown away. Consequently, we always want to pick a package that has no versions available before all others. Now let's imagine that one of the packages is limited to a range that only has one version in it. When we decide to look at that package we will have no choice but to add that versions dependencies to the constraints, which can only limit options going forward. Consequently, we should pick a package that only has one version available before packages with more. Generalizing this to 'fewest versions', is just a heuristic but not a bad one."

> Due to potential backtracking, the resolver may ask dependencies of a package multiple times.

I think #58 will removed recalculating dependencies when they are already in the `incompatibilities` ([here](https://github.com/pubgrub-rs/pubgrub/pull/58/commits/52850a4768c6eb7859e691154ec098adb88a0127#diff-cbd16a2b1fdb7e2f39b676831dfaee5250000b403a9f1fcfcbfede424d7cbf73R170-R174)). I think that logic can be applied separately if the other changes in that PR don't end up working out. Given that I'd prefer not to claim that we will ask multiple times, although I would also not be ready to guarantee that we don't.

mpizenberg:
> I think #58 will removed recalculating dependencies when they are already in the incompatibilities

Thanks fixed that.

> The paragraph that was bouncing around my head was something like

Modified it and integrated it in the "Picking a package" paragraph.

mpizenberg:
Rebased on dev

mpizenberg:
The CI is configured to publish the guide on Netlify at this address: https://pubgrub-rs-guide.netlify.app/ I've configured the site on my Netlify account with auth tokens added as secrets in this repo config.

**Question**: should we keep the guide under the same repository?

One advantage of doing so is that the guide versions follow it's associated crate version.
One inconvenience of doing so is the same, so it may prevent publishing guide improvements when in between crate publications.

I think the guide should be in it's own repository so that people can easily contribute to it and improve it independently of the main repository. If that's ok with you I'll create a new repository under pubgrub-rs/guide.

aleksator:
> I think the guide should be in it's own repository so that people can easily contribute to it and improve it independently of the main repository. If that's ok with you I'll create a new repository under pubgrub-rs/guide.

I don't mind, sounds similar to what The Rust Book does.
Question is, do we follow `master`/`dev` branch split for a guide like we do in this repository? Meaning, guide in `master` should explain what currently is in published version, while `dev` guide gets constantly updated when we do development changes.

Eh2406:
Is it possible to have two copies of the book, one for master / one for dev?

mpizenberg:
> Is it possible to have two copies of the book, one for master / one for dev?

Yep quite easily by configuring two "sites" on netlify, each have a different unique ID and publishing each with different CI rules

mpizenberg:
I'll move everything here into a new repo now before @aleksator starts reviewing this.

mpizenberg:
Switch to https://github.com/pubgrub-rs/guide done

Eh2406:
If everything is safely moved over, can we delete the branch? (just to clean up the history.)

Eh2406:
Also a link I am very excited to read is https://github.com/nnethercote/perf-book

PR #46, opened 2020-10-21
refactor!: define Dependencies type

aleksator:
This is a work I started doing while working on Clippy suggestions. I've extracted it into a separate PR since this affects the public API and is rather large on its own.

What I've done here was to extract a separate `Dependencies` type that is now returned from `DependencyProvider`.
Before:
```
fn get_dependencies(...) -> Result<Option<Map<P, Range<V>>>, Box<dyn Error>>
```
After:
```
fn get_dependencies(...) -> Result<Dependencies<P, V>, Box<dyn Error>>
```

Doing this was suggested by [type_complexity](https://rust-lang.github.io/rust-clippy/master/index.html#type_complexity) lint, which my first reaction was to just disable. After thinking more about it though, it actually did make sense for me to
- Define our own enum here instead of `Option`. This encodes our intent in the type system and clarifies that the difference between `None` and an `empty Map`. It was mentioned in the docs, but I think it could be more easy glanced over while actually coding.
- Clarify what this `Map` should contain using a newtype ~`AllowedVersions`~ `DependencyConstraints` with its own documentation.

That's the substance, other changes just piled on to make the tests pass.

I see this as an ergonomics improvement and can't wait to hear your thoughts about it!

### DIFF
--- a/src/solver.rs
+++ b/src/solver.rs
@@ -68,7 +68,6 @@
 use std::collections::BTreeSet as Set;
 use std::error::Error;
 use std::hash::Hash;
-use std::iter::FromIterator;
 
 use crate::error::PubGrubError;
 use crate::internal::core::State;
@@ -187,41 +186,7 @@ pub enum Dependencies<P: Package, V: Version> {
 /// inside [DependencyConstraints] and [Dependencies::Unknown]:
 /// the former means the package has no dependencies and it is a known fact,
 /// while the latter means they could not be fetched by [DependencyProvider].
-#[derive(Debug, Clone)]
-#[cfg_attr(
-    feature = "serde",
-    derive(serde::Serialize, serde::Deserialize),
-    serde(transparent)
-)]
-pub struct DependencyConstraints<P: Package, V: Version>(Map<P, Range<V>>);
-
-impl<P: Package, V: Version> Default for DependencyConstraints<P, V> {
-    fn default() -> Self {
-        DependencyConstraints(Map::default())
-    }
-}
-
-impl<P: Package, V: Version> DependencyConstraints<P, V> {
-    /// An iterator over inner values of [Map<P, Range<V>>](crate::type_aliases::Map)
-    pub fn iter(&self) -> impl Iterator<Item = (&P, &Range<V>)> {
-        self.0.iter()
-    }
-}
-
-impl<P: Package, V: Version> IntoIterator for DependencyConstraints<P, V> {
-    type Item = (P, Range<V>);
-    type IntoIter = std::collections::hash_map::IntoIter<P, Range<V>>;
-
-    fn into_iter(self) -> Self::IntoIter {
-        self.0.into_iter()
-    }
-}
-
-impl<P: Package, V: Version> FromIterator<(P, Range<V>)> for DependencyConstraints<P, V> {
-    fn from_iter<T: IntoIterator<Item = (P, Range<V>)>>(iter: T) -> Self {
-        DependencyConstraints(iter.into_iter().collect())
-    }
-}
+pub type DependencyConstraints<P, V> = Map<P, Range<V>>;
 
 /// Trait that allows the algorithm to retrieve available packages and their dependencies.
 /// An implementor needs to be supplied to the [resolve] function.
@@ -307,10 +272,7 @@ impl<P: Package, V: Version + Hash> OfflineDependencyProvider<P, V> {
     /// Lists dependencies of a given package and version.
     /// Returns [None] if no information is available regarding that package and version pair.
     fn dependencies(&self, package: &P, version: &V) -> Option<DependencyConstraints<P, V>> {
-        self.dependencies
-            .get(package)?
-            .get(version)
-            .map(|m| m.iter().map(|x| (x.0.clone(), x.1.clone())).collect())
+        self.dependencies.get(package)?.get(version).cloned()
     }
 }
 
--- a/src/solver.rs
+++ b/src/solver.rs
@@ -305,8 +305,7 @@ impl<P: Package, V: Version + Hash> OfflineDependencyProvider<P, V> {
     }
 
     /// Lists dependencies of a given package and version.
-    /// Returns [Dependencies::Unknown] if no information is available
-    /// regarding that package and version pair.
+    /// Returns [None] if no information is available regarding that package and version pair.
     fn dependencies(&self, package: &P, version: &V) -> Option<DependencyConstraints<P, V>> {
         self.dependencies
             .get(package)?
--- a/src/solver.rs
+++ b/src/solver.rs
@@ -76,7 +76,7 @@ use crate::internal::incompatibility::Incompatibility;
 use crate::internal::partial_solution::PartialSolution;
 use crate::package::Package;
 use crate::range::Range;
-use crate::type_aliases::Map;
+use crate::type_aliases::{Map, SelectedDependencies};
 use crate::version::Version;
 
 /// Main function of the library.
@@ -85,7 +85,7 @@ pub fn resolve<P: Package, V: Version>(
     dependency_provider: &impl DependencyProvider<P, V>,
     package: P,
     version: impl Into<V>,
-) -> Result<Map<P, V>, PubGrubError<P, V>> {
+) -> Result<SelectedDependencies<P, V>, PubGrubError<P, V>> {
     let mut state = State::init(package.clone(), version.into());
     let mut added_dependencies: Map<P, Set<V>> = Map::default();
     let mut next = package;
--- a/src/solver.rs
+++ b/src/solver.rs
@@ -68,6 +68,7 @@
 use std::collections::BTreeSet as Set;
 use std::error::Error;
 use std::hash::Hash;
+use std::iter::FromIterator;
 
 use crate::error::PubGrubError;
 use crate::internal::core::State;
@@ -134,13 +135,13 @@ pub fn resolve<P: Package, V: Version>(
                 version: v.clone(),
                 source: err,
             })? {
-            None => {
+            Dependencies::Unavailable => {
                 state.add_incompatibility(|id| {
                     Incompatibility::unavailable_dependencies(id, p.clone(), v.clone())
                 });
                 continue;
             }
-            Some(x) => x,
+            Dependencies::Known(x) => x,
         };
 
         // Add that package and version if the dependencies are not problematic.
@@ -170,6 +171,58 @@ pub fn resolve<P: Package, V: Version>(
     }
 }
 
+/// An enum used by [DependencyProvider] that holds information about package dependencies.
+/// For each [Package] there is a [Range] of concrete versions it allows as a dependency.
+#[derive(Clone)]
+pub enum Dependencies<P: Package, V: Version> {
+    /// Package dependencies are unavailable.
+    Unavailable,
+    /// Container for all available package versions.
+    Known(AllowedVersions<P, V>),
+}
+
+/// Subtype of [Dependencies] which holds information about
+/// all possible versions a given package can accept.
+/// There is a difference in semantics between an empty [Map<P, Range<V>>](crate::type_aliases::Map)
+/// inside [AllowedVersions] and [Dependencies::Unavailable]:
+/// the former means the package has no dependencies and it is a known fact,
+/// while the latter means they could not be fetched by [DependencyProvider].
+#[derive(Debug, Clone)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(transparent)
+)]
+pub struct AllowedVersions<P: Package, V: Version>(Map<P, Range<V>>);
+
+impl<P: Package, V: Version> Default for AllowedVersions<P, V> {
+    fn default() -> Self {
+        AllowedVersions(Map::default())
+    }
+}
+
+impl<P: Package, V: Version> AllowedVersions<P, V> {
+    /// An iterator over inner values of [Map<P, Range<V>>](crate::type_aliases::Map)
+    pub fn iter(&self) -> impl Iterator<Item = (&P, &Range<V>)> {
+        self.0.iter()
+    }
+}
+
+impl<P: Package, V: Version> IntoIterator for AllowedVersions<P, V> {
+    type Item = (P, Range<V>);
+    type IntoIter = std::collections::hash_map::IntoIter<P, Range<V>>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.into_iter()
+    }
+}
+
+impl<P: Package, V: Version> FromIterator<(P, Range<V>)> for AllowedVersions<P, V> {
+    fn from_iter<T: IntoIterator<Item = (P, Range<V>)>>(iter: T) -> Self {
+        AllowedVersions(iter.into_iter().collect())
+    }
+}
+
 /// Trait that allows the algorithm to retrieve available packages and their dependencies.
 /// An implementor needs to be supplied to the [resolve] function.
 pub trait DependencyProvider<P: Package, V: Version> {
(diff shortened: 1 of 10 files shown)


aleksator:
`commitlint` was failing the pipeline, because the the first commit had an exclamation mark to indicate a breaking change.  
I found a workaround of extending `config-conventional` and pulling relevant parts from `config-angular` that we were using before, and used it in the second commit.

[I have created an issue in their repository](https://github.com/conventional-changelog/commitlint/issues/2226).

src/solver.rs
```diff
@@ -135,13 +137,13 @@ pub fn resolve<P: Package, V: Version>(
                 version: v.clone(),
                 source: err,
             })? {
-            None => {
+            Unavailable => {
```
mpizenberg: Maybe keeping the `Dependencies::` prefix would be clearer

src/solver.rs
```diff
@@ -68,13 +68,15 @@
...
 use std::hash::Hash;
+use std::iter::FromIterator;
 
 use crate::error::PubGrubError;
 use crate::internal::core::State;
 use crate::internal::incompatibility::Incompatibility;
 use crate::internal::partial_solution::PartialSolution;
 use crate::package::Package;
 use crate::range::Range;
+use crate::solver::Dependencies::{Known, Unavailable};
```
mpizenberg: That's weird inside `solver.rs` in my opinion

src/solver.rs
```diff
@@ -171,6 +173,58 @@ pub fn resolve<P: Package, V: Version>(
     }
 }
 
+/// An enum used by [DependencyProvider] that holds information about package dependencies.
+/// For each [Package] there is a [Range] of concrete versions it allows as a dependency.
+#[derive(Clone)]
+pub enum Dependencies<P: Package, V: Version> {
```
mpizenberg, line 176: Much nicer than the option indeed!

src/solver.rs
```diff
@@ -171,6 +173,58 @@ pub fn resolve<P: Package, V: Version>(
...
+/// inside [AllowedVersions] and [Dependencies::Unavailable]:
+/// the former means the package has no dependencies and it is a known fact,
+/// while the latter means they could not be fetched by [DependencyProvider].
+#[derive(Debug, Clone)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(transparent)
+)]
+pub struct AllowedVersions<P: Package, V: Version>(Map<P, Range<V>>);
```
mpizenberg: I don't like much that type. It feels too much

src/solver.rs
```diff
@@ -202,7 +256,7 @@ pub trait DependencyProvider<P: Package, V: Version> {
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(transparent))]
 pub struct OfflineDependencyProvider<P: Package, V: Version + Hash> {
-    dependencies: Map<P, Map<V, Map<P, Range<V>>>>,
+    dependencies: Map<P, Map<V, AllowedVersions<P, V>>>,
```
mpizenberg: If the type was introduced because of this type complexity, I think I'd prefer another type defined here.
Maybe something like `type DependencyStore = Map<P, Map<V, ...>>`

src/solver.rs
mpizenberg: Or maybe that's the type that should be named `Dependencies` and the other should be renamed since those are not dependencies but the result of a dependency request

mpizenberg (commented):
I globally like the refactor I think but not as much the names. I feel like `AllowedVersions` should in fact be `Dependencies` and `Dependencies` should be something else. I'll let you know if I think of a better name

src/solver.rs
aleksator: This exact change wasn't due to `type_complexity` lint, but because it was a bit hard for me to keep that triple map structure in mind. Using newly introduced type here made it easier for me to understand at a glance.

But that may be just me.

src/solver.rs
aleksator: Yeah I made the same mistake in a sister PR, defining a type where I wanted to use it and not thinking more about general structure.

Where would you prefer it to be placed? Maybe a new `dependency` module, where we could move this and `DependencyProvider`?

src/solver.rs
aleksator: Agree, will change.

aleksator:
> I globally like the refactor I think but not as much the names. I feel like AllowedVersions should in fact be Dependencies and Dependencies should be something else. I'll let you know if I think of a better name

Thanks! Naming is of course a subject for discussion, feel free to start a Zulip thread if you have any suggestions.

There reason why named it `AllowedVersions` rather than `Dependencies` was because the word "dependency" brings a picture of (package, exact_version) in my mind, and I wanted to indicate that it's not a final dependency, but a range of possible solutions.

src/solver.rs
aleksator: I'll sleep on the naming.

src/solver.rs
mpizenberg: ah no I meant it's weird to use the crate you're in just to expose enum variants. But if you use them qualified, then you can remove this `use` line

mpizenberg:
> There reason why named it `AllowedVersions` rather than `Dependencies` was because the word "dependency" brings a picture of (package, exact_version) in my mind, and I wanted to indicate that it's not a final dependency, but a range of possible solutions.

Interesting, for me it's the opposite, I picture it as the range of possible versions

mpizenberg:
And since it's my "hey here is a great talk about it" time tonight, I highly recommend "Spec-ulation" by Rich Hickey: https://youtu.be/oyLBGkS5ICk

aleksator:
> And since it's my "hey here is a great talk about it" time tonight, I highly recommend "Spec-ulation" by Rich Hickey: https://youtu.be/oyLBGkS5ICk

Thank you, that's going to be a talk for my evening today :)

> Interesting, for me it's the opposite, I picture it as the range of possible versions

I don't mind renaming it in this PR to `Dependencies`. Just curious, what would you call a concrete picked dependency version?

mpizenberg:
> Just curious, what would you call a concrete picked dependency version?

Maybe locked dependency versions, exact dependencies or selected dependency versions depending on the context. "picked dependency versions" is not bad :). Maybe one could be `DependencyConstraints` and the other `DependencyVersions`? Like this there is nothing just called `Dependencies`.

aleksator:
> And since it's my "hey here is a great talk about it" time tonight, I highly recommend "Spec-ulation" by Rich Hickey: https://youtu.be/oyLBGkS5ICk

Great talk btw, still awake because I couldn't stop watching it. I wonder if `cargo` makes his point moot with its auto namespace for conflicting versions.

> Maybe one could be `DependencyConstraints` and the other `DependencyVersions`? Like this there is nothing just called `Dependencies`.

Hm, not sure if `DependencyConstraints::Unavailable` sounds right. Does `DependencyConstraints::Unknown` sound better?
As for `DependencyVersions`, I think both that and just `Dependencies` fit nicely here.

I'll compose all the combinations in this message and have a look at them again in the morning:
```
pub enum DependencyConstraints<P: Package, V: Version> {
    Unavailable,
    Known(DependencyVersions<P, V>),
}
```
```
pub enum DependencyConstraints<P: Package, V: Version> {
    Unknown,
    Known(DependencyVersions<P, V>),
}
```
```
pub enum DependencyConstraints<P: Package, V: Version> {
    Unavailable,
    Known(Dependencies<P, V>),
}
```
```
pub enum DependencyConstraints<P: Package, V: Version> {
    Unknown,
    Known(Dependencies<P, V>),
}
```

mpizenberg:
> Hm, not sure if DependencyConstraints::Unavailable sounds right

The `DependencyConstraints` vs `DependencyVersions` was for your question about difference between `Map<P, Range<V>>` and `Map<P, V>`. For the result of the `get_dependencies` method, we'd have to signal that it's the result of a request that may have failed. The `Option` is type is not descriptive enough since too generic. `Unknown` is a good name for the variant expressing we don't know what are the dependencies, and `Known(DependencyConstraints)` would be a good  name in my opinion for the variant expressing that we could find those. Now the question is what to put before. If I call it "Tbf" (to be found ^^), it would appear in the code as:

```rust
// in type signatures
Tbf<DependencyConstraints<P,V>>

// or alternatively in signatures depending on where we put the genericity
Tbf<P,V>

// in first variant
Tbf::Unknown

// in second variant
Tbf::Known(dependency_constraints)
```

mpizenberg:
Maybe `Response` is quite fitting. In `type_aliases.rs` we could define:
```rust
enum Response<T> {
    Unknown,
    Known(T),
}
```
and then have
```rust
    fn get_dependencies(
        &self,
        package: &P,
        version: &V,
    ) -> Result<Response<DependencyConstraints<P, V>>, Box<dyn Error>>;
```

aleksator:
> The DependencyConstraints vs DependencyVersions was for your question about difference between Map<P, Range<V>> and Map<P, V>.

Regarding `DependencyVersions`: I very much like the idea of returning a struct instead of  `Map<P, V>` directly in our resolve function. Great suggestion!
I suggest naming it `SelectedDependencies` to emphasize that it's our library's output. Let's see:
```
pub fn resolve<P: Package, V: Version>(
    dependency_provider: &impl DependencyProvider<P, V>,
    package: P,
    version: impl Into<V>,
) -> Result<SelectedDepedencies<P, V>, PubGrubError<P, V>> {
```
I like it!

aleksator:
> Maybe Response is quite fitting.
> 
> ```
> enum Response<T> {
>    Unknown,
>    Known(T),
>}
>```
Response does sounds fitting, but now `Response::Unknown` gives a wrong impression that we didn't anticipate it.

What if we leave it as just `Dependencies`?
```
    fn get_dependencies(
        &self,
        package: &P,
        version: &V,
    ) -> Result<Dependencies<P, V>, Box<dyn Error>>;
```
`Dependencies::Unknown` and `Dependencies::Known(DependencyConstraints)` are read just like English sentences :)

> In type_aliases.rs we could define:

Should the module be renamed to something? `enum` is not a type alias after all.

mpizenberg:
> Result<SelectedDepedencies<P, V>, PubGrubError<P, V>>

Reads very well. But the result should also be easily handled by the caller code. Defining the type alias `type SelectedDependencies<P, V> = Map<P, V>` would do it. If using a struct instead, usage depends on how it is exposed, but would require at least one level of indirection.

mpizenberg:
@aleksator would you mind summarize the different options and ask here and on our zulip thread what people think?

aleksator:
> Reads very well. But the result should also be easily handled by the caller code. Defining the type alias type SelectedDependencies<P, V> = Map<P, V> would do it. If using a struct instead, usage depends on how it is exposed, but would require at least one level of indirection.

Yeah there are some design considerations if we want to do a newtype for that. I have some ideas (we can hide inner Map type and leave it as implementation detail for example) and would like to discuss them in a separate issue.

For now let's just do a type alias here. I'll do it shortly.

aleksator:
> @aleksator would you mind summarize the different options and ask here and on our zulip thread what people think?

Sure. I'll implement what we agreed to so far and we'll be able to discuss that directly using Github review tools.

aleksator:
TLDR: @Eh2406 @Dylan-DPC Please :+1: / :-1: new public types :smile: 

Summary of the current state:

This PR focuses on public API of our library, namely return types of `resolve` and `DependencyProvider.get_dependencies` functions.

Several new public types are defined (Let me know if the intended purpose is unclear from the docs):
```rust
/// An enum used by [DependencyProvider] that holds information about package dependencies.
/// For each [Package] there is a [Range] of concrete versions it allows as a dependency.
pub enum Dependencies<P: Package, V: Version> {
    /// Package dependencies are unavailable.
    Unknown,
    /// Container for all available package versions.
    Known(DependencyConstraints<P, V>),
}
```
```rust
/// Subtype of [Dependencies] which holds information about
/// all possible versions a given package can accept.
/// There is a difference in semantics between an empty [Map<P, Range<V>>](crate::type_aliases::Map)
/// inside [DependencyConstraints] and [Dependencies::Unknown]:
/// the former means the package has no dependencies and it is a known fact,
/// while the latter means they could not be fetched by [DependencyProvider].
pub struct DependencyConstraints<P: Package, V: Version>(Map<P, Range<V>>);
```

```rust
/// Concrete dependencies picked by the library during [resolve](crate::solver::resolve)
/// from [DependencyConstraints](crate::solver::DependencyConstraints)
pub type SelectedDependencies<P, V> = Map<P, V>;
```
This is a type alias for now, but I'd like to see if we could make it a struct and hide currently exposed Map alias in a separate PR.

As a result function signatures now look like this:
```rust
pub fn resolve<P: Package, V: Version>(
    dependency_provider: &impl DependencyProvider<P, V>,
    package: P,
    version: impl Into<V>,
) -> Result<SelectedDependencies<P, V>, PubGrubError<P, V>> {
```
```rust
    fn get_dependencies(
        &self,
        package: &P,
        version: &V,
    ) -> Result<Dependencies<P, V>, Box<dyn Error>>;
```

If you have any comments about the ergonomics - let me know.

---

We also need to decide the exact names for these public types, feel free to add suggestions.
The only suggestion left that diverges from the latest commit is:

@mpizenberg suggested to rename `Dependencies` to `Response` and change `get_dependencies` signature to:
```rust
    fn get_dependencies(
        &self,
        package: &P,
        version: &V,
    ) -> Result<Response<DependencyConstraints<P, V>>, Box<dyn Error>>;
```
The resulting method signature is pretty clear. I have a concern when doing matching on this type though: `Response::Unknown` makes it look like we didn't expect this from a `DependencyProvider`.

mpizenberg:
> @mpizenberg suggested to rename Dependencies to Response

Not "exactly" a rename since one is `Dependencies<P, V>` the other is `Response<DependencyConstraints<P, V>>` but yep, the rest is well described.

Eh2406:
> This is a type alias for now, but I'd like to see if we could make it a struct and hide currently exposed Map alias in a separate PR.

If someday we want to add any other form of output, like statistics or as a dependency graph, we could add them as methods to this new type. So that seems like a good way to future proof things.

Eh2406:
This all still looks good to me!

mpizenberg (approved):
Same here. I have some preferences, expressed in some comments, but they are just that, preferences. We have confirmation from Jacob so when you have time for it, you can merge this @aleksator .

Eh2406:
> pub struct DependencyConstraints

I like that this is a new type, as if we want to add new kinds of dependencies (like week dependencies) we can in a compatible way.

mpizenberg:
I've just re-reviewed the code here and I've changed the `DependencyConstraints` opaque struct to just a type alias due to the following reasons.

With an opaque struct, the `get_dependencies` method of any dependency provider now needs to build dependency constraints from an iterator while before, it could just be cloned as a whole. I realize that an iterator was already used in the `OfflineDependencyProvider.dependencies` method before this PR. The reason for that iterator seems to be a result of PR #14 that added benchmarks, but also used a BTreeMap for readability (since it's sorted) and thus had to convert between the BTreeMap and the HashMap of the interface. Today we are back to a HashMap and thus could simply clone the map instead of building it with an iterator.

For these reasons I don't think we should introduce an opaque struct for now. It is adding complexity where we don't need the flexibility of an opaque type for the time being. So I've changed it to a type alias instead and made the `clone()` change. When running the benchmarks on my computer, it's a perf improvement of 6% to 9%. So since it's less code and better performance, for the time beeing we'll just keep the type alias. We can reassess later if another design is needed.

mpizenberg:
Everything else than the type alias for `DependencyConstraints` was approved by everyone so I'll just go ahead and merge this. We can start another PR if needed later to change things.

aleksator:
@mpizenberg Great! I thought this PR needed some work on the `Response` type and didn't realize you were actually okay with the PR as is.
Good job benchmarking and redefining type alias!

Issue #43, opened 2020-10-18
Large files in the repository


mpizenberg:
Currently, the compressed files stored in the repository amount for 1.3 MB of data. Uncompressed, that data is partly due to the source code of this repository, and partly due to assets such as in the gh-pages branch. To understand the repartition of that data, let's have a look at the uncompressed data in the `gh-pages` and `new-benchmark` branches.

## gh-pages

That branch contains the API docs generated by `cargo doc` that we are publishing on https://pubgrub-rs.github.io/pubgrub/pubgrub/ for people to have a look at the current API docs on the dev branch without having to clone the repo. Removing the `.git/` repository, that data amounts to 2.6 MB.

It is significant compared to text files of the source code. I'd prefer if GitHub would deal with pages the same way Gitlab does, as build artifacts of CI. Unfortunately, it has to be in-tree. For this reason, I've setup the gh-pages action to always reset the `gh-pages` branch as an orphan branch and replace the commit instead of pushing new commits on top.

Another option would be to use an external service for our static pages such as documentation. I've used netlify for a while now and really like them. I see that there are actions to deploy to netlify so we could do that instead: https://github.com/marketplace/actions/netlify-deploy.

## new-benchmark

Currently, the uncompressed data amounts to 460 KB, with 184 KB being the ron example files, so a bit more than a third of it. Those files are only needed when working on optimizations, so not useful for most users and contributors. If we keep gh-pages in tree, this is marginal. If we decide to move pages outside of the repo however, it becomes non-negligible and there are different ways we could deal with it.

First, we could just remove them and share them between each other by messages / emails. _crowd screaming in the back_ Ok well not a very good idea ^^.

Alternatively, we could use a git submodule. This is another repo, which is declared in the dependencies of this repo, so if we clone with the `--recursive` option we get also retrieve the submodules. Managing the files in the submodules is a bit different than in the root repository, but if you are not familiar with git submodules, and if we choose this option at some point, I can give some advice on how to use them easily without having to manage two separate repositories on your file system.
 
Another option is using a git LFS provider. LFS (Large File Storage for git) seems to be the most adequate solution for this since file versions can be tracked (as with submodules) but in a way that is optimized for large files. However, it can be challenging to use correctly on CI, and there is a 1GB monthly bandwidth limit on the free tier in GitHub. That bandwidth can easily be exceeded with multiple forks / clones and CI misconfigurations. So LFS isn't the right fit for us I think.

## Diversity of benchmarks

[...]

PR #47, opened 2020-10-21
test: switch to criterion

Eh2406:
This switches our benchmark(s) to use criterion. Newer versions of criterion works better with slower tests, so it works out pretty well.

This automatically adds a benchmark for each file in `test-examples/`, removing the duplicated code pointed out in https://github.com/pubgrub-rs/pubgrub/pull/34#issuecomment-708426561. This also allows decoupling between the code and the  ron files. For example if in #43 we decide not to store all the ron files in tree, this will run the benchmarks on only what is in the file system.

### DIFF
--- a/benches/large_case.rs
+++ b/benches/large_case.rs
@@ -1,22 +1,38 @@
 // SPDX-License-Identifier: MPL-2.0
+use std::time::Duration;
 
-#![feature(test)]
-extern crate test;
-use test::Bencher;
+extern crate criterion;
+use self::criterion::*;
 
-use pubgrub::solver::{resolve, OfflineDependencyProvider};
+use pubgrub::solver::{resolve, DependencyProvider, OfflineDependencyProvider};
 use pubgrub::version::NumberVersion;
 
-#[cfg(feature = "serde")]
-#[bench]
-/// This is an entirely synthetic benchmark. It may not be realistic.
-fn large_case(b: &mut Bencher) {
-    let s = std::fs::read_to_string("test-examples/large_case_u16_NumberVersion.ron").unwrap();
-    let dependency_provider: OfflineDependencyProvider<u16, NumberVersion> =
-        ron::de::from_str(&s).unwrap();
+fn bench_nested(c: &mut Criterion) {
+    let mut group = c.benchmark_group("large_cases");
+    group.measurement_time(Duration::from_secs(20));
 
-    // bench
-    b.iter(|| {
-        let _ = resolve(&dependency_provider, 0, 0);
-    });
+    for case in std::fs::read_dir("test-examples").unwrap() {
+        let case = case.unwrap().path();
+
+        group.bench_function(
+            format!("{}", case.file_name().unwrap().to_string_lossy()),
+            |b| {
+                let s = std::fs::read_to_string(&case).unwrap();
+                let dependency_provider: OfflineDependencyProvider<u16, NumberVersion> =
+                    ron::de::from_str(&s).unwrap();
+                let all_versions = dependency_provider.list_available_versions(&0).unwrap();
+
+                b.iter(|| {
+                    for &n in &all_versions {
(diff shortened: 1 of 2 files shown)


aleksator:
You are reading my mind! I wanted to suggest migration to `criterion` after reading more about its use cases (seems to be right for us). And here you are with an awesome pull request! :smile:

aleksator:
I'll study criterion docs to give a proper review. They seem to have [a book](https://bheisler.github.io/criterion.rs/book/criterion_rs.html) on the subject.

Just running the benchmark I was getting warnings that 10 seconds is not enough to collect enough samples, increasing the amount in this line fixed the problem:
```
    group.measurement_time(Duration::from_secs(10));
```

So far the benchmark seemed unstable, resulting in reports like `-12%` performance, `+5%`, etc from run to run without doing changes to the code. I wonder if we are measuring something else in addition to the algorithm (like Ron deserialization), is it that unstable on its own, or is there a problem with the benchmark.
I need to learn more about `criterion` before being able to answer that.

Another thought:
We are disabling default test harness with the line `harness = false` as the docs recommend, noting that we should provide our own harness and disabling default one is required to avoid conflicts.
I also need to check how it is defined and whether we do that in this PR.

Eh2406:
Benchmarks are very sensitive to other things going on at the same time. Criterion uses statistics to accurately measure the smallest differences. I can get a reliable performance difference by changing the number of tabs open in a minimized chrome.

I think `criterion_main` builds a harness, but if so maybe we need a PR to there docs to use consistent terminology.

mpizenberg:
@aleksator (and others) I can highly recommend that talk: https://youtu.be/r-TLSBdHe1A ("Performance Matters" by Emery Berger)

mpizenberg:
Thanks for the change to criterion! I won't review thoroughly cause haven't used criterion for a long time but it looks good.

benches/large_case.rs
```diff
@@ -1,22 +1,38 @@
...
-    let dependency_provider: OfflineDependencyProvider<u16, NumberVersion> =
-        ron::de::from_str(&s).unwrap();
-
-    // bench
-    b.iter(|| {
-        let _ = resolve(&dependency_provider, 0, 0);
-    });
+fn bench_nested(c: &mut Criterion) {
+    let mut group = c.benchmark_group("large_cases");
+    group.measurement_time(Duration::from_secs(10));
```
aleksator: If I assume correctly, the time here is very machine dependent. Criterion suggests adding a couple seconds here on my laptop. It would be best if we all were able to run benchmarks without changes.

Are there any other options we could use instead of the time? Hard to suggest anything concrete without diving into the API myself, will do so later.

When I looked through the API, I saw `group.confidence_interval` (or similar), where you set up a desirable level. Maybe that or some other option would do the trick.

aleksator (changes requested):

aleksator:
> Eh2406 requested a review from aleksator 18 hours ago

Sorry for the delay. Seemed fine at a glance, but I need to dive into the docs before I could comment meaningfully. I will do so today or tomorrow. The only question I had was about `10 seconds`, left a review comment above.

Alternatively, since the change is quite simple, we could merge as is and submit changes later.
I don't want to hold this PR hostage :smile:

benches/large_case.rs
Eh2406: It is also very dependent on optimizations, as the code gets faster the time can go down. I picked 10 working on top of a bunch of wip perf commits. I will change it to 20 and we can adjust as needed.

Unfortunately, `confidence_interval` does not affect the measurement time only the amount of difference that is declared `regressed`.

aleksator: Thank you for the change :+1: 

> Unfortunately, confidence_interval does not affect the measurement time only the amount of difference that is declared regressed.

Yeah, that's unfortunate. Since we don't know any better options other than time, let's go with that.

aleksator (approved):

Issue #40, opened 2020-10-17
feat: let the dependency provider choose a package


mpizenberg:
We already let the `DependencyProvider` the responsibility of ordering versions by preference. This gives a great deal of flexibility since the provider can then change that order to implement multiple strategies such as newest versions first, oldest versions first, locked versions first, stored on file system versions first etc.

Recently, we have been discussing how to improve the performance of solver by smartly picking the next package to evaluate #32. It has been a very good improvement, but I can't stop but wonder if that should be our choice. I don't think so. And just like with versions, giving that choice back to the `DependencyProvider` would bring more flexibility I think. For correctness, we still need to pre-select the `potential` packages, but once that's done, we could let the `DependencyProvider` decide which one to choose from a provided iterator of potential packages. Having the precomputed term intersection provided by the solver as in #37 would even result in cleaner code since the `partial_solution` doesn't need to call `list_available_versions` anymore, the `DependencyProvider` can do that on its own however it wants to do it.

Eh2406:
So I experimented with adding a `pick_package` that takes a `impl Iterator<Item = (P, Term<V>)>` and returns a `Result<P, `. There is a provided implementation that does fewest versions using `list_available_versions`, but it can be overwritten. For example `OfflineDependencyProvider` can use the same heuristic with many fewer allocations in its impl. This was going great! So time to clean the api, Can we prevent a bad implementation from returning a `P` that is not in the `Iterator`? (Yes) Can we limit how much of `Term`s api we make public? (Yes)

Then I noticed that the only thing we do with the resulting `P` was pass it to `list_available_versions` and the only thing we do with that was take the first in the list. So do we want to make a more radical change to the api by doing away with `list_available_versions` and having `pick_package` return a `Result<(P, V), `?

mpizenberg:
Very interesting!

> So do we want to make a more radical change to the api by doing away with list_available_versions and having pick_package return a Result<(P, V), ?

Regrouping them in one call in the retriever is a good idea! I'd say we should go for it and see how that goes. Since it's referred to as "decision making" in [pubgrub doc](https://github.com/dart-lang/pub/blob/master/doc/solver.md#decision-making), it could be renamed `make_decision` maybe? `pick_package_version` fits well also.

mpizenberg:
[...]

Issue #21, opened 2020-10-07
Move incompatibility identifiers outside of the incompatibility struct?


mpizenberg:
I usually don't like when things know how they are stored. Having an `id` inside the incompatibility type is one such case. At the time of introducing the incompatibility store, I actually tried three possible designs for the recursive incompatibility type. (1) Id inside the struct, (2) id outside the struct, (3) using Rc or Box. (3) was eliminated when I ported the "no solution" reporting code, because having ids for this is very useful. But (1) and (2) are still viable options.

Current development has been done on top of (1) but I personally think (2) might be better, even though slightly more verbose. You can see the concrete difference between the two with the [comparison of the incompat-indices-in and incompat-indices-out branches](https://github.com/mpizenberg/pubgrub-rs/compare/incompat-indices-in...incompat-indices-out).

Eh2406:
If I understand correctly all `incompatibilities` should be a copy of an item in `incompatibility_store`. That pattern is called an arena. That suggests an approach (4) use a crate like [typed_arena](https://docs.rs/typed-arena/2.0.1/typed_arena/) so that we normally work with a `&'arena Incompatibility`, specifically so `DerivedFrom` can store `&'arena Incompatibility<'arena, P, V>` and we get rid of ids entirely.

Advantages:
- no more ids
- less memory usage (as we don't have more than one copy of an `Incompatibility`)
- less clones as `&_` impls copy

Disadvantages:
- another dependency, one with unsafe code
- an `'arena` lifetime on almost everything
- gnarly (for rust) error messages when other lifetimes get (incorrectly) entangled

And yes I did make sure this worked before suggesting it. Progress at [no-ids branch](https://github.com/pubgrub-rs/pubgrub/compare/dev...no-ids).

mpizenberg:
Oh, didn't know there was a named pattern for that :+1: 
I like the advantages and dislike the disadvantages XD. Regarding those disadvantages:
- maybe there is a safe arena version somewhere?
- if the `'arena` lifetime doesn't creep up to the exposed API that could be acceptable even though quite annoying

[...]

PR #48, opened 2020-10-21
refactor: apply Clippy suggestions

aleksator:
This PR alongside with https://github.com/pubgrub-rs/pubgrub/pull/46 fixes all Clippy suggestions in the current state of the repo.
After both are merged, Clippy can be added to CI.

I've also taken a liberty to clean up small TODOs.

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -8,7 +8,7 @@ use std::{collections::HashSet as Set, rc::Rc};
 use crate::error::PubGrubError;
 use crate::internal::assignment::Assignment::{Decision, Derivation};
 use crate::internal::incompatibility::{Incompatibility, Relation};
-use crate::internal::partial_solution::PartialSolution;
+use crate::internal::partial_solution::{DecisionLevel, PartialSolution};
 use crate::package::Package;
 use crate::report::DerivationTree;
 use crate::version::Version;
@@ -175,7 +175,7 @@ impl<P: Package, V: Version> State<P, V> {
         &mut self,
         incompat: Incompatibility<P, V>,
         incompat_changed: bool,
-        decision_level: usize,
+        decision_level: DecisionLevel,
     ) {
         self.partial_solution.backtrack(decision_level);
         if incompat_changed {
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -81,7 +81,7 @@ impl<P: Package, V: Version> State<P, V> {
                         // root_cause is guaranteed to be almost satisfied by the partial solution
                         // according to PubGrub documentation.
                         match self.partial_solution.relation(&root_cause) {
-                            Relation::AlmostSatisfied(package_almost, term) => {
+                            Relation::AlmostSatisfied((package_almost, term)) => {
                                 changed = vec![package_almost.clone()];
                                 // Add (not term) to the partial solution with incompat as cause.
                                 self.partial_solution.add_derivation(package_almost, term.negate(), root_cause);
@@ -89,7 +89,7 @@ impl<P: Package, V: Version> State<P, V> {
                             _ => return Err(PubGrubError::Failure("This should never happen, root_cause is guaranteed to be almost satisfied by the partial solution".into())),
                         }
                     }
-                    Relation::AlmostSatisfied(package_almost, term) => {
+                    Relation::AlmostSatisfied((package_almost, term)) => {
                         changed.push(package_almost.clone());
                         // Add (not term) to the partial solution with incompat as cause.
                         self.partial_solution.add_derivation(
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -154,8 +154,8 @@ impl<P: Package, V: Version> State<P, V> {
                                 id,
                                 &current_incompat,
                                 &cause,
-                                package,
-                                term,
+                                &package,
+                                &term,
                             );
                             // eprintln!("\ncause 1: {}", &current_incompat);
                             // eprintln!("cause 2: {}", &cause);
--- a/src/internal/incompatibility.rs
+++ b/src/internal/incompatibility.rs
@@ -45,6 +45,9 @@ enum Kind<P: Package, V: Version> {
     DerivedFrom(usize, usize),
 }
 
+/// A type alias for a pair of [Package] and a corresponding [Term].
+pub type PackageTerm<P, V> = (P, Term<V>);
+
 /// A Relation describes how a set of terms can be compared to an incompatibility.
 /// Typically, the set of terms comes from the partial solution.
 #[derive(Eq, PartialEq)]
@@ -54,10 +57,10 @@ pub enum Relation<P: Package, V: Version> {
     Satisfied,
     /// We say that S contradicts I
     /// if S contradicts at least one term in I.
-    Contradicted(P, Term<V>),
+    Contradicted(PackageTerm<P, V>),
     /// If S satisfies all but one of I's terms and is inconclusive for the remaining term,
     /// we say S "almost satisfies" I and we call the remaining term the "unsatisfied term".
-    AlmostSatisfied(P, Term<V>),
+    AlmostSatisfied(PackageTerm<P, V>),
     /// Otherwise, we say that their relation is inconclusive.
     Inconclusive,
 }
@@ -231,13 +234,13 @@ impl<P: Package, V: Version> Incompatibility<P, V> {
             match incompat_term.relation_with(&terms(package)) {
                 term::Relation::Satisfied => {}
                 term::Relation::Contradicted => {
-                    relation = Relation::Contradicted(package.clone(), incompat_term.clone());
+                    relation = Relation::Contradicted((package.clone(), incompat_term.clone()));
                     break;
                 }
                 term::Relation::Inconclusive => {
                     if relation == Relation::Satisfied {
                         relation =
-                            Relation::AlmostSatisfied(package.clone(), incompat_term.clone());
+                            Relation::AlmostSatisfied((package.clone(), incompat_term.clone()));
                     } else {
                         relation = Relation::Inconclusive;
                     }
--- a/src/internal/partial_solution.rs
+++ b/src/internal/partial_solution.rs
@@ -40,6 +40,11 @@ pub struct AssignmentDepth<P: Package, V: Version> {
     assignment: Assignment<P, V>,
 }
 
+pub struct AssignmentHistory<'a, P: Package, V: Version> {
+    assignment: AssignmentDepth<P, V>,
+    history: &'a [AssignmentDepth<P, V>],
+}
+
 /// The partial solution is the current state
 /// of the solution being built by the algorithm.
 /// It is composed of a succession of assignments,
@@ -204,13 +209,14 @@ impl<P: Package, V: Version> PartialSolution<P, V> {
         &self,
         incompat: &Incompatibility<P, V>,
     ) -> (Assignment<P, V>, DecisionLevel, DecisionLevel) {
-        let (
-            AssignmentDepth {
-                decision_level: satisfier_level,
-                assignment: satisfier,
-            },
-            previous_assignments,
-        ) = Self::find_satisfier(incompat, self.history.as_slice())
+        let AssignmentHistory {
+            assignment:
+                AssignmentDepth {
+                    decision_level: satisfier_level,
+                    assignment: satisfier,
+                },
+            history: previous_assignments,
+        } = Self::find_satisfier(incompat, self.history.as_slice())
             .expect("We should always find a satisfier if called in the right context.");
         let previous_satisfier_level =
             Self::find_previous_satisfier(incompat, &satisfier, previous_assignments);
@@ -223,7 +229,7 @@ impl<P: Package, V: Version> PartialSolution<P, V> {
     fn find_satisfier<'a>(
         incompat: &Incompatibility<P, V>,
         history: &'a [AssignmentDepth<P, V>],
-    ) -> Option<(AssignmentDepth<P, V>, &'a [AssignmentDepth<P, V>])> {
+    ) -> Option<AssignmentHistory<'a, P, V>> {
         Self::find_satisfier_helper(incompat, Self::new_accum_satisfied_from(incompat), history)
     }
 
@@ -242,10 +248,15 @@ impl<P: Package, V: Version> PartialSolution<P, V> {
         let mut accum_satisfied = Self::new_accum_satisfied_from(incompat);
         accum_satisfied.insert(package, (is_satisfied, satisfier_term));
         // Search previous satisfier.
-        Self::find_satisfier_helper(incompat, accum_satisfied, previous_assignments)
-            .map_or(DecisionLevel(1), |assignment_history| {
-                assignment_history.0.decision_level.max(DecisionLevel(1))
-            })
+        Self::find_satisfier_helper(incompat, accum_satisfied, previous_assignments).map_or(
+            DecisionLevel(1),
+            |assignment_history| {
+                assignment_history
+                    .assignment
+                    .decision_level
+                    .max(DecisionLevel(1))
+            },
+        )
     }
 
     fn new_accum_satisfied_from(incompat: &Incompatibility<P, V>) -> Map<P, (bool, Term<V>)> {
@@ -262,7 +273,7 @@ impl<P: Package, V: Version> PartialSolution<P, V> {
         incompat: &Incompatibility<P, V>,
         accum_satisfied: Map<P, (bool, Term<V>)>,
         all_assignments: &'a [AssignmentDepth<P, V>],
-    ) -> Option<(AssignmentDepth<P, V>, &'a [AssignmentDepth<P, V>])> {
+    ) -> Option<AssignmentHistory<'a, P, V>> {
         let mut accum_satisfied = accum_satisfied;
         for (idx, assignment_depth) in all_assignments.iter().enumerate() {
             let package = assignment_depth.assignment.package();
@@ -282,13 +293,13 @@ impl<P: Package, V: Version> PartialSolution<P, V> {
             // Check if we have found the satisfier
             // (all booleans in accum_satisfied are true).
             if *is_satisfied && accum_satisfied.iter().all(|(_, (satisfied, _))| *satisfied) {
-                return Some((
-                    AssignmentDepth {
+                return Some(AssignmentHistory {
+                    assignment: AssignmentDepth {
                         decision_level: assignment_depth.decision_level,
                         assignment: assignment_depth.assignment.clone(),
                     },
-                    &all_assignments[0..idx],
-                ));
+                    history: &all_assignments[0..idx],
+                });
             }
         }
         None
--- a/src/internal/partial_solution.rs
+++ b/src/internal/partial_solution.rs
@@ -17,6 +17,12 @@ use crate::{
     solver::DependencyProvider,
 };
 
+#[derive(Clone)]
+pub struct AssignmentDepth<P: Package, V: Version> {
+    decision_level: usize,
+    assignment: Assignment<P, V>,
+}
+
 /// The partial solution is the current state
 /// of the solution being built by the algorithm.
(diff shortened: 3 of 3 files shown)


Eh2406:
Lots of little things that each look good. This overall looks very good. I will try to revue one commit at time to ensure I like each thing.

aleksator:
> Lots of little things that each look good. This overall looks very good. I will try to revue one commit at time to ensure I like each thing.

Yeah they are all atomic. Let me know if this is too much and I should pull some things out of this PR like I did with #46.

src/internal/partial_solution.rs
```diff
@@ -59,6 +59,8 @@ pub struct PartialSolution<P: Package, V: Version> {
     memory: Memory<P, V>,
 }
 
+pub type PackageTerm<P, V> = (P, Term<V>);
```
Eh2406: Why is this here instead of in `incompatibility.rs`?

aleksator: No reason besides the return type of functions I was trying to simplify was in `partial_solution.rs`. I will move it to `incompatibility.rs`, it does make more sense there since it's used in `Relation` enum.

Are you fine with it being a type alias rather than a separate type? I have a commit where I had it extracted into a struct, let me know if you would prefer that :smile:

Eh2406 (approved):
Just the one nit.

src/internal/partial_solution.rs
Eh2406: I have no methods to add to it so type alias works.

Dylan-DPC-zz (approved):

src/internal/partial_solution.rs
aleksator: Alright. Thank you spotting this! I changed the last commit.

aleksator:
I've changed the last commit (moved `PackageTerm` to `incompatibility.rs`).

Would you prefer to have all of them squashed or kept separate? (squash merge vs rebase merge)

Eh2406:
The commits are atomic so I'd be ok with rebase, but in general @mpizenberg prefers squash. So I'll leave it up to him.

src/internal/core.rs
```diff
@@ -18,20 +18,14 @@ use crate::version::Version;
...
-
-    /// TODO: remove pub.
-    pub incompatibilities: Rc<Vec<Incompatibility<P, V>>>,
-
+    incompatibilities: Rc<Vec<Incompatibility<P, V>>>,
     /// Partial solution.
-    /// TODO: remove pub.
-    pub partial_solution: PartialSolution<P, V>,
-
+    pub(crate) partial_solution: PartialSolution<P, V>,
```
mpizenberg: I'm not a big fan of the `pub(crate)` thing as you may have noticed. I generally think that if I need something to be pub, everybody needs it. And I think the round of `pub(crate)` we did on the `term` module because I didn't spend enought time on the reporting API will byte us for #40 .

What I meant here by "remove pub" was "the outside world (outside this module)" should not know how this is done internally. And in fact #21 solves already two of those pubs by re-organizing indices. So we should maybe leave those out of this PR

mpizenberg: In addition, the `pub(crate)` here is equivalent to the `pub` anyway since `internal::` is not `pub`.

src/range.rs
```diff
@@ -183,34 +185,38 @@ impl<V: Version> Range<V> {
                 }
 
                 // Right contains an infinite interval:
-                (Some((l1, Some(l2))), Some((r1, None))) => {
-                    if l2 < r1 {
+                (Some((l1, Some(l2))), Some((r1, None))) => match l2.cmp(r1) {
+                    Ordering::Less => {
```
mpizenberg: This feels a lot less readable to me. What is Clippy saying about this one?

src/internal/partial_solution.rs
```diff
@@ -77,13 +86,23 @@ impl<P: Package, V: Version> PartialSolution<P, V> {
...
         *self = Self::from_assignments(
             std::mem::take(&mut self.history)
                 .into_iter()
                 .take(pos + 1)
-                .map(|(_, a)| a),
+                .map(
+                    |AssignmentDepth {
+                         decision_level: _,
+                         assignment,
+                     }| assignment,
```
mpizenberg: I'm not certain, but I believe you can write something like `|AssignmentDepth { assignment, .. }| `

mpizenberg: same for the change above on rposition

src/internal/partial_solution.rs
```diff
@@ -17,6 +17,12 @@ use crate::{
     solver::DependencyProvider,
 };
 
+#[derive(Clone)]
+struct AssignmentDepth<P: Package, V: Version> {
```
mpizenberg: Maybe `AssignmentWithDecisionLevel` would be more adequate even though a bit redundant with its content. `AssignmentDepth` sounds like something it's not. Is it the fact that `usize` does not convey a lot of meaning that bothers you? Maybe a type alias would be enough here?

src/internal/partial_solution.rs
```diff
@@ -17,9 +17,26 @@ use crate::{
     solver::DependencyProvider,
 };
 
+#[derive(Debug, Copy, Clone, Ord, PartialOrd, Eq, PartialEq)]
+pub struct DecisionLevel(u32);
```
mpizenberg, line 22: Alright just in the following commit :+1: 
I feel like a type alias would be enough though.

mpizenberg: But ok for the newtype

src/internal/partial_solution.rs
```diff
@@ -214,13 +219,14 @@ impl<P: Package, V: Version> PartialSolution<P, V> {
...
         incompat: &Incompatibility<P, V>,
     ) -> (Assignment<P, V>, DecisionLevel, DecisionLevel) {
-        let (
-            AssignmentDepth {
-                decision_level: satisfier_level,
-                assignment: satisfier,
-            },
-            previous_assignments,
-        ) = Self::find_satisfier(incompat, self.history.as_slice())
+        let AssignmentHistory {
```
mpizenberg: These really should not be put together though in my opinion. They have no other reason to be together than being two outputs of a function.

mpizenberg: The logic inside the satisfier isn't something very stable either. so having the flexibility of easily changing the return types there is important.

src/internal/incompatibility.rs
```diff
@@ -44,6 +44,9 @@ enum Kind<P: Package, V: Version> {
     DerivedFrom(usize, usize),
 }
 
+/// A type alias for a pair of [Package] and a corresponding [Term].
+pub type PackageTerm<P, V> = (P, Term<V>);
```
mpizenberg, line 49: Ok

mpizenberg: But I don't see much value. `(P, Term<V>)` seems pretty straightforward

mpizenberg: Sometimes adding a level of indirection is not necessary

mpizenberg (commented):
I have mixed feelings on a lot of things in this PR. I'm basically ok with all Clippy suggestions but not ok with some names used and some changes. Detailed in review items.

Eh2406:
If there are a few lintes that are holding this PR back, then we should split them out and merge the clear cut parts.

src/range.rs
aleksator:
```
warning: `if` chain can be rewritten with `match`
   --> src/range.rs:202:21
    |
202 | /                     if r2 < l1 {
203 | |                         right = right_iter.next();
204 | |                     } else if r2 == l1 {
205 | |                         segments.extend(right_iter.cloned());
...   |
211 | |                         break;
212 | |                     }
    | |_____________________^
    |
    = help: Consider rewriting the `if` chain to use `cmp` and `match`.
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#comparison_chain
```

aleksator: I like it because it shows the intent: you are matching over all possible comparison values.

There's a principle I use while coding that can be formulated as "use the least powerful construct that does the job".

Like when you are iterating over values in a list, `foreach` is preferred over `for (int i = 0; i < len; ++i)` since the latter is more powerful construct, using that indicates that you probably want to manipulate indexes somehow. If you aren't doing that, `foreach` says "simple iteration".

Same deal here: match over possible comparison values says exactly that. If chain is more powerful, you would use it if you wanted to do something additional in the conditions and you have to check whether that is the case. It is not the case here, hence we can use less powerful tool and save ourselves condition checking.

src/internal/partial_solution.rs
aleksator: Or even better, we can just do `.map(|assignment_depth| assignment_depth.assignment)`. Thanks for bringing my attention to these lines.

src/internal/partial_solution.rs
aleksator: > Maybe AssignmentWithDecisionLevel would be more adequate even though a bit redundant with its content. 

I can rename that no problem. A bit verbose, but can't come up with anything better. `DecidedAssignment`? :smile: 

> Is it the fact that usize does not convey a lot of meaning that bothers you?

I like using type system to indicate meaning and make it easily greppable. Allows for easy type changes as well: say we want to reduce `u32` to `u16`, it would simplify doing such refactoring a lot.

> Maybe a type alias would be enough here?

Let's see, type alias also provides some of those benefits. There is a problem that you could forget to use it, so searching usages becomes less reliable. Even if you never forget new instance creation becomes hidden (just 0/1 in code). It also can not guard against adding that and something else.
As for benefits... Hm... Both are fully optimized away, so I don't know. Why do you like it more?

aleksator: > `DecidedAssignment`

`AppliedAssignment` as another option.

> Maybe a type alias would be enough here?

Wait, I see another thread talking about type alias/newtype for `DecisionLevel`. Were you asking about the whole `AssignmentDepth` here? If so, I got confused by "Is it the fact that usize does not convey a lot of meaning that bothers you?" question.

Let's see... It keeps return types in functions the same as newtype, quite readable. I'm only concerned about matching on it: you'd have to look up the type definition and do tuple destructuring in that case, while newtype allows field access by name.

src/internal/partial_solution.rs
aleksator: > The logic inside the satisfier isn't something very stable either. so having the flexibility of easily changing the return types there is important.

Doesn't having a grouping type make it more easily changeable? You could extend the struct/change inner fields without touching type signature.

> These really should not be put together though in my opinion. They have no other reason to be together than being two outputs of a function.

True, the purpose of the change was to provide some context on those return types.
Before:
```
 ) -> Option<(AssignmentDepth<P, V>, &'a [AssignmentDepth<P, V>])>
```
After:
```
 ) -> Option<AssignmentHistory<'a, P, V>>
```
I think looking at type signature of "before" version does not make it clear why those are separate and why AssignmentDepth is not inside the slice, and what that slice means.

Would you be okay with making this a type alias if you think a struct definition doesn't pull it's weight here?
I would like having one of those as a return type.

src/internal/incompatibility.rs
aleksator: I admit this was just to fix a clippy lint, both before and after are quite easy to read IMO. I weighted the options of adding a suppress for the lint, and defining a type alias for PackageTerm and went with the latter.

The reason was we already have this meaning implicitly defined in our library: we have `package_terms` map that we iterate over, and the iterator returns this tuple. So it didn't feel like a new thing you have to look up a definition of.

If you disagree with the reasoning, let me know and I will change it to a lint suppression instead.

aleksator:
> If there are a few lintes that are holding this PR back, then we should split them out and merge the clear cut parts.

I think what was holding this PR back was me not working through the comments until today :)

We could merge a few uncontroversial changes. There are a lot of discussion points though, and I don't know how to move them to another PR. Gitlab has "resolve in another issue" button, but I don't see anything similar here.
Does anyone know?

Alternatively, we could fix everything in this PR since I have time this weekend to correct everything.

src/internal/core.rs
aleksator: Will revert.

src/internal/partial_solution.rs
aleksator: Changed both in the latest commit "refactor: shorten AssignmentDepth matching".

src/internal/partial_solution.rs
aleksator: > Alright just in the following commit +1

Sorry, I don't follow what you mean here :)

aleksator:
> We could merge a few uncontroversial changes. There are a lot of discussion points though, and I don't know how to move them to another PR. Gitlab has "resolve in another issue" button, but I don't see anything similar here.
Does anyone know?

Just realized I could do the opposite: extract approved parts into a separate PR and leave all the questionable items here. Let me do just that.

src/internal/partial_solution.rs
mpizenberg: Just because I was reviewing one commit after the other. In the previous commit, I suggested giving a type alias for decision level and here in the following commit you created a type for that. Newtype is good :+1:

src/internal/partial_solution.rs
mpizenberg: > Doesn't having a grouping type make it more easily changeable?

Yeah you're right on this one XD.

> I think looking at type signature of "before" version does not make it clear why those are separate and why AssignmentDepth is not inside the slice, and what that slice means.

I'll try to think a bit more about that return type and come back with some ideas (or not)

src/internal/incompatibility.rs
mpizenberg: Alright, I guess both are ok. Choose what you prefer :)

mpizenberg (commented):
Ok so my last issues in this PR are with the `AssignmentDepth` and `AssignmentHistory` types. I'll try to spend some time of this tomorrow or Monday.

src/internal/partial_solution.rs
aleksator: Ah, and I misunderstood that you were arguing against adding a newtype or type alias for `decision_level` and wanted to keep a plain integer type. Got it now :+1:

src/internal/incompatibility.rs
aleksator: Thanks. Let's keep a type alias then :)

aleksator:
Summary:

`AssignmentDepth` could be named better:
- `AssignmentWithDecisionLevel`
- `DecidedAssignment`
- `AppliedAssignment`

`AssignmentHistory` - alright name, but questionable grouping.

mpizenberg:
Regarding `AssignmentDepth`. The main purpose of this struct is to give a name to the integer recording the decision level associated with an Assignement. For context, initially the decision level was a field inside `Assignment` that used to be a struct. Then I extracted it out of the `Assignment` since an assignment doesn't have to know when it was created. This enabled some simplifications in the code structure, and `Assignment` switched from a struct to an enum.

When the decision level was extracted, it was then stored side-by-side with the assignment in a pair `(usize, Assignment)` inside a vec called `history` being the ordered history of all assignments created for the partial solution.

In this PR, we agreed to replace the `usize` (that should have been a `u32`) by the newtype `DecisionLevel(u32)` which is more appropriate. As such, I think the pair `(DecisionLevel, Assignment)` is perfect for the generic type for `history`. So I'd go for `history: Vec<(DecisionLevel, Assignment<P, V>)>`. But if you really think a struct would be better, I'd suggest something like `AssignmentRecord`. History is after all a big record of "dated" events. Here the "date" is the decision level.

mpizenberg:
The `find_satisfier` and `find_previous_satisfier` methods have had fluctuating interfaces. At the beginning they returned the same type, but after few refactors, each one of those was stripped of unneeded outputs and eventually, only the `find_satisfier_and_previous_satisfier_level` method was publicly exposed. But I kept the `find_satisfier` and `find_previous_satisfier` names mainly because naming is hard XD.

In order to find a good name, I'll try summarizing what those functions do. The purpose of `find_satisfier` is to find the oldest decision that caused the current conflict we are trying to solve. Let's say we order assignements in the history of our partial solution like this:

```txt
decision 1, derivation 1a, derivation 1b, decision 2, derivation 2a, decision 3, derivation 3a
```

And let's say we have the incompatibility `{p: 1, q: 1}` that was just created and that we just discovered is satisfied by our current partial solution. We now have to backtrack, but the question is how much? And the purpose of `find_satisfier` is to find out which assignments are irrelevant and what are the oldest assignments relevant to this incompatibility. It turns out for example that decision 1 was `p: 1` and derivation 2a was `q: 1`. If all assignments between decision 1 and derivation 2a are irrelevant, then derivation 2a is our satisfier. The function would then return `(DecisionLevel(2), derivation_2a)` for the satisfier. It also returns the slice of all assignements previous to this one so `&[decision_1 .. decision_2]` if you allow the inaccurate notation.

If we were to return those two things in a struct (name to be found), accurate (verbose) names of the fields could be

```rust
struct Tbf {
    satisfier: (DecisionLevel, Assignment<P,V>),
    history_previous_to_satisfier: &[(DecisionLevel, Assignment<P,V>)],
}
```

where the pair above could be replaced by the `AssignmentRecord` variation we discussed in a previous comment.

Eh2406:
What is this PR waiting on, is it me? If so to share my opinion on what questions?

mpizenberg:
I gave my point of view on the choices for the types `AssignmentDepth` and `AssignmentHistory` in my last two comments. Would love to have an answer from alex or you on thos points

Eh2406:
> DecisionLevel(u32)

Big +1, great improvement.

> history: Vec<(DecisionLevel, Assignment<P, V>)>

Seames good to me. If we need to hide the type more I'd rather see a `history: History<P, V>` over a `history: Vec<AssignmentRecord<P, V>>`, as I could imagine other ways to store this... Edit: removed as it is off topic, and if we are adding a `AssignmentRecord` for other reason then use it.

> find_satisfier_and_previous_satisfier_level 

This section of the code seams under polished in general. (To be clear that is not a complaint. make it work make it clean, make it fast. I never got to the the first one.) As such I am good with incremental improvements. If `AssignmentRecord` makes things clearer for now, wonderfull. If we need to turn that lint off for those functions that is also good.
If we are pulling it out into a struct mybee `SatisfierContext`, it is used all at the bottom of the one file so it doesn't have to be perfectly memorable.

mpizenberg:
Thanks for the feedback! @aleksator when you have time for that, pick up whatever you like from my last comments (or not) and you can merge this. As Jacob says, it's not the most polished part of the code and will eventually need some refinement later anyway.

Eh2406:
> ```rust
> struct Tbf {
>     satisfier: (DecisionLevel, Assignment<P,V>),
>     history_previous_to_satisfier: &[(DecisionLevel, Assignment<P,V>)],
> }
> ```
If I am reading the code wright `satisfier` is always `all_assignments[history_previous_to_satisfier.len()]` and `history_previous_to_satisfier` is always `all_assignments[0..history_previous_to_satisfier.len()]`, if that is true then it can be simplified by returning the slice with `satisfier` being the last element, or even just returning the index of `satisfier`.
Edit: not to say that that should or has to happen in this PR, just a thing I noticed.

mpizenberg:
> If I am reading the code wright satisfier is always all_assignments[history_previous_to_satisfier.len()] and history_previous_to_satisfier is always all_assignments[0..history_previous_to_satisfier.len()]

That's right. We could just return the index, or return the satisfier and the index. Haven't tried those options so I don't know if the code would be clearer or not that way.

mpizenberg:
Renamed `AssignmentDepth` to `DatedAssignment` since "dated" fits well with "history".

Changed `AssignmentHistory` struct to the more verbose but more adequate:

```rust
pub struct SatisfierAndPreviousHistory<'a, P: Package, V: Version> {
    satisfier: DatedAssignment<P, V>,
    previous_history: &'a [DatedAssignment<P, V>],
}
```

mpizenberg:
All changes here were already validated except the two small name changes I just did. They fit better than the previous names but if they are still not perfect, we can change those later. For the time being I'll just go ahead and merge this.

Issue #44, opened 2020-10-18
Benchmark on crates.io index


mpizenberg:
It is a bit far from now, but one thing that would be great to optimize "real-world" performances would be to use "real-world" packages registries. For example, we could build the index of crates.io and make a benchmark that would solve dependencies for the last version of every crate. It would take into account the performances of both the pubgrub implementation and a dependency retriever implementation.

Solving dependencies of crates.io is not possible for the time being though since we don't handle all the possibilities that cargo enables. But we could probably simplify the index in ways that enables benchmarking even if we don't support everything yet. Features could for example be removed from the index. We could consider only versions below 1.0 if we don't want to handle multiple major. Remove packages using pre-releases or replacements etc.

Alternatively, we could start with simpler indexes from other programming languages. I think for example that we could already solve elm packages. I don't know about others.

Ultimate test would probably be solving dependencies of all NPM packages ahah. That could be a blog post XD.

Eh2406:
I think python dependencies do ~~not need any unsupported features~~ pre-release and atypical ranges. But traditionally you need to run `setup.py` to get the dependencies. 🤦 What could go wrong with a turing complete way to specify dependencies?

Anyway conda is a better package manager for python, and it does have an index file.
conda index [link 4.8M](https://repo.anaconda.com/pkgs/free/win-64/)
conda-forge index [link 56.6 MB](https://conda.anaconda.org/conda-forge/win-64/repodata.json)
and the [dependency syntax](https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html#package-match-specifications). 
and a [blog post](https://www.anaconda.com/blog/understanding-and-improving-condas-performance) about why there dependency resolution is so slow. Apparently there solver is a pluggable component so if things go really well...

mpizenberg:
Registry of all elm packages since elm 0.14, which represents 11079 distinct package versions + 11 elm versions: 
[elm-packages.zip](https://github.com/pubgrub-rs/pubgrub/files/5541663/elm-packages.zip) (contains a 4.1Mb `elm-packages.ron` file)

I have done some rudimentary cleaning but not tried to solve dependencies yet, might need some additional cleaning.

Eh2406:
That is wonderful! I started by using dev and looping over all versions and running the SAT test against it. All passed!
[...]

PR #50, opened 2020-10-23
feat!: let DependencyProvider prioritize dependencies

### LINKED ISSUE #18: Don't require owned dependencies from provider
Is there a way to avoid the clones and allocation in https://github.com/mpizenberg/elm-test-rs/blob/31596f874a6e9340350de3e7c783a5cde9ca396b/src/solver.rs#L439? In Cargo we use `Rc<Vec<Summary>>` https://github.com/rust-lang/cargo/blob/07162dbaa84aaad97d6f3be5069265e784179bcd/src/cargo/core/resolver/dep_cache.rs#L35, so that a cache hit is  and Rc clone instead of a Vec clone. I wonder if there is a trate bound that works for the algorithm and allows returning `Vec<>`, `Rc<Vec<_>>`, `im::Vector<_>`, or even a leaked `&'static [_]`?

Originally brought up in https://github.com/mpizenberg/pubgrub-rs/pull/15#issuecomment-704308551

Eh2406:
This follows up on #40. 

The first commit explores `pick_package` approach as described in https://github.com/pubgrub-rs/pubgrub/issues/40#issuecomment-714865685.
The second commit explores the `make_decision` approach.
The third commit uses a BTreeMap to remove Hash req on `OfflineDependencyProvider`

Beadback is definitely appreciated.

Edit: somewhat accidentally closes #18

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,15 +77,15 @@
 //! ```
 //! # use pubgrub::solver::{DependencyProvider, Dependencies};
 //! # use pubgrub::version::SemanticVersion;
+//! # use pubgrub::range::Range;
+//! # use pubgrub::type_aliases::Map;
 //! # use std::error::Error;
+//! # use std::borrow::Borrow;
 //! #
 //! # struct MyDependencyProvider;
 //! #
 //! impl DependencyProvider<String, SemanticVersion> for MyDependencyProvider {
-//!     fn list_available_versions(
-//!         &self,
-//!         package: &String
-//!     ) -> Result<Vec<SemanticVersion>, Box<dyn Error>> {
+//!     fn choose_package_version<T: Borrow<String>, U: Borrow<Range<SemanticVersion>>>(&self,packages: impl Iterator<Item=(T, U)>) -> Result<(T, Option<SemanticVersion>), Box<dyn Error>> {
 //!         unimplemented!()
 //!     }
 //!
@@ -98,11 +98,20 @@
 //!     }
 //! }
 //! ```
+//!
 //! The first method
-//! [list_available_versions](crate::solver::DependencyProvider::list_available_versions)
-//! should return all available versions of a package.
-//! The second method
-//! [get_dependencies](crate::solver::DependencyProvider::get_dependencies)
+//! [choose_package_version](crate::solver::DependencyProvider::choose_package_version)
+//! chooses a package and available version compatible with the provided options.
+//! A helper function
+//! [choose_package_with_fewest_versions](crate::solver::choose_package_with_fewest_versions)
+//! is provided for convenience
+//! in cases when lists of available versions for packages are easily obtained.
+//! The strategy of that helper function consists in choosing the package
+//! with the fewest number of compatible versions to speed up resolution.
+//! But in general you are free to employ whatever strategy suits you best
+//! to pick a package and a version.
+//!
+//! The second method [get_dependencies](crate::solver::DependencyProvider::get_dependencies)
 //! aims at retrieving the dependencies of a given package at a given version.
 //! Returns [None] if dependencies are unknown.
 //!
--- a/src/solver.rs
+++ b/src/solver.rs
@@ -65,14 +65,13 @@
 //! to satisfy the dependencies of that package and version pair.
 //! If there is no solution, the reason will be provided as clear as possible.
 
-use std::collections::BTreeSet as Set;
+use std::borrow::Borrow;
+use std::collections::{BTreeMap, BTreeSet as Set};
 use std::error::Error;
-use std::hash::Hash;
 
 use crate::error::PubGrubError;
 use crate::internal::core::State;
 use crate::internal::incompatibility::Incompatibility;
-use crate::internal::partial_solution::PartialSolution;
 use crate::package::Package;
 use crate::range::Range;
 use crate::type_aliases::{Map, SelectedDependencies};
@@ -95,43 +94,56 @@ pub fn resolve<P: Package, V: Version>(
 
         state.unit_propagation(next)?;
 
-        // Pick the next package.
-        let (p, term) = match state.partial_solution.pick_package(dependency_provider)? {
-            None => {
-                return state.partial_solution.extract_solution().ok_or_else(|| {
-                    PubGrubError::Failure(
-                        "How did we end up with no package to choose but no solution?".into(),
-                    )
-                })
-            }
-            Some(x) => x,
-        };
-        next = p.clone();
-        let available_versions =
-            dependency_provider
-                .list_available_versions(&p)
-                .map_err(|err| PubGrubError::ErrorRetrievingVersions {
-                    package: p.clone(),
-                    source: err,
-                })?;
+        let potential_packages = state.partial_solution.potential_packages();
+        if potential_packages.is_none() {
+            drop(potential_packages);
+            // The borrow checker did not like using a match on potential_packages.
+            // This `if ... is_none ... drop` is a workaround.
+            // I believe this is a case where Polonius could help, when and if it lands in rustc.
+            return state.partial_solution.extract_solution().ok_or_else(|| {
+                PubGrubError::Failure(
+                    "How did we end up with no package to choose but no solution?".into(),
+                )
+            });
+        }
+        let decision = dependency_provider
+            .choose_package_version(potential_packages.unwrap())
+            .map_err(PubGrubError::ErrorChoosingPackageVersion)?;
+        next = decision.0.clone();
 
         // Pick the next compatible version.
-        let v = match PartialSolution::<P, V>::pick_version(&available_versions[..], &term) {
+        let v = match decision.1 {
             None => {
+                let term = state
+                    .partial_solution
+                    .term_intersection_for_package(&next)
+                    .expect("a package was chosen but we don't have a term.")
+                    .clone();
                 state.add_incompatibility(|id| {
-                    Incompatibility::no_versions(id, p.clone(), term.clone())
+                    Incompatibility::no_versions(id, next.clone(), term.clone())
                 });
                 continue;
             }
             Some(x) => x,
         };
+        if !state
+            .partial_solution
+            .term_intersection_for_package(&next)
+            .expect("a package was chosen but we don't have a term.")
+            .contains(&v)
+        {
+            return Err(PubGrubError::ErrorChoosingPackageVersion(
+                "choose_package_version picked an incompatible version".into(),
+            ));
+        }
 
         if added_dependencies
-            .entry(p.clone())
+            .entry(next.clone())
             .or_default()
             .insert(v.clone())
         {
             // Retrieve that package dependencies.
+            let p = &next;
             let dependencies =
                 match dependency_provider
                     .get_dependencies(&p, &v)
@@ -182,11 +194,13 @@ pub fn resolve<P: Package, V: Version>(
                     "Root package depends on itself at a different version?".into(),
                 ));
             }
-            state.partial_solution.add_version(p, v, &dep_incompats);
+            state
+                .partial_solution
+                .add_version(p.clone(), v, &dep_incompats);
         } else {
             // `dep_incompats` are already in `incompatibilities` so we know there are not satisfied
             // terms and can add the decision directly.
-            state.partial_solution.add_decision(p, v);
+            state.partial_solution.add_decision(next.clone(), v);
         }
     }
 }
@@ -212,10 +226,35 @@ pub type DependencyConstraints<P, V> = Map<P, Range<V>>;
 /// Trait that allows the algorithm to retrieve available packages and their dependencies.
 /// An implementor needs to be supplied to the [resolve] function.
 pub trait DependencyProvider<P: Package, V: Version> {
-    /// Lists available versions for a given package.
-    /// The strategy of which version should be preferably picked in the list of available versions
-    /// is implied by the order of the list: first version in the list will be tried first.
-    fn list_available_versions(&self, package: &P) -> Result<Vec<V>, Box<dyn Error>>;
+    /// [Decision making](https://github.com/dart-lang/pub/blob/master/doc/solver.md#decision-making)
+    /// is the process of choosing the next package
+    /// and version that will be appended to the partial solution.
+    /// Every time such a decision must be made,
+    /// potential valid packages and version ranges are preselected by the resolver,
+    /// and the dependency provider must choose.
+    ///
+    /// The strategy employed to choose such package and version
+    /// cannot change the existence of a solution or not,
+    /// but can drastically change the performances of the solver,
+    /// or the properties of the solution.
+    /// The documentation of Pub (PubGrub implementation for the dart programming language)
+    /// states the following:
+    ///
+    /// > Pub chooses the latest matching version of the package
+    /// > with the fewest versions that match the outstanding constraint.
+    /// > This tends to find conflicts earlier if any exist,
+    /// > since these packages will run out of versions to try more quickly.
+    /// > But there's likely room for improvement in these heuristics.
+    ///
+    /// A helper function [choose_package_with_fewest_versions] is provided to ease
+    /// implementations of this method if you can produce an iterator
+    /// of the available versions in preference order for any package.
+    ///
+    /// Note: the type `T` ensures that this returns an item from the `packages` argument.
+    fn choose_package_version<T: Borrow<P>, U: Borrow<Range<V>>>(
+        &self,
+        potential_packages: impl Iterator<Item = (T, U)>,
+    ) -> Result<(T, Option<V>), Box<dyn Error>>;
 
     /// Retrieves the package dependencies.
     /// Return [Dependencies::Unknown] if its dependencies are unknown.
@@ -235,15 +274,45 @@ pub trait DependencyProvider<P: Package, V: Version> {
     }
 }
 
+/// This is a helper function to make it easy to implement
+/// [DependencyProvider::choose_package_version].
+/// It takes a function `list_available_versions` that takes a package and returns an iterator
+/// of the available versions in preference order.
+/// The helper finds the package from the `packages` argument with the fewest versions from
+/// `list_available_versions` contained in the constraints. Then takes that package and finds the
+/// first version contained in the constraints.
+pub fn choose_package_with_fewest_versions<P: Package, V: Version, T, U, I, F>(
+    list_available_versions: F,
+    potential_packages: impl Iterator<Item = (T, U)>,
+) -> (T, Option<V>)
+where
+    T: Borrow<P>,
+    U: Borrow<Range<V>>,
+    I: Iterator<Item = V>,
+    F: Fn(&P) -> I,
+{
+    let count_valid = |(p, range): &(T, U)| {
+        list_available_versions(p.borrow())
+            .filter(|v| range.borrow().contains(v.borrow()))
+            .count()
+    };
+    let (pkg, range) = potential_packages
+        .min_by_key(count_valid)
+        .expect("potential_packages gave us an empty iterator");
+    let version =
+        list_available_versions(pkg.borrow()).find(|v| range.borrow().contains(v.borrow()));
+    (pkg, version)
+}
+
 /// A basic implementation of [DependencyProvider].
 #[derive(Debug, Clone, Default)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(transparent))]
-pub struct OfflineDependencyProvider<P: Package, V: Version + Hash> {
-    dependencies: Map<P, Map<V, DependencyConstraints<P, V>>>,
+pub struct OfflineDependencyProvider<P: Package, V: Version> {
+    dependencies: Map<P, BTreeMap<V, DependencyConstraints<P, V>>>,
 }
 
-impl<P: Package, V: Version + Hash> OfflineDependencyProvider<P, V> {
+impl<P: Package, V: Version> OfflineDependencyProvider<P, V> {
     /// Creates an empty OfflineDependencyProvider with no dependencies.
     pub fn new() -> Self {
         Self {
@@ -282,12 +351,10 @@ impl<P: Package, V: Version + Hash> OfflineDependencyProvider<P, V> {
         self.dependencies.keys()
     }
 
-    /// Lists versions of saved packages.
+    /// Lists versions of saved packages in sorted order.
     /// Returns [None] if no information is available regarding that package.
-    fn versions(&self, package: &P) -> Option<Set<V>> {
-        self.dependencies
-            .get(package)
-            .map(|k| k.keys().cloned().collect())
+    pub fn versions(&self, package: &P) -> Option<impl Iterator<Item = &V>> {
+        self.dependencies.get(package).map(|k| k.keys())
     }
 
     /// Lists dependencies of a given package and version.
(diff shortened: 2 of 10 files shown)


src/solver.rs
```diff
@@ -187,6 +208,30 @@ pub trait DependencyProvider<P: Package, V: Version> {
         version: &V,
     ) -> Result<Option<Map<P, Range<V>>>, Box<dyn Error>>;
 
+    /// Chooses which package the [resolve] should decide on next.
+    /// The default implementation picks the package that has the fewest versions contained in the constraints.
+    /// Override to avoid the calls to [list_available_versions], or to use a different heuristic.
+    /// Note: the type `T` ensures that this returns an item from the `packages` argument.
```
mpizenberg: That's a really cool trick!

src/solver.rs
```diff
@@ -343,9 +308,26 @@ impl<P: Package, V: Version + Hash> DependencyProvider<P, V> for OfflineDependen
 
             if key < min_key {
                 min_key = key;
-                out = Some(p);
+                package = Some((p, term));
             }
         }
-        Ok(out.expect("get_dependencies gave us an empty iterator"))
+        let (package, term) = package.expect("get_dependencies gave us an empty iterator");
```
mpizenberg: Is it `potential_packages` instead of `get_dependencies`?

src/solver.rs
```diff
@@ -121,10 +121,15 @@ pub fn resolve<P: Package, V: Version>(
             }
             Some(x) => x,
         };
-        debug_assert!(state
+        if state
             .partial_solution
             .term_intersection_for_package(&next)
-            .contains(&v));
+            .contains(&v)
```
mpizenberg, line 133: shouldn't be the reverse? the version is supposed to be contained in the intersection of terms.

mpizenberg (commented):
I personally like the complete control given back to the dependency provider with `make_decision`. The usage of the BTree is also a good idea to remove the `hash` constraint. The trick with `Borrow` to enforce a correct package is also really nice.

There are few oddities though in package docs. Names that do not make a lot of sense anymore like the "reverse" terminology for the dependency provider. And a few other things that I found weird at first read like the error when the version is in the term (should be the opposite?).

Eh2406:
I force pushed to fix the commit names. I also addressed @mpizenberg's comments. I have 2 remaining concerns:

1. The module documentation still refers to `list_available_versions`. I could use some help describing the new method.
2. I added a new trait `Constraints` to hide `Term` is this a good idea? Alternatively we can make `Term` public with only the one method available, or as we are filtering to positive terms we can convert to a `Range` which is already public.

mpizenberg:
> Alternatively we can make Term public with only the one method available, or as we are filtering to positive terms we can convert to a Range which is already public

Good point! Since potential packages are only picked if there is at least one positive derivation, the intersection is necessarily a positive term. Using a `Range` in the public argument type would make more sense from the point of view of the dependency provider!

Eh2406:
switched to `Range`. I (at this time of night) could not get the lifetimes to work for `&Range`  so used `Borrow<Range<V>>`

src/solver.rs
```diff
@@ -93,50 +92,54 @@ pub fn resolve<P: Package, V: Version>(
...
-        let available_versions =
-            dependency_provider
-                .list_available_versions(&p)
-                .map_err(|err| PubGrubError::ErrorRetrievingVersions {
-                    package: p.clone(),
-                    source: err,
-                })?;
+        let iter = state.partial_solution.potential_packages();
+        if iter.is_none() {
+            drop(iter);
```
Eh2406: The borrow checker did not like using a match on iter. This `if ... is_none ... drop` is a workaround. I believe this is a case where [`Polonius`](https://github.com/rust-lang/polonius) could help, when and if it lands in rustc.

src/solver.rs
```diff
@@ -93,50 +92,54 @@ pub fn resolve<P: Package, V: Version>(
...
-        };
-        next = p.clone();
-        let available_versions =
-            dependency_provider
-                .list_available_versions(&p)
-                .map_err(|err| PubGrubError::ErrorRetrievingVersions {
-                    package: p.clone(),
-                    source: err,
-                })?;
+        let iter = state.partial_solution.potential_packages();
```
mpizenberg: What about `potential_packages = state.partial_solution.potential_packages();` for the name instead of `iter`?

mpizenberg (approved):
Looks good to me, with maybe some changes to some doc comments. Waiting on others reviews

src/solver.rs
Eh2406: good point will do.

Eh2406:
> with maybe some changes to some doc comments

What did you have in mind? Also:

> The module documentation still refers to list_available_versions. I could use some help describing the new method.

aleksator:
Hopefully I'll be able to look at PRs tomorrow or the day after.

src/solver.rs
```diff
@@ -166,17 +172,29 @@ pub fn resolve<P: Package, V: Version>(
...
+    /// This method holds a lot of control over the performance and behaviour of the algorithm.
+    /// It is normal to pick the package that has the fewest versions contained in the constraints,
+    /// as they are likely to find problems faster. Also normal is to pick the most recent matching
+    /// version.
+    ///
+    /// Note: the type `T` ensures that this returns an item from the `packages` argument.
+    fn make_decision<T: Borrow<P>, U: Borrow<Range<V>>>(
+        &self,
+        packages: impl Iterator<Item = (T, U)>,
+    ) -> Result<(T, Option<V>), Box<dyn Error>>;
```
aleksator, line 257: I propose to provide a default implementation for this method with the heuristic we were using previously.

This way we still give control to the users who want it, but provide the option to leave the way we pick packages as the library suggests.

src/solver.rs
```diff
@@ -93,50 +92,57 @@ pub fn resolve<P: Package, V: Version>(
...
                 continue;
             }
             Some(x) => x,
         };
+        if !state
+            .partial_solution
+            .term_intersection_for_package(&next)
+            .contains(&v)
+        {
+            return Err(PubGrubError::Failure(
```
aleksator: Should we return a newly introduced error here?

tests/proptest.rs
```diff
@@ -18,19 +18,42 @@ use proptest::string::string_regex;
...
-
-impl<P: Package, V: Version, DP: DependencyProvider<P, V>> DependencyProvider<P, V>
-    for ReverseDependencyProvider<DP>
-{
-    fn list_available_versions(&self, p: &P) -> Result<Vec<V>, Box<dyn Error>> {
-        self.0.list_available_versions(p).map(|mut v| {
-            v.reverse();
-            v
-        })
+struct OldestDependencyProvider<P: Package, V: Version>(OfflineDependencyProvider<P, V>);
```
aleksator: Isn't this `DepedencyProvider` rather new? :smile: 

Maybe `OldestVersionsDependencyProvider` would be better. Alternatively we leave it as `ReverseDependencyProvider` if we go with the suggestion to return newest versions first as the default trait implementation.

src/solver.rs
Eh2406: Do you want to keep the `list_available_versions`? Because without that I don't know how to build a default implementation. It was removed as we otherwise don't need it.

aleksator:
> This does simplify the code, but the performance characteristics are different. In the first case, we only need to browse terms for any positive before computing intersection. In the refactor we always compute the intersection. If benchmarking showcases similar performances, it's worth it. Otherwise let's leave it for another round of performance adjustments.

On the other hand, we don't iterate over not intersected derivations every time.

I run the benchmarks a few times before submitting the change.
Before the change it has completed in 1.65ms on my machine, after the change in 1.67ms. I'm not sure if this is spurious or not, since when I benchmarked something else before I was getting much larger variations without any code changes.

src/solver.rs
Eh2406: or just panic, it is not clear to me how a user of our code could use a `match` in this case. Like the only fix is to fix there implementation.

src/solver.rs
mpizenberg: I don't think we should keep `list_available_versions` in the interface. Maybe we could have a free function (or collection of free functions) that could help. Something like `pick_first_strategy`, `pick_lowest_number_of_versions_strategy` etc. The names are terrible, but that's just for the sake of the example

src/solver.rs
mpizenberg: I'm not sure what panic brings. With an error, the caller can do whatever they want when a `Failure` happens (they could panic). And new error wouldn't bring much either since they have to fix their bug anyway.

Eh2406:
When I ran the benchmarks from #34, criterion was fairly sure there is a small regression comparing ecaeb390311297b478e73945290c304ed086d074 with acc391ec75876e5dfaa19216510612f792a8d881. It is made smaller if we eagerly do the intersections and go ahead and do away with `derivations_not_intersected_yet` entirely. But criterion thinks it is still a 3-5% regression.
It is a code complexity vs perf trade off, I'd be ok either way.

aleksator:
> When I ran the benchmarks from #34, criterion was fairly sure there is a small regression comparing [ecaeb39](https://github.com/pubgrub-rs/pubgrub/commit/ecaeb390311297b478e73945290c304ed086d074) with [acc391e](https://github.com/pubgrub-rs/pubgrub/commit/acc391ec75876e5dfaa19216510612f792a8d881). It is made smaller if we eagerly do the intersections and go ahead and do away with `derivations_not_intersected_yet` entirely. But criterion thinks it is still a 3-5% regression.
> It is a code complexity vs perf trade off, I'd be ok either way.

I've restored "if conditions" and still simplified the code a little with an early return.

src/solver.rs
Eh2406: I was sceptical, but gave it a try. And it worked out really well! Take a look and let me know what you think.

src/lib.rs
```diff
@@ -99,8 +99,11 @@
 //! }
 //! ```
 //! The first method
-//! [list_available_versions](crate::solver::DependencyProvider::list_available_versions)
-//! should return all available versions of a package.
+//! [make_decision](crate::solver::DependencyProvider::make_decision)
```
mpizenberg: I'd write something like this:
```txt
The first method [make_decision](...) chooses a package and available version
compatible with the provided options.
A helper function [make_fewest_...](...) is provided for convenience
in cases when lists of available versions for packages are easily obtained.
The strategy of that helper function consists in choosing the package
with the fewest number of compatible versions to speed up resolution.
But in general you are free to employ whatever strategy suits you best
to pick a package and a version.
```

tests/proptest.rs
```diff
@@ -28,32 +30,10 @@ impl<P: Package, V: Version> DependencyProvider<P, V> for OldestVersionsDependen
...
-        let ver = self
-            .0
-            .versions(package.borrow())
-            .into_iter()
-            .flatten()
-            .filter(|&v| term.borrow().contains(v))
-            .next()
-            .cloned();
-        Ok((package, ver))
+        Ok(make_fewest_versions_decision_helper(
```
mpizenberg: Like the factorization!

mpizenberg (approved):

src/lib.rs
Eh2406: Great wordsmithing! incorporated.

mpizenberg:
The last commit makes the code cleaner by using `min_by` but I think it doubles the amount of calls to `list_available_versions` so don't hesitate to revert that one. It's weird that there is not function like `min_with` that would take a `F: Fn(&Self::Item) -> impl Ord` or similar.

Eh2406:
Dose `min_by_key` work here?

mpizenberg:
Just after that comment I realized the function exists XD and is called `min_by_key`.

mpizenberg (approved):

src/solver.rs
aleksator: AFAIK panicking should be done on internal library bugs, which the user can do nothing about. There are a few we define as Failures currently, which I wanted to change to panics in some future PR.

As for user introduced bugs... I don't know actually what the consensus is. I just changed the error to be more specific in the latest commit for now, but I'm open to the discussion on the topic.

src/solver.rs
aleksator: @Eh2406 
> Do you want to keep the list_available_versions?

@mpizenberg 
> I don't think we should keep list_available_versions in the interface.

Why not? What if our interface is 3 functions:
1) `make_decision()`
2) `list_available_versions()`
3) `get_dependencies()`

The first one has default implementation in terms on `list_available_versions`, no need to change it right away for the algorithm to work but the ability to customize it is there.

The idea is somewhat similar to what iterators to with the `next()` function. You define it and get some others for free. Here we would implement 1) in terms of 2).

The only not so pretty thing about that is if the user comes up with a way to redefine `make_decision` without using `list_available_versions`, but the API still requires it. In that case they could leave it `unimplemented!`. Is it an API smell? Hmmm.

aleksator:
I've reviewed the PR so far, great work introducing the ability to customize the lib's behavior.

The one thing I wanted to discuss about the API change is [a default implementation for `make_decision`](https://github.com/pubgrub-rs/pubgrub/pull/50#discussion_r523457875). We can leave that discussion for a potential separate PR if you'd like.

src/solver.rs
aleksator: I forgot about `should_cancel` but the point stands.

tests/proptest.rs
```diff
@@ -2,35 +2,39 @@
...
-};
-
 use crate::sat_dependency_provider::SatResolve;
 
 mod sat_dependency_provider;
 
-/// The same as DP but takes versions from the opposite end:
-/// if DP returns versions from newest to oldest, this returns them from oldest to newest.
+/// The same as [OfflineDependencyProvider] but takes versions from the opposite end:
+/// if [OfflineDependencyProvider] returns versions from newest to oldest, this returns them from oldest to newest.
```
aleksator, line 25: Is it still true?

```
        Ok(choose_package_with_fewest_versions(
            v
            |p| self.0.versions(p).into_iter().flatten().cloned(),
        })
            potential_packages,
        ))
```
Does not look like we pick "oldest" versions here anymore :thinking:

tests/proptest.rs
```diff
@@ -2,35 +2,39 @@
...
-            v.reverse();
-            v
-        })
+struct OldestVersionsDependencyProvider<P: Package, V: Version>(OfflineDependencyProvider<P, V>);
+
+impl<P: Package, V: Version> DependencyProvider<P, V> for OldestVersionsDependencyProvider<P, V> {
+    fn make_decision<T: std::borrow::Borrow<P>, U: std::borrow::Borrow<Range<V>>>(
+        &self,
+        potential_packages: impl Iterator<Item = (T, U)>,
+    ) -> Result<(T, Option<V>), Box<dyn Error>> {
```
aleksator, line 33: As we note in the docs:
>   /// The strategy employed to choose such package and version
    /// cannot change the existence of a solution or not,

So why should it be possible to return an error here?

What we can do is to return `Option<(T, Option<V>)>` here instead of the result.
Then change `choose_package_with_fewest_versions` to return an `Option` as well and instead of
```
        .expect("potential_packages gave us an empty iterator");
```
we can just return `None`.

Alternatively we get don't introduce the `Option` in this type signature (keeping just `(T, Option<V>)`) and we can leave the above mentioned panic in.

Thoughts?

tests/proptest.rs
```diff
@@ -2,35 +2,39 @@
...
-{
-    fn list_available_versions(&self, p: &P) -> Result<Vec<V>, Box<dyn Error>> {
-        self.0.list_available_versions(p).map(|mut v| {
-            v.reverse();
-            v
-        })
+struct OldestVersionsDependencyProvider<P: Package, V: Version>(OfflineDependencyProvider<P, V>);
+
+impl<P: Package, V: Version> DependencyProvider<P, V> for OldestVersionsDependencyProvider<P, V> {
+    fn make_decision<T: std::borrow::Borrow<P>, U: std::borrow::Borrow<Range<V>>>(
```
aleksator: What about naming this `[pick/choose]_next_package`?

aleksator (changes requested):

tests/proptest.rs
mpizenberg: There are lots of things that can go wrong while giving control back to the implementer of `make_decision` like when reading files on disk. Without a result, we force the implementer to panic inside `make_decision`. With a result, they can handle the error that is propagated to `resolve` however they want.

tests/proptest.rs
mpizenberg: Could have been possible. `make_decision` fits well with PubGrub terminology since this step is called ["decision making"](https://github.com/dart-lang/pub/blob/master/doc/solver.md#decision-making)

tests/proptest.rs
Eh2406: `versions` currently returns them in sorted order, So taking the first will be the smallest/oldest.
`versions` does not document this behavior, but this is only used for testing so I don't mind depending on undocumented behavure.

src/solver.rs
Eh2406: It is a workable solution, if we had a lot of users of the current api that would be a wonderful way to add an opt in for performance without breaking existing users. The boilerplate required to use the helper function, did not end up being so onerous IMO. I think it is a code smell to require our users to provide something that we will not be using.
I don't think "avoid the breaking change" justifies that smell  given our user base, nor do I think "make it easy to get started" justifies it with the helper function. But it can work if that is what we want to do.

Eh2406:
Using the benchmark of solving all versions in [elm-packages.ron](https://github.com/pubgrub-rs/pubgrub/issues/44#issuecomment-727289232), criterion thinks this PR makes a 19% improvement from 2.2sec -> 1.8sec.
Solving for all versions in the checked in large_case.ron criterion thinks this PR makes a 23% improvement.

tests/proptest.rs
aleksator: Makes sense.

src/solver.rs
aleksator: > I don't think "avoid the breaking change" justifies that smell given our user base

My motivation is not avoiding breaking changes, but rather the easiness of API usage.

Before this PR it was crystal clear what the algorithm needed from the user to function (from the user's perspective):
1) A way to get dependencies for a given package version.
2) A way to get all versions for the package.

All makes sense why it is needed and how to implement it.

Now:
1) Unchanged.
2) Hm, actually you need to implement part of our algorithm to use our library.

How would people do that? What if they don't want to understand how exactly we do the decision making?

The only practical solution is to copy paste the code from `OfflineDependencyProvider`.
I'm afraid that requiring every user to do that is too much of an ergonomic hit.

Does anyone else share my concerns?

tests/proptest.rs
aleksator: That's true. The question is who do we optimize the easiness to understand for:
1) library authors
2) users

I think `make_decision` optimizes for us, while naming function based on what it does optimizes for users.

What if we use descriptive function name from the user perspective and put a link to "decision making" in the function docs?

src/solver.rs
mpizenberg: I get your point about ease of use. However, I think people not wanting to use `OfflineDependencyProvider` and willing to define their own dependency provider are willing to read the implementation of an 11-lines function doing nothing more than a function call to a helper function. Also there is an even simpler [6-lines example implementation in the guide](https://pubgrub-rs-guide.netlify.app/pubgrub_crate/custom_dep_provider.html) on the section regarding custom implementation of a dependency provider.

If we get more users, and some of them point `make_decision` as being difficult to implement, I think I'd rather have a two different traits with different ease/control compromises (a bit like the [sandbox/element/document/application variants of an elm app](https://package.elm-lang.org/packages/elm/browser/latest/)) than having a single trait with an ambiguous API.

tests/proptest.rs
mpizenberg: Yeah I don't know, tough naming call. If you prefer explicitness from the user point of view, maybe `choose_package_version_within` is the most descriptive name? I don't mind if you go for it, but can you also update the guide then please (PR #45).

aleksator: https://github.com/dart-lang/pub/blob/master/lib/src/solver/version_solver.dart#L320 :smile:

aleksator: I went with `choose_package_version`; how does it look now?

I haven't yet read the guide as I'm focusing on PR reviews first. Will update once I get there :slightly_smiling_face:

src/solver.rs
aleksator: > If we get more users, and some of them point make_decision as being difficult to implement, I think I'd rather have a two different traits with different ease/control compromises (a bit like the sandbox/element/document/application variants of an elm app) than having a single trait with an ambiguous API.

Actually that's what my girlfriend suggested as well! :laughing: I was pretty impressed because she's sophomore (2nd year in the uni) and is doing that as a way to change her profession, meaning she has very little experience in programming yet.

In any case, doing that is out of scope of this PR and needs a separate discussion; I'm totally okay merging this version!

aleksator:
@Eh2406 That's a lot! Must be usage of iterators instead of allocating new vectors for `list_available_versions` that was done before?

@Eh2406 @mpizenberg 
No blockers from me anymore, let's merge this if you agree with the naming changes in the latest commit discussed in https://github.com/pubgrub-rs/pubgrub/pull/50#discussion_r523755854.
Feel free to resolve and merge.

There is also an interesting discussion on future possible API changes, where @mpizenberg thought of 2 separate traits for customizability/ease of use: https://github.com/pubgrub-rs/pubgrub/pull/50#discussion_r513702296
This is in no way a blocker for this PR, just something we can keep in mind and possibly implement in the future if we desire to do so. I thought it's pretty clever :smile:

src/solver.rs
aleksator: I'm glad we had this discussion as this is an interesting design problem I haven't had to think of before, thank you @Eh2406 and @mpizenberg.

aleksator (approved):

src/solver.rs
```diff
@@ -215,8 +215,11 @@ pub type DependencyConstraints<P, V> = Map<P, Range<V>>;
 /// Trait that allows the algorithm to retrieve available packages and their dependencies.
 /// An implementor needs to be supplied to the [resolve] function.
 pub trait DependencyProvider<P: Package, V: Version> {
-    /// Decision making is the process of choosing the next package
-    /// and version that will be appended to the partial solution.
+    /// This method chooses the next package and its version that will be tested whether
```
mpizenberg: A bit awkward said like that. We've already determined that any existing package version within the constraint is valid so "that will be tested whether ..." is more confusing in my opinion. We should just stick to say that we choose any existing package version within the provided constraints.

mpizenberg (changes requested):

Eh2406:
> Must be usage of iterators instead of allocating new vectors for list_available_versions that was done before?

I think so. Allocating a BTreeSet (in versions) and Vec (in list_available_versions) just to get a count for all available packages for each decision adds up.

@mpizenberg is better at naming than I, so I will leave final approval to him.

src/solver.rs
mpizenberg: you have a wise girlfriend @aleksator  ahah, keep her! XD

src/solver.rs
aleksator: I've restored the wording that was there before the change, just added a link to Pubgrub docs regarding the process.

mpizenberg (approved):
Alright congrats all on finalizing that PR!

mpizenberg:
@Eh2406 There is a conflict and I'm in the middle of some uncommitted things from the old `dev` branch for my analysis of elm packages. Could you do the merge? Don't hesitate to squash everything before rebase since that means only 1 conflictual commit.

Eh2406:
Did a squash and a rebase. I think nothing was lost in that, but just in case I will leave it for one of you to merge.

aleksator:
There was a conflict because of my merged PR; fixed.

PR #53, opened 2020-10-25
ci: add clippy

aleksator:
CI for this PR should pass once #46 and #48 are merged.


mpizenberg (approved):
Awesome that we can check style with clippy now in addition to the other safety nets!

PR #55, opened 2020-10-26
fix: don't panic on Range::none() dependencies

### LINKED ISSUE #36: don't panic: on bad dependencies
Minimizing from the panic found in https://github.com/mpizenberg/pubgrub-rs/pull/34#issuecomment-708764474. I get the small test case of:

```rust
#[test]
fn should_always_find_a_satisfier() {
    let mut dependency_provider = OfflineDependencyProvider::<_, NumberVersion>::new();
    dependency_provider.add_dependencies("a", 0, vec![("b", Range::none())]);

    // Run the algorithm.
    let _ = resolve(&dependency_provider, "a", 0);
}
```
While this is a silly input, we should not panic on it.

Eh2406:
This fixes #36. By filtering out the meaningless `Term` `Negative(Range::none)` in the specific case when it is generated from a dependency on `Range::none()`.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -43,6 +43,21 @@ pub enum PubGrubError<P: Package, V: Version> {
         source: Box<dyn std::error::Error>,
     },
 
+    /// Error arising when the implementer of
+    /// [DependencyProvider](crate::solver::DependencyProvider)
+    /// returned a dependency on an empty range.
+    /// This technically means that the package can not be selected,
+    /// but is clearly some kind of mistake.
+    #[error("Retrieving dependency {dependent} of {package} {version} is the empty set")]
+    ForbiddenEmptyDependency {
+        /// Package whose dependencies we want.
+        package: P,
+        /// Version of the package for which we want the dependencies.
+        version: V,
+        /// The dependent package that requires us to pick from the empty set.
+        dependent: P,
+    },
+
     /// Error arising when the implementer of [DependencyProvider](crate::solver::DependencyProvider)
     /// returned an error in the method [should_cancel](crate::solver::DependencyProvider::should_cancel).
     #[error("We should cancel")]
--- a/src/error.rs
+++ b/src/error.rs
@@ -48,7 +48,7 @@ pub enum PubGrubError<P: Package, V: Version> {
     /// returned a dependency on an empty range.
     /// This technically means that the package can not be selected,
     /// but is clearly some kind of mistake.
-    #[error("Retrieving dependency {dependent} of {package} {version} is the empty set")]
+    #[error("Package {dependent} required by {package} {version} depends on the empty set")]
     ForbiddenEmptyDependency {
         /// Package whose dependencies we want.
         package: P,
--- a/src/error.rs
+++ b/src/error.rs
@@ -49,7 +49,7 @@ pub enum PubGrubError<P: Package, V: Version> {
     /// This technically means that the package can not be selected,
     /// but is clearly some kind of mistake.
     #[error("Package {dependent} required by {package} {version} depends on the empty set")]
-    ForbiddenEmptyDependency {
+    DependencyOnTheEmptySet {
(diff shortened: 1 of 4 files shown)


mpizenberg (commented):
I don't think we should let users put `Range::none()` in dependencies. Saying "package a depends on package b at no version" is in fact saying "package a is forbidden" with how incompatibilities are built. I don't think that's ever going to be what a user may want.

In addition, there is an issue with reporting, due to not having a single source of truth for the incompatibility. The report is built using the `kind` field and thus saying "a depends on b ∅" while in fact the solver is using the `package_terms` field and thus understanding "a is forbidden".

I think we should add a `PubGrubError::ForbiddenEmptyDependency` error or something similar, and return that when users make this mistake.

Alternatively, we could use another type than `Range`, one that cannot be empty. That would mean a type where there is no `none()/empty()` constructor, and no `intersection()` method. Then we could convert that into a range internally but with the type system preventing it from being empty.

Eh2406:
That is fair, I do not think our users will do this intentionally. So we should not "just make it work" but should raise an error.

aleksator:
> Alternatively, we could use another type than Range, one that cannot be empty. That would mean a type where there is no none()/empty() constructor, and no intersection() method. Then we could convert that into a range internally but with the type system preventing it from being empty.

This goes well with "make invalid states unrepresentable" philosophy. I wonder how much it complicates the code.

mpizenberg:
@Eh2406 would you mind also preventing use of the same package like `add_dependency( "p", 0, vec![ ("p", whatever) ])`?

Eh2406:
Happy to, what error type should that return?

aleksator:
Maybe a new `CircularDependencies` error?

This made me thinking about how we handle circular dependencies in general. I've created one in `no_conflict` test. Looks like we don't.
```
--- tests/examples.rs	(revision cbc6d4d4d1ecf8f3200569282b6f4c81135a4db2)
+++ tests/examples.rs	(date 1603922999786)
@@ -19,6 +19,11 @@
         "foo", (1, 0, 0),
         vec![("bar", Range::between((1, 0, 0), (2, 0, 0)))],
     );
+    dependency_provider.add_dependencies(
+        "bar",
+        (1, 0, 0),
+        vec![("foo", Range::between((1, 0, 0), (2, 0, 0)))],
+    );
     dependency_provider.add_dependencies("bar", (1, 0, 0), vec![]);
     dependency_provider.add_dependencies("bar", (2, 0, 0), vec![]);
 ```
The test still passes.

mpizenberg:
> This made me thinking about how we handle circular dependencies in general

Indeed, we don't handle circular dependencies, they are authorized. In practice for dependency managers they should not exist since a packages depended upon should exist before the one depending on it.

> Happy to, what error type should that return?

It isn't really a cycle because a user could say `a: 1` depends on `a: 2`. Maybe simply `InvalidDependencies(String)`

Eh2406:
> Maybe simply `InvalidDependencies(String)`

would that subsume `ForbiddenEmptyDependency` or would that be in addition to it?

mpizenberg:
> would that subsume ForbiddenEmptyDependency or would that be in addition to it?

Maybe `ForbiddenSelfDependency` instead of `InvalidDependency` then?

Eh2406:
How does that look?

mpizenberg (approved):
Very much like it!

Eh2406:
The requested revue was not to pressure you, but was to keep track that you can merge if you like this.

aleksator:
> The requested revue was not to pressure you, but was to keep track that you can merge if you like this.

Don't worry, it's hard to pressure someone who wasn't even opening Github :laughing:
Now I can have almost 2 productive weeks before 24/7 work duty strikes again!

On topic: I did some naming and description changes, if you like them you can press that green button :slightly_smiling_face:

aleksator (approved):

Issue #56, opened 2020-10-30
In memory, no need to keep derivations when there is a decision


mpizenberg:
I realized that while writing the guide. Every usage of memory we have will either do something if there is a decision or something else if there isn't one yet. The only method returning a mix of decision and derivation is `assignment_intersection` but in fact, if there is a decision, we don't need derivations term, since the intersection will directly be the term of the decision.

So in fact we should have `PackageAssignment` be an enum instead of a struct. That would imply some changes in the API, and especially some algorithm changes regarding `remove_decision` that should not be used anymore, but I believe this is doable.

Eh2406:
That is a grate point. I think it will clean things up a lot. If it does then I'd be for it even if it is a small performance regression.

PR #58, opened 2020-11-02
Use an enum in memory

Eh2406:
This is an impl of #56.

The main questionable decision is: `remove_last_decision` was handled by removing `add_version` and using `add_decision` directly.  I think the next call to `unit_propagation` will clean up if one of the dependency is satisfied. But I don't know how important that fast path was.

### DIFF
--- a/src/internal/incompatibility.rs
+++ b/src/internal/incompatibility.rs
@@ -238,11 +238,11 @@ impl<P: Package, V: Version> Incompatibility<P, V> {
                     return Relation::Contradicted((package.clone(), incompat_term.clone()));
                 }
                 None | Some(term::Relation::Inconclusive) => {
-                    // If a package is not present, the intersection is the same as any.
-                    // And according to the rules of satisfactions, the relation would be inconclusive.
-                    // It could also be satisfied if the incompatibility term was also any.
-                    // But we systematically remove those from incompatibilities.
-                    // So we're safe on that front.
+                    // If a package is not present, the intersection is the same as [Term::any].
+                    // According to the rules of satisfactions, the relation would be inconclusive.
+                    // It could also be satisfied if the incompatibility term was also [Term::any],
+                    // but we systematically remove those from incompatibilities
+                    // so we're safe on that front.
                     if relation == Relation::Satisfied {
                         relation =
                             Relation::AlmostSatisfied((package.clone(), incompat_term.clone()));
--- a/src/internal/incompatibility.rs
+++ b/src/internal/incompatibility.rs
@@ -229,16 +229,15 @@ impl<P: Package, V: Version> Incompatibility<P, V> {
     }
 
     /// CF definition of Relation enum.
-    pub fn relation(&self, mut terms: impl FnMut(&P) -> Term<V>) -> Relation<P, V> {
+    pub fn relation(&self, mut terms: impl FnMut(&P) -> Option<Term<V>>) -> Relation<P, V> {
         let mut relation = Relation::Satisfied;
         for (package, incompat_term) in self.package_terms.iter() {
-            match incompat_term.relation_with(&terms(package)) {
-                term::Relation::Satisfied => {}
-                term::Relation::Contradicted => {
-                    relation = Relation::Contradicted((package.clone(), incompat_term.clone()));
-                    break;
+            match terms(package).map(|term| incompat_term.relation_with(&term)) {
+                Some(term::Relation::Satisfied) => {}
+                Some(term::Relation::Contradicted) => {
+                    return Relation::Contradicted((package.clone(), incompat_term.clone()));
                 }
-                term::Relation::Inconclusive => {
+                _ => {
                     if relation == Relation::Satisfied {
                         relation =
                             Relation::AlmostSatisfied((package.clone(), incompat_term.clone()));
--- a/src/internal/incompatibility.rs
+++ b/src/internal/incompatibility.rs
@@ -237,7 +237,12 @@ impl<P: Package, V: Version> Incompatibility<P, V> {
                 Some(term::Relation::Contradicted) => {
                     return Relation::Contradicted((package.clone(), incompat_term.clone()));
                 }
-                _ => {
+                None | Some(term::Relation::Inconclusive) => {
+                    // If a package is not present, the intersection is the same as any.
+                    // And according to the rules of satisfactions, the relation would be inconclusive.
+                    // It could also be satisfied if the incompatibility term was also any.
+                    // But we systematically remove those from incompatibilities.
+                    // So we're safe on that front.
                     if relation == Relation::Satisfied {
                         relation =
                             Relation::AlmostSatisfied((package.clone(), incompat_term.clone()));
--- a/src/internal/memory.rs
+++ b/src/internal/memory.rs
@@ -137,7 +137,7 @@ impl<V: Version> PackageAssignments<V> {
     /// Mutates itself to store the intersection result.
     fn assignment_intersection(&mut self) -> &Term<V> {
         match self {
-            PackageAssignments::Decision((_, term)) => &*term,
+            PackageAssignments::Decision((_, term)) => term,
             PackageAssignments::Derivations {
                 intersected,
                 not_intersected_yet,
@@ -146,7 +146,7 @@ impl<V: Version> PackageAssignments<V> {
                     *intersected = intersected.intersection(&derivation);
                 }
                 not_intersected_yet.clear();
-                &*intersected
+                intersected
             }
         }
     }
--- a/src/internal/memory.rs
+++ b/src/internal/memory.rs
@@ -76,6 +76,7 @@ impl<P: Package, V: Version> Memory<P, V> {
         use std::collections::hash_map::Entry;
         match self.assignments.entry(package) {
             Entry::Occupied(mut o) => match o.get_mut() {
+                // Check that add_derivation is never called in the wrong context.
                 PackageAssignments::Decision(_) => debug_assert!(false),
                 PackageAssignments::Derivations {
                     intersected: _,
--- a/src/internal/memory.rs
+++ b/src/internal/memory.rs
@@ -38,11 +38,10 @@ impl<P: Package, V: Version> Memory<P, V> {
     }
 
     /// Retrieve intersection of terms in memory related to package.
-    pub fn term_intersection_for_package(&mut self, package: &P) -> Term<V> {
-        match self.assignments.get_mut(package) {
-            None => Term::any(),
-            Some(pa) => pa.assignment_intersection().clone(),
-        }
+    pub fn term_intersection_for_package(&mut self, package: &P) -> Option<&Term<V>> {
+        self.assignments
+            .get_mut(package)
+            .map(|pa| pa.assignment_intersection())
     }
 
     /// Building step of a Memory from a given assignment.
@@ -161,10 +160,10 @@ impl<V: Version> PackageAssignments<V> {
     /// selected version (no "decision")
     /// and if it contains at least one positive derivation term
     /// in the partial solution.
-    fn potential_package_filter<'a, 'b, P: Package>(
+    fn potential_package_filter<'a, P: Package>(
         &'a mut self,
-        package: &'b P,
-    ) -> Option<(&'b P, &'a Term<V>)> {
+        package: &'a P,
+    ) -> Option<(&'a P, &'a Term<V>)> {
         match self {
             PackageAssignments::Decision(_) => None,
             PackageAssignments::Derivations {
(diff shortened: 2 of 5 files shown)


src/internal/memory.rs
```diff
@@ -92,11 +116,20 @@ impl<P: Package, V: Version> Memory<P, V> {
     /// a corresponding decision that satisfies that assignment,
     /// it's a total solution and version solving has succeeded.
     pub fn extract_solution(&self) -> Option<Map<P, V>> {
-        if self.assignments.values().all(|pa| pa.is_valid()) {
+        if self.assignments.iter().all(|(_, pa)| match pa {
```
mpizenberg: If we embed the validation directly in `extract_solution` we might as well do a single pass on assignments. We can initialize an empty `decisions` hashmap, iterate in a for loop on assignments, grow the decisions until we reach the end (and return `Some(decisions)`) or a positive derivation (and return `None`).

src/internal/incompatibility.rs
```diff
@@ -225,16 +225,15 @@ impl<P: Package, V: Version> Incompatibility<P, V> {
...
-                term::Relation::Contradicted => {
-                    relation = Relation::Contradicted(package.clone(), incompat_term.clone());
-                    break;
+            match terms(package).map(|term| incompat_term.relation_with(&term)) {
+                Some(term::Relation::Satisfied) => {}
+                Some(term::Relation::Contradicted) => {
+                    return Relation::Contradicted(package.clone(), incompat_term.clone());
                 }
-                term::Relation::Inconclusive => {
+                _ => {
```
mpizenberg: Ok, if a package is not present, the intersection is the same as `any` and according to the rules of satisfactions, the relation would be inconclusive. It could also be satisfied if the incompatibility term was also `any` but we systematically remove those from incompatibilities so we're safe on that front. We might add a comment saying just that.

src/internal/memory.rs
```diff
@@ -36,11 +38,10 @@ impl<P: Package, V: Version> Memory<P, V> {
     }
 
     /// Retrieve intersection of terms in memory related to package.
-    pub fn term_intersection_for_package(&mut self, package: &P) -> Term<V> {
-        match self.assignments.get_mut(package) {
-            None => Term::any(),
-            Some(pa) => pa.assignment_intersection(),
-        }
+    pub fn term_intersection_for_package(&mut self, package: &P) -> Option<&Term<V>> {
```
mpizenberg, line 41: Now returning a reference (in addition to option), ok that's better.

src/internal/memory.rs
```diff
@@ -106,69 +139,44 @@ impl<P: Package, V: Version> Memory<P, V> {
...
-        for derivation in self.derivations_not_intersected_yet.iter() {
-            self.derivations_intersected = self.derivations_intersected.intersection(derivation);
+    fn assignment_intersection(&mut self) -> &Term<V> {
+        match self {
+            PackageAssignments::Decision((_, term)) => &*term,
+            PackageAssignments::Derivations {
+                intersected,
+                not_intersected_yet,
+            } => {
+                for derivation in not_intersected_yet.drain(..) {
```
mpizenberg: Dumb performance question. Is `drain` optimized to be as fast as a `clear` when all values are extracted? I've not tested it but went the other way the first time just from intuition it may not.

src/solver.rs
```diff
@@ -126,47 +126,49 @@ pub fn resolve<P: Package, V: Version>(
...
-        // Add that package and version if the dependencies are not problematic.
-        let start_id = state.incompatibility_store.len();
-        let dep_incompats =
-            Incompatibility::from_dependencies(start_id, p.clone(), v.clone(), &dependencies);
         if added_dependencies
             .entry(p.clone())
             .or_default()
             .insert(v.clone())
         {
+            // Retrieve that package dependencies.
```
mpizenberg, line 134: that diff block is actually just moved code ok

src/solver.rs
```diff
@@ -126,47 +126,49 @@ pub fn resolve<P: Package, V: Version>(
...
-            .any(|incompat| state.is_terminal(incompat))
-        {
-            // For a dependency incompatibility to be terminal,
-            // it can only mean that root depend on not root?
-            return Err(PubGrubError::Failure(
-                "Root package depends on itself at a different version?".into(),
-            ));
-        }
-        state.partial_solution.add_version(p, v, &dep_incompats);
+        state.partial_solution.add_decision(p, v);
```
mpizenberg: So `add_version` is replaced by `add_decision`.

src/internal/memory.rs
```diff
@@ -53,27 +54,50 @@ impl<P: Package, V: Version> Memory<P, V> {
...
     }
 
     /// Add a derivation to a Memory.
     fn add_derivation(&mut self, package: P, term: Term<V>) {
-        let pa = self
-            .assignments
-            .entry(package)
-            .or_insert_with(PackageAssignments::new);
-        pa.derivations_not_intersected_yet.push(term);
+        let pa =
```
mpizenberg: We might have a `PackageAssignments::init_with_derivation(term)` constructor

src/internal/memory.rs
```diff
@@ -53,27 +54,50 @@ impl<P: Package, V: Version> Memory<P, V> {
...
+            self.assignments
+                .entry(package)
+                .or_insert_with(|| PackageAssignments::Derivations {
+                    intersected: Term::any(),
```
mpizenberg: If there is only one term, no need to put it in the `not_intersected_yet` vec, we can put it directly in `intersected`

src/internal/memory.rs
```diff
@@ -53,27 +54,50 @@ impl<P: Package, V: Version> Memory<P, V> {
...
+                    not_intersected_yet: Vec::with_capacity(1),
+                });
+        match pa {
+            PackageAssignments::Decision((version, _)) => {
+                if cfg!(debug_assertions) {
+                    debug_assert!(term.contains(&version))
```
mpizenberg: I wonder if you could even put `debug_assert!(false)`. I don't think we ever add a derivation for a package where a decision was already taken since we only add derivations for almost satisfied packages on incompatibilities (in unit propagation) and a package with a decision already cannot return `Inconclusive` I think

src/internal/memory.rs
```diff
@@ -53,27 +54,50 @@ impl<P: Package, V: Version> Memory<P, V> {
...
+                }
+            }
+            PackageAssignments::Derivations {
+                intersected: _,
+                not_intersected_yet,
+            } => {
+                not_intersected_yet.push(term);
+            }
+        }
```
mpizenberg, line 94: In sum, I'd do something like this. Except there is a clone there because of moved value that we should get rid of. I'm not sure exactly how.
```rust
    /// Add a derivation to a Memory.
    fn add_derivation(&mut self, package: P, term: Term<V>) {
        self.assignments
            .entry(package)
            .and_modify(|pa| match pa {
                PackageAssignments::Decision(_) => {
                    if cfg!(debug_assertions) {
                        debug_assert!(false)
                    }
                }
                PackageAssignments::Derivations {
                    intersected: _,
                    not_intersected_yet,
                } => {
                    not_intersected_yet.push(term.clone());
                }
            })
            .or_insert_with(|| PackageAssignments::Derivations {
                intersected: term,
                not_intersected_yet: vec![],
            });
    }
```

src/internal/memory.rs
```diff
@@ -53,27 +54,50 @@ impl<P: Package, V: Version> Memory<P, V> {
 
     /// Add a decision to a Memory.
```
mpizenberg, line 55:
```rust
    /// We can add the version to the partial solution as a decision
    /// if it doesn't produce any conflict with the new incompatibilities.
    /// In practice I think it can only produce a conflict if one of the dependencies
    /// (which are used to make the new incompatibilities)
    /// is already in the partial solution with an incompatible version.
```

mpizenberg: That check that was done in the removed `add_version` was important. It prevents generating a future conflict and thus avoid a full pass of conflict resolution. That's the heart of what I meant when I said that removing the `remove_decision` function would mean a few changes in the algorithm. We have to have a way to prevent that decision been added if one of it's dependencies incompatibilities would become satisfied.

One option I can see is checking if the dependencies are almost satisfied before adding the decision. If that's the case, it is necessarily because the unsatisfied term is the decision we want to add. And in that case we shouldn't add it since it would mean that we will satisfy one of those incompatibilities.

mpizenberg: > One option I can see is checking if the dependencies are almost satisfied before adding the decision

Actually this is wrong since the derivations intersection for the package we are adding could already match the decision term, and in such case the incompatibility could already be fully satisfied, not just almost satisfied. But since we've already validated the dependencies at that point, the correct (and most efficient) way is I think to use the dependency ranges directly. We could check that none of the `not(range)` terms for dependency ranges are satisfied by the partial solution before adding the decision. Because if any of them are satisfied, we know for sure that the incompatibility `{p: v, dep: not(range)}` will be satisfied once we add the decision for version `v`.

Eh2406: > avoid a full pass of conflict resolution.

The full pass will end up in the same place, so that is what I had in mind by calling this a "fast path". But if it is important there are a number of "simple" ways to add it back in. One is to redefine `remove_last_decision` as `bactrack(self.decision -1)`. Another is to "ask permission not forgiveness" in this case, and... it is hard to say in words but it worked so take a look.

Also now I think we only need to do this if the dependencies are new, as otherwise the incompatibilities are already accounted for by the previous call to `unit_propagation`.

src/internal/memory.rs
Eh2406: I considered that. It felt like a close call between "one pass over assignments" vs "only allocates if needed". Now I see that it is called only when we expect it to work, so "one pass" is better, will fix.

src/internal/incompatibility.rs
Eh2406: Will add

src/internal/memory.rs
Eh2406: Now that you say it, `drain(..)` probably has to do some bookkeeping for panic safety that iter and clear do not, I will switch it back.

src/internal/memory.rs
```diff
@@ -53,27 +54,48 @@ impl<P: Package, V: Version> Memory<P, V> {
...
+        match self.assignments.entry(package) {
+            Entry::Occupied(mut o) => match o.get_mut() {
+                PackageAssignments::Decision(_) => debug_assert!(false),
+                PackageAssignments::Derivations {
+                    intersected: _,
+                    not_intersected_yet,
+                } => {
+                    not_intersected_yet.push(term);
+                }
+            },
```
mpizenberg, line 87: Very elegant!

mpizenberg (approved):
Regarding my commit removing `&*` was there a reason I'm not aware of for having those?
Last commit is a matter of taste I believe. I tend to avoid for loops if possible. Feel free to change back if you don't like the change.

aleksator (approved):

Eh2406:
Shit. The commit created by the squash does not follow the style. I can't force push to dev. So I don't know how to fix it. Sorry!

aleksator:
Interesting. That's what `bors` would prevent, right?

aleksator:
> Eh2406 merged commit 9f6b511 into dev 1 hour ago

It says 5 checks pass, what's the problem? :thinking:

aleksator:
Ah I see, no conventional commit prefix there. I'll see if I can fix it.

Eh2406:
https://github.com/pubgrub-rs/pubgrub/commit/9f6b511b02f6c090d0e0105954f0f9f122a95056

Has a red x

aleksator:
I temporarily allowed force pushes and changed commit name. :man_shrugging:

Eh2406:
Starting now.

Eh2406:
Oops you did it already.

Eh2406:
I tried to put the protections back, but I don't know if I got it correct. Sorry for messing things up.

aleksator:
> I tried to put the protections back, but I don't know if I got it correct. Sorry for messing things up.

I restored them after pushing, they should have been already fine :slightly_smiling_face: Don't worry! No big deal.

Eh2406:
>  they should have been already fine

before I deleted it. 🤦 
You fixed it. Then I posted. Then I removed the protections. (dumly by delling, instead of unchecking the one that matters.) Then I noticed that you took care of it, and tried to put things back.

aleksator:
It's fine, relax! 😃

I've checked the settings for the `dev` branch, the only difference is we didn't have `Require status checks to pass before merging` selected before. I don't think turning it on even worked since no required checks were selected (there are CI jobs that you mark as required after choosing that option). I've unchecked the whole thing just to make it clearer.

In my understanding, what that option does is it requires the branch to be rebased on dev before you can use `squash merge` to run the branch through CI again. That's something we might want (or not). We can discuss it on Zulip if you feel like.

Issue #27, opened 2020-10-09
Benchmarking tools


mpizenberg:
Performance improvements will need some benchmarking. Things are being setup for that, but for the time being we have one big registry generated for tests, that has been committed in the repository. So I'll use this one as support for the commands listed below.

Usually, performance improvement work goes uses a cycle like this one
1. Measure
2. Profile
3. Optimize
4. back to 1.

To measure timing, we can setup benchmarks. [Criterion](https://github.com/bheisler/criterion.rs) is good for micro-benchmark and supported on stable. The default [cargo-bench](https://doc.rust-lang.org/cargo/commands/cargo-bench.html) is good enough for bigger benchmarks but requires nightly. We'll use the later for our big registry.
Profiling enables a more granular view of which functions take the most time. We usually profile both CPU and memory. Once the culprit have been identified, it is time to do some optimization and back to measure!

## Measure

First we need to add debug symbols to our compilation profiles, so that the binary we generate for benchmarks can be profiled later. So we add the following in `Cargo.toml`.

```toml
# in Cargo.toml
[profile.release]
debug = true

[profile.bench]
debug = true
```

Then, we will be running the `large_case` benchmark available in the repo. It needs serde to load the registry configuration which is stored as data in a `Ron` file.

```sh
cargo +nightly bench large_case --features=serde
```

It will print to stdout the name of the generated executable for the benchmark, something like:

```txt
     Running target/release/deps/large_case-96ac47771292fd35
```

It will also print the timings results, we should note them to see if we can make improvements later.

### Using [flamegraph](https://github.com/flamegraph-rs/flamegraph)

```sh
flamegraph -o flamegraph.svg target/release/deps/large_case-96ac47771292fd35
```

![pubgrub-flamegraph](https://user-images.githubusercontent.com/2905865/95625909-9742f380-0a79-11eb-8ee2-39533e49c85e.png)

### Using [perf](https://perf.wiki.kernel.org/index.php/Main_Page)

```sh
sudo perf record target/release/deps/large_case-96ac47771292fd35
sudo perf report
```

![pubgrub-perf](https://user-images.githubusercontent.com/2905865/95625950-ac1f8700-0a79-11eb-8474-a7c81bd62516.png)

### Using [valgrind + callgrind + KCachegrind](http://www.codeofview.com/fix-rs/2017/01/24/how-to-optimize-rust-programs-on-linux/)

```sh
[...]

Issue #49, opened 2020-10-23
Is the bump method on the Version trait actually necessary?


mpizenberg:
The ability to know the next version, expressed by the `bump(&self) -> Self` method on the `Version` trait does not seem to be vital. In the code, it is only used twice. The first time is for the definition of `Range::exact(Version) -> Range`. The second time is in the implementation of `Display` for an exact range, in order to have more readable reports. There, it is used to check if we have a range like `(v1, Some(v2))` where `v2 == v1.bump()`. In such case it would be displayed `v1` instead of `v1 <= v < v2`.

In addition, requiring a `bump` method seems like something that will prevent usage of things like pre-releases or versions based on dates, floating point numbers or other potentially valid versions otherwise.

If we were to remove the `bump` method, how could be represent ranges containing a single version that would still play nice with set operations like intersection and union?

mpizenberg:
> floating point numbers 

well maybe not that since it's not `Ord`

Eh2406:
Is it worth thinking even more meta, can we take a `Range` as a user provided type? What functionality do we need from a `Range`?

mpizenberg:
> Is it worth thinking even more meta, can we take a Range as a user provided type? What functionality do we need from a Range?

Indeed. I'm not 100% sure, but I believe the `negate`, `intersection` and `contains` methods on a trait would be enough for everything (+ some trait bounds like `Display` and others).

mpizenberg:
Skimming through the code, `Range::none()` would probably also be needed, and should be renamed `Range::empty()` in my opinion. And `Range::exact(v)` would also be needed, and could also be renamed `singleton`.

The `negate()` method should also be renamed `complement()` I think to avoid confusion with term negation, and because `Range` really describes mathematical sets and so set terminology would be appropriate. I remember hesitating between `Set` and `Range` the first time when choosing the name but eventually settling on `Range` because `Set` would be confusing with those defined in std.

Eh2406:
> seems like something that will prevent usage of things like pre-releases or versions based on dates, floating point numbers or other potentially valid versions otherwise.

For example:
a. `Range([1.0.0, Some(1.1.0-beta)])` meaning something like normal versions between `1.0.0` and `1.1.0` and pre-releases of `1.1.0`
b. `Range([1.1.0-alpha, Some(1.2.0)])`meaning something like normal versions between `1.1.0` and `1.2.0` and pre-releases of `1.1.0`

[...]

PR #59, opened 2020-11-03
perf: optimize term::relation_with()

Eh2406:
In #27 it was measured that we spend a lot of our runtime in `intersection`s in `relation_with`. For that use case we do not need the exact list of what versions are in the `intersection` just a trinary state. So this adds a function to `Range` to compute that directly.

Criterion thinks this is a ~20% improvement in runtime of the benchmarks from #34.
This is mostly because of skipping the cache miss and allocation of making a new `Range`. But also because we can short circuit when we see a `Relation::Inconclusive`.

This still needs wordsmithing.

### DIFF
--- a/src/internal/incompatibility.rs
+++ b/src/internal/incompatibility.rs
@@ -9,9 +9,9 @@ use std::fmt;
 use crate::internal::arena::{Arena, Id};
 use crate::internal::small_map::SmallMap;
 use crate::package::Package;
-use crate::range::Range;
+use crate::range::{self, Range};
 use crate::report::{DefaultStringReporter, DerivationTree, Derived, External};
-use crate::term::{self, Term};
+use crate::term::Term;
 use crate::version::Version;
 
 /// An incompatibility is a set of terms for different packages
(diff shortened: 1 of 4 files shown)


mpizenberg:
At a first glance, it seems to be a non-negligible readability/performance tradeoff. Would you mind waiting a bit for those kind of perf improvements? It also would complicate the promising `Range` trait approach I believe

Eh2406:
> It also would complicate the promising `Range` trait approach I believe

Yes (cc #49) it would need to be added to the proposed trate. We could provide a default implementation that uses `intersection` as we do now, and users can override it when performance matters.

> Would you mind waiting a bit for those kind of perf improvements?

Yes, we can definitely weight. Lets get 0.2 out.

Eh2406:
Using the [elm benchmark](https://github.com/pubgrub-rs/pubgrub/issues/44#issuecomment-727347423) this was a smaller improvement then the noize. I.E. Criterion thought this was faster but not statistically significantly so. So this should be put on the back burner for a while. When we find other ways to make the elm benchmark faster, this may pull its weight.

Eh2406:
running on top of #74 using the benchmarks from #75

```
Benchmarking large_cases/elm-packages_str_SemanticVersion.ron: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 20.0s. You may wish to increase target time to 31.2s, or reduce sample count to 60.
                        time:   [310.00 ms 310.27 ms 310.55 ms]
                        change: [-6.6554% -6.5353% -6.4141%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
                        time:   [16.331 ms 16.344 ms 16.357 ms]
                        change: [-9.1269% -9.0116% -8.8980%] (p = 0.00 < 0.05)
                        Performance has improved.
```

So it now looks like an improvement. Can the code be cleaned up to make it worth it? It needs work, but we have something to talk about.

mpizenberg:
I'm personally still not very attracted by this change. Thanks to all other changes, the perf improvement here is down to 9% where it was roughly 20% at the beginning. And this makes the code more complex to read and to explain, it also doesn't match anymore with the expression in the guide documentation. Let's keep it around and focus on features important for cargo instead. Once we have something working for cargo, we can see if that perf improvement is still worth the change.

Eh2406:
I am closing, thanks to `SmallVec` this is not the best perf/complexity change we can make.

Eh2406:
In https://github.com/pubgrub-rs/pubgrub/pull/157#issuecomment-1833373825 reported large improvements from reducing the calls to `V.clone()`, so at some point this optimization will be helpful to them.

Issue #57, opened 2020-10-31
Conflict resolution improvements


mpizenberg:
Writing the [detailed explanation on conflict resolution](https://github.com/pubgrub-rs/pubgrub/blob/guide/guide/src/internals/conflict_resolution.md) helped me better understand how it works. In particular I think we can improve the conflict resolution code in many ways, including:

1. Avoiding the computation of `self.partial_solution.relation(&root_cause)` after conflict resolution. That also means avoiding one impossible `Failure` case so that's great.
2. Simplify prior cause computation.
3. Doing only one pass instead of two by keeping the decision level of the last assignment that made a term of the incompatibility satisfied.

(1) I now understand that since decision levels of the satisfier and previous satisfier are different, backtracking to the previous decision level implies that only one term will be unsatisfied in the incompatibility, so the incompatibility is guaranteed to be almost satisfied. And we know that the unsatisfied package is the one of the satisfier. So no need to redo that `self.partial_solution.relation(&root_cause)` call after conflict resolution, we can instead return that package term of the incompatibility in the output of conflict resolution.

(2) Currently, when computing prior cause, we are explicitly passing it the term of the derivation. But that term is guaranteed to be the negation of the term corresponding to that package in the `cause` of that derivation. So we can simply remove that from the `prior_cause` method arguments. Then, we can rewrite prior cause as:

```diff
diff --git a/src/internal/incompatibility.rs b/src/internal/incompatibility.rs
index 64c7a20..31a113a 100644
--- a/src/internal/incompatibility.rs
+++ b/src/internal/incompatibility.rs
@@ -207,15 +207,14 @@ impl<P: Package, V: Version> Incompatibility<P, V> {
         satisfier: &Term<V>,
     ) -> Self {
         let kind = Kind::DerivedFrom(incompat.id, satisfier_cause.id);
-        let mut t1 = incompat.package_terms.clone();
-        let mut t2 = satisfier_cause.package_terms.clone();
-        t1.remove(package);
-        t2.remove(package);
-        let mut prior_cause_terms = Self::intersection(&t1, &t2);
-        let term = incompat.package_terms.get(package).unwrap();
-        let p_term = term.union(&satisfier.negate());
-        if p_term != Term::any() {
-            prior_cause_terms.insert(package.clone(), p_term);
+        let mut incompat1 = incompat.package_terms.clone();
+        let mut incompat2 = satisfier_cause.package_terms.clone();
[...]

PR #60, opened 2020-11-03
refactor: don't store term in structs

Eh2406:
This is part 1 and 2 of #57, submitting a PR as we are saving part 3 for after 0.2.

### DIFF
--- a/src/internal/assignment.rs
+++ b/src/internal/assignment.rs
@@ -24,8 +24,6 @@ pub enum Assignment<P: Package, V: Version> {
     Derivation {
         /// The package corresponding to the derivation.
         package: P,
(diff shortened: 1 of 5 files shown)


Eh2406:
To be clear each commit was tested by adding a `assert_eq!(old, new)` and getting test to pass before commit.

mpizenberg (approved):
All good to me, except maybe the removal of the derivation term that may be controversial but justifiable

Eh2406:
Yes, if `as_term` starts to appear in profiles then we should try adding the `term` field back in. For now I think that having less in memory and having fewer meaningless states representable is probably worth it.

aleksator (approved):

PR #62, opened 2020-11-15
refactor!: pluralize NoVersion incompatibilities

aleksator:
Just a small thing that bothered me.

I've also noticed that we mostly duplicate incompatibility::Kind with report::External, so maybe we should see if we could combine them.

### DIFF
--- a/examples/branching_error_reporting.rs
+++ b/examples/branching_error_reporting.rs
@@ -56,7 +56,7 @@ fn main() {
     match resolve(&dependency_provider, "root", (1, 0, 0)) {
         Ok(sol) => println!("{:?}", sol),
         Err(PubGrubError::NoSolution(mut derivation_tree)) => {
-            derivation_tree.collapse_noversion();
+            derivation_tree.collapse_no_versions();
             eprintln!("{}", DefaultStringReporter::report(&derivation_tree));
             std::process::exit(1);
         }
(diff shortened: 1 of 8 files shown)


mpizenberg:
Thanks for the doc comments on `Kind`! As for the plural form, I thought we should not say "there are no versions" but "there is no version", in the sense that when something is not present we should not use the plural form. But I may be wrong about that rule and mixing with another language.

aleksator:
A couple of links I looked at when trying to find a rule show examples of "no" + plural:
https://www.englishclub.com/grammar/determiners-quantifiers-some-any-no.htm
https://www.tolearnenglish.com/exercises/exercise-english-2/exercise-english-117016.php

Pub uses plural: https://github.com/dart-lang/pub/blob/23a59a380cdc88aca6a50f9b0147ecdd3b948919/lib/src/solver/incompatibility_cause.dart#L26

Maybe both are correct. This way sounded more natural to me. :shrug:

mpizenberg (approved):

mpizenberg:
> I've also noticed that we mostly duplicate incompatibility::Kind with report::External, so maybe we should see if we could combine them.

They aren't exactly the same, and I agree reporting will need some work. I suggest reading the guide first to have more insight on why those enums are the way they are.

> Maybe both are correct. This way sounded more natural to me. shrug

alright, looks good to me then

PR #63, opened 2020-11-16
Version 0.2 of pubgrub

mpizenberg:
This is for publication of version 0.2 of pubgrub. Once this is merged in dev, I'll just create a merge commit in master.

PR #65, opened 2020-11-17
docs: change doc comment of NoVersions

mpizenberg:


Eh2406 (approved):

aleksator:
Comment from https://github.com/pubgrub-rs/pubgrub/pull/63#issuecomment-729079888:
> Ok for the plural form but the comment is not correct. This really means that there exist no version in the given range. So for example if that range is 2.1.0 <= v it means there doesn't exist any version higher than 2.1.0 (included) for this package. I'll change it back, but was there something in particular that made this confusing for you?

We are talking about the same thing.

Here's the doc comment from pub ([link](https://github.com/dart-lang/pub/blob/23a59a380cdc88aca6a50f9b0147ecdd3b948919/lib/src/solver/incompatibility_cause.dart#L24-L26)):
```
  /// The incompatibility indicates that the package has no versions that match
  /// the given constraint.
  static const IncompatibilityCause noVersions = _Cause('no versions');
```
They might have gone with that wording since they don't store what those "constraints" are, but we do and that is `Range`. I agree that since we do store them, we can just refer to that `Range` and your proposed change is clearer.

PR #70, opened 2020-11-20
docs(changelog): fix the link to v0.1.0..v0.2.0 diff

aleksator:


CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased [(diff)][diff-unreleased]
 
-## [0.2.0] - 2020-11-19 - [(diff with 0.1.0)][diff-0.2.0]
+## [0.2.0] - 2020-11-19 - [(diff with 0.1.0)](https://github.com/pubgrub-rs/pubgrub-rs/compare/v0.1.0...v0.2.0)
```
mpizenberg:
```suggestion
## [0.2.0] - 2020-11-19 - [(diff with 0.1.0)](https://github.com/pubgrub-rs/pubgrub/compare/v0.1.0...v0.2.0)
```

Weird that github manages to resolve the link even when it is wrong.
Is there a particular reason you prefer those been inlined? (I'm fine with it)

mpizenberg (changes requested):

CHANGELOG.md
aleksator: > Weird that github manages to resolve the link even when it is wrong.

Well, this repo was named `pubgrub-rs` and we renamed it after transferring to the organization, so there is a redirect established.
It was not my intention to use it though, thanks!

As for inlining... I didn't think that having a reusable link to the diff serves any purpose as it's used only in one place. Also when the amount of diffs grows with new releases, it would grow the bottom of the file.

If you think it makes it easier to add new links in the future, I could revert the inlining.

mpizenberg (approved):
I just thought it was cleaner to have all versions and diffs links in one place but it's fine like that too

aleksator:
> I just thought it was cleaner to have all versions and diffs links in one place but it's fine like that too

You might be right. I shifted it to the bottom again, but put it next to the version we are making a diff with.
This way after 10 released we wouldn't have to add [0.10.0], scroll down to all releases and add [diff-0.10.0], but we can just add them next to each other.

After doing that, I thought that it looks better if it's put between the versions and the tag matches the text, so I've added 2 more commits to do that :)

mpizenberg (approved):
Slightly better indeed

PR #72, opened 2020-11-21
perf: make our own smallvec with no unsafe

Eh2406:
As discussed in https://rust-lang.zulipchat.com/#narrow/stream/260232-t-cargo.2FPubGrub/topic/post.200.2E2.20perf.20work.3A.20small_vec this makes an unreasonably large difference to the runtime on my laptop. Almost all `Range`s and `Incompatibility`s are len < 2. This PR means that in that happy path we do not need to do an allocation nor do we have a cache miss to read them.

I have added a change that removes the fallback to `Map` for `prior_cause`.

### DIFF
--- a/src/internal/small_vec.rs
+++ b/src/internal/small_vec.rs
@@ -1,0 +1,84 @@
+use std::fmt;
+
+#[derive(Clone)]
+pub enum SmallVec<T> {
+    Empty,
+    One([T; 1]),
+    Two([T; 2]),
+    Flexible(Vec<T>),
+}
+
+impl<T> SmallVec<T> {
+    pub fn empty() -> Self {
+        Self::Empty
+    }
+
+    pub fn one(t: T) -> Self {
+        Self::One([t])
+    }
+
+    pub fn as_slice(&self) -> &[T] {
+        match self {
+            Self::Empty => &[],
+            Self::One(v) => v,
+            Self::Two(v) => v,
+            Self::Flexible(v) => v,
+        }
+    }
+
+    pub fn push(&mut self, new: T) {
+        *self = match std::mem::take(self) {
+            Self::Empty => Self::One([new]),
+            Self::One([v1]) => Self::Two([v1, new]),
+            Self::Two([v1, v2]) => Self::Flexible(vec![v1, v2, new]),
+            Self::Flexible(mut v) => {
+                v.push(new);
+                Self::Flexible(v)
+            }
+        }
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = &T> {
+        self.as_slice().iter()
+    }
+}
+
+impl<T> Default for SmallVec<T> {
+    fn default() -> Self {
+        Self::Empty
+    }
+}
+
+impl<T: PartialEq> Eq for SmallVec<T> {}
+
+impl<T: PartialEq> PartialEq<SmallVec<T>> for SmallVec<T> {
+    fn eq(&self, other: &SmallVec<T>) -> bool {
+        self.as_slice() == other.as_slice()
+    }
+}
+
+impl<T: fmt::Debug> fmt::Debug for SmallVec<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.as_slice().fmt(f)
+    }
+}
+
+#[cfg(feature = "serde")]
+impl<T: serde::Serialize> serde::Serialize for SmallVec<T> {
+    fn serialize<S: serde::Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
+        serde::Serialize::serialize(self.as_slice(), s)
+    }
+}
+
+#[cfg(feature = "serde")]
+impl<'de, T: serde::Deserialize<'de>> serde::Deserialize<'de> for SmallVec<T> {
+    fn deserialize<D: serde::Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
+        let items: Vec<T> = serde::Deserialize::deserialize(d)?;
+
+        let mut v = Self::empty();
+        for item in items {
+            v.push(item);
+        }
+        Ok(v)
+    }
+}
--- a/src/range.rs
+++ b/src/range.rs
@@ -17,6 +17,7 @@
 use std::cmp::Ordering;
 use std::fmt;
 
+use crate::internal::small_vec::SmallVec;
 use crate::version::Version;
 
 /// A Range is a set of versions.
@@ -24,7 +25,7 @@ use crate::version::Version;
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(transparent))]
 pub struct Range<V: Version> {
-    segments: Vec<Interval<V>>,
+    segments: SmallVec<Interval<V>>,
 }
 
 type Interval<V> = (V, Option<V>);
@@ -34,7 +35,7 @@ impl<V: Version> Range<V> {
     /// Empty set of versions.
     pub fn none() -> Self {
         Self {
-            segments: Vec::new(),
+            segments: SmallVec::empty(),
         }
     }
 
@@ -47,14 +48,14 @@ impl<V: Version> Range<V> {
     pub fn exact(v: impl Into<V>) -> Self {
         let v = v.into();
         Self {
-            segments: vec![(v.clone(), Some(v.bump()))],
+            segments: SmallVec::one((v.clone(), Some(v.bump()))),
         }
     }
 
     /// Set of all versions higher or equal to some version.
     pub fn higher_than(v: impl Into<V>) -> Self {
         Self {
-            segments: vec![(v.into(), None)],
+            segments: SmallVec::one((v.into(), None)),
         }
     }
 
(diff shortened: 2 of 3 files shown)


aleksator:
Whoops, wanted to delete the local branch to pull this in for review and instead miss-clicked. Sorry.

aleksator:
Can we split this into 2 PRs, one for small_vec and one for small_map? Alternatively, we can keep it as is and do `rebase merge` with 2 commits in the end.

Cargo.toml
```diff
@@ -26,6 +26,7 @@ ron="0.6"
 varisat="0.2.2"
 criterion = "0.3"
 
+
```
aleksator: On purpose?

src/internal/small_map.rs
```diff
@@ -0,0 +1,203 @@
...
+                }
+            }
+            Self::Flexible(data) => data.get_mut(q),
+        }
+    }
+
+    pub fn remove(&mut self, q: &K) -> Option<V> {
+        let s = std::mem::replace(self, Self::Flexible(Default::default()));
+        let out;
+        *self = match s {
```
aleksator: If this was `mem::replace`'d with `Flexible`, how can it be anything else?

src/internal/small_vec.rs
```diff
@@ -0,0 +1,93 @@
...
+        match self {
+            Self::One(v) => &*v,
+            Self::Two(v) => &*v,
+            Self::Flexible(v) => &*v,
+        }
+    }
+
+    pub fn push(&mut self, new: T) {
+        let s = std::mem::replace(self, Self::zero());
+        *self = match s {
```
aleksator: If this was `mem::replace`'d with `zero()`, how can it be anything else?

aleksator: Replace returns previous value, got it.

src/internal/small_map.rs
```diff
@@ -0,0 +1,193 @@
...
+    /// When a key is common to both,
+    /// apply the provided function to both values.
+    /// If the result is None, remove that key from the merged map,
+    /// otherwise add the content of the Some(_).
+    pub fn merge<'a>(
+        &'a mut self,
+        map_2: impl Iterator<Item = (&'a K, &'a V)>,
+        f: impl Fn(&V, &V) -> Option<V>,
+    ) {
+        for (key, val_2) in map_2 {
```
mpizenberg: Depending on the order of the keys, considering two `SmallMap` of 2 elements, there is a 50% chance that it grows to 3 elements before removing a key back to 2. If we could find a benchmark with high conflict rate, it may be interesting to check if increasing the small map enum to 3 or 4 elements might be worth it.

src/internal/small_vec.rs
```diff
@@ -0,0 +1,87 @@
...
+            Self::Flexible(v) => v,
+        }
+    }
+
+    pub fn push(&mut self, new: T) {
+        let s = std::mem::replace(self, Self::empty());
+        *self = match s {
+            Self::Empty => Self::One([new]),
+            Self::One([v1]) => Self::Two([v1, new]),
+            Self::Two([v1, v2]) => Self::Flexible(vec![v1, v2, new]),
```
mpizenberg, line 33: Since I believe `Vec` first breaking point in memory allocation is 4 elements, it might be worth checking if adding enum variants for 3 and 4 elements could improve (or not) performances.

src/range.rs
```diff
@@ -125,8 +122,8 @@ impl<V: Version> Range<V> {
     /// For example:
     ///    [ (v1, None) ] => [ (start, Some(v1)) ]
     ///    [ (v1, Some(v2)) ] => [ (start, Some(v1)), (v2, None) ]
-    fn negate_segments(start: V, segments: &[Interval<V>]) -> Vec<Interval<V>> {
-        let mut complement_segments = Vec::with_capacity(1 + segments.len());
+    fn negate_segments(start: V, segments: &[Interval<V>]) -> Range<V> {
+        let mut complement_segments = SmallVec::with_capacity(1 + segments.len());
```
mpizenberg: Negation (complement) of a range result at most into one additional segment. Since this is used a lot, it would be worth trying to extend the smallvec enum to 3 or 4 elements, and initialize this with `SmallVec::empty` instead I think.

src/range.rs
```diff
@@ -149,16 +149,9 @@ impl<V: Version> Range<V> {
...
-            segments: Self::intersection_segments(&self.segments, &other.segments),
-        }
-    }
-
-    /// Helper function performing intersection of two interval segments.
-    fn intersection_segments(s1: &[Interval<V>], s2: &[Interval<V>]) -> Vec<Interval<V>> {
-        let mut segments = Vec::with_capacity(s1.len().min(s2.len()));
-        let mut left_iter = s1.iter();
-        let mut right_iter = s2.iter();
+        let mut segments = SmallVec::with_capacity(self.segments.len().min(other.segments.len()));
```
mpizenberg: Same remark that with `negate`. Though there is a small mistake here. I thought that the number of segments would not go further than the one with the least amount of segments but this is wrong (obvious contradiction with intersection with `Range::any()`). I think the number of segments in the intersection is limited by the sum of the number of segments in each but anyway that is not the most common case.

Important point is that I believe we should initialize with `SmallVec::empty()` instead, and probably extend our smallvec with variants of 3 or 4 elements.

src/internal/small_vec.rs
```diff
@@ -0,0 +1,87 @@
...
+            Self::Flexible(mut v) => {
+                v.push(new);
+                Self::Flexible(v)
+            }
+        }
+    }
+
+    pub fn with_capacity(len: usize) -> Self {
+        Self::Flexible(Vec::with_capacity(len))
```
mpizenberg: Should we match against the size in case `len` is less than the flexible breakpoint?

Eh2406: I thought I did. Looks like @aleksator removed it, why have it allocate if the len is breakpoint?

Eh2406:
> Can we split this into 2 PRs, one for small_vec and one for small_map? Alternatively, we can keep it as is and do rebase merge with 2 commits in the end.

I will rework the history and make this just for the vec part.

src/internal/small_vec.rs
Eh2406: I did some experiments and adding more variants did not help with the elm benchmark. It makes all Ranges 1/n bigger on the stack for ocasinaly not having to allocatie. It is not surprising that there is a tipping point somewhere. Exactly where optimizes runtime will depend on the size of V, and the size of system cache, an other optimizations. So we should continue to experiment.

Eh2406:
I think I addressed the comments. The follow up branch is called `small_map` and will be submitted when we are happy with this one.

mpizenberg (approved):
Great job! Criterion gives me -32% duration (just SmallVec) so it is a great improvement

mpizenberg:
> I did some experiments and adding more variants did not help with the elm benchmark

Same here so I agree stopping at `Two` is the right choice for `SmallVec`.

aleksator:
> Great job! Criterion gives me -32% duration (just SmallVec) so it is a great improvement

I might have introduced some regressions with my edits, needs checking.

I still think we should just pull an actual small_vec though even after spending time on this.

mpizenberg:
> I still think we should just pull an actual small_vec though even after spending time on this.

I tend to I disagree here. Usually I tend to prefer using a dependency for things a bit complicated but this is just "making a simple data structure that fits very well with our needs". One thing to value is also the absence of `unsafe`. But if you find a crate which is also unsafe-free, with the same perf improvements, and not dependency-heavy then yes it might be worth it. What are the reasons that make you prefer using an external crate?

Eh2406:
> I might have introduced some regressions with my edits, needs checking.

I did review your edits carefully for that. I had separate commits to document the places I changed. But then ended up squashing the history in one commit. Sorry.

In general I don't want the perfect to be the enemy of progress. This is faster and has no unsafe, nor changes to the api. If at some point we find that a crate and/or unsafe and/or tweaks are faster or more maintainable I will not object to a new PR that wipes this all away.

The crates I know about are:
- `small_vec` witch has a drop in api with `Vec` witch is a very big plus, but has a history of getting things very wrong with unsafe. The people that (I trust to) know unsafe the best are very wary of using it where there are other options. <100 lines of code feels like a good call to not use it. If it grows to be >300 lines or we start using unsafe for it then I would definitely rethink it. 
- `tinyvec` witch has no unsafe, but is backed by a `[T::defalt(); N]`, so either we add `Default` to the definition of `Version`, witch feels like and unnecessary change to the api. Or we use and `T: Option<Interval<V>>`, in which will lead to more unraps as it has more representable states then we need. Or we `impl Default for Interval<V>` but the only way to gen a `V` is `::lowest()` witch may not be cheap, so why call it more than needed.

Overall I'm happy to consider crates or other implementations, but this feels like the thing to beet. So I'd lean towards having those conversations in follow up PRs.

Eh2406:
I was trying to remember what I was up to on this, I think we want @aleksator feedback before merge.
Also if we add a [Range trait](https://rust-lang.zulipchat.com/#narrow/stream/260232-t-cargo.2FPubGrub/topic/A.20Range.20trate), then if users want a different perf/unsafe/simplicity trade off then thy can provide it. So it is no longer "the range implementation" but "a range implementation we run benchmarks on".

PR #74, opened 2021-01-23
perf: reuse the same buffer in unit_propagation

Eh2406:
Using `dhat-rs` I saw that most of our allocations are short lived allocation inlined into `unit_propagation`. The only allocation in there is the calls to `vec![`, so this saves one vec in State so we can reuse the same buffer.

Running the benchmarks this provides a major speed up when applied on top of #73

```
Warning: Unable to complete 100 samples in 20.0s. You may wish to increase target time to 30.3s, or reduce sample count to 60.
large_cases/elm-packages_str_SemanticVersion.ron                                                                            
                        time:   [301.46 ms 301.66 ms 301.86 ms]
                        change: [-19.093% -19.010% -18.927%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  8 (8.00%) high mild
large_cases/large_case_u16_NumberVersion.ron                                                                             
                        time:   [16.784 ms 16.798 ms 16.813 ms]
                        change: [-21.591% -21.498% -21.406%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  8 (8.00%) high mild
  1 (1.00%) high severe

```

PR #76, opened 2021-01-25
refactor: do less work in backtrack

Eh2406:
Using `dhat-rs` I saw that most of the bytes we allocate over a run are in `PartialSolution::from_assignments`. This cleans up some things so we can reuse the main `Memory.assignments` and `history` allocations. `dhat-rs` is happy, we are allocating less memory. `criterion` is unimpressed, no change to runtime on my computers.

I do have thoughts for other ways to store history/memory that have a better Big-O on backtrack, but this is a PR for a small refactor, not for a big redesign.

### DIFF
--- a/src/internal/memory.rs
+++ b/src/internal/memory.rs
@@ -38,6 +38,11 @@ impl<P: Package, V: Version> Memory<P, V> {
         }
     }
 
+    /// Clears the memory.
+    pub fn clear(&mut self) {
+        self.assignments.clear();
+    }
+
     /// Retrieve intersection of terms in memory related to package.
     pub fn term_intersection_for_package(&mut self, package: &P) -> Option<&Term<V>> {
         self.assignments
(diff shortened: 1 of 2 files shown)


Eh2406:
For reference. Using dhat-rs on this with #74 on `large_case_u16_NumberVersion`
befor: `dhat: Total:     25,909,948 bytes in 36,567 blocks`
after: `dhat: Total:     20,513,896 bytes in 35,475 blocks`

mpizenberg (approved):
Same as you, I don't see any time improvement on benchmark. But less allocation is better, especially in environment where allocations are more expensive like wasm for example. I also had ideas for different ways to store the partial solution at some point. I don't remember very well, but I believe I wanted to find a way to have a single representation instead of a dual memory/history. Anyway, I like this change! thanks!

Eh2406:
Retriggering CI by closing and reopening

Eh2406:
All concerns have been addressed and there have been 10 days without new concerns. So this can be merged if you still like it.

Eh2406:
rebased

PR #77, opened 2021-02-08
refactor: use arena to have type safe ids

### LINKED ISSUE #21: Move incompatibility identifiers outside of the incompatibility struct?

Eh2406:
closes #21.

Conceptually, this moves the id's out of the `Incompatibility`. Then it changes `usize` to an `Id<Incompatibility<P, V>>` to make sure that we only use these Ids to index into the `incompatibility_store`.

In the first commit this is accomplished by using the `id-arena` crate. That crate has some complexity to make sure at run time that an `Id<T>` is only used with the correct `Arena<T>`. We only have the one `Arena` so don't need that check. And if we get more `Arena`'s I can only imagine getting one per type.  So the second commit wraps a `Vec` ourselves. For ~80 LOC I don't think the dependency pulls its weight.

There is an alternative approach based on the `typed-arena` crate, I tested that approach in the [no-ids branch](https://github.com/pubgrub-rs/pubgrub/tree/no-ids). That crate uses some unsafe to let us use `&Incompatibility<P, V>` as our `Id` type. This is very convenient in many ways, mostly that we don't need a `Id` and a `&incompatibility_store` to get at methods of the `Incompatibility`. But it does need a lifetime to make the compiler happy, and the `'arena` is just as inconvenient to carry around as the `&incompatibility_store`. So all told I don't think it pulls its weight, as compared to the version in this PR.

The exact perf impacts depend on which variation is being tested, and on top of which other optimizations. For example this PR (commit 33735521dcbe6fbf31a78bab28b3c18ca3ca0bf0) on top of Dev (commit f8432fd84a67bc901f96939cad85ea4e3a306517 + #47 + #46).

```
                        time:   [295.60 ms 295.90 ms 296.22 ms]
                        change: [-11.279% -11.151% -11.028%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
                        time:   [16.566 ms 16.609 ms 16.654 ms]
                        change: [-8.4532% -8.1611% -7.8912%] (p = 0.00 < 0.05)
                        Performance has improved.
```
dhat-rs befor
```
dhat: Total:     20,513,896 bytes in 35,475 blocks
dhat: At t-gmax: 430,096 bytes in 108 blocks
```
dhat-rs after
```
dhat: Total:     9,220,412 bytes in 34,708 blocks
dhat: At t-gmax: 121,088 bytes in 100 blocks
```

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -13,7 +13,10 @@ use crate::internal::incompatibility::{Incompatibility, Relation};
 use crate::internal::partial_solution::{DecisionLevel, PartialSolution};
 use crate::package::Package;
 use crate::report::DerivationTree;
+use crate::solver::DependencyConstraints;
 use crate::version::Version;
+use std::ops::RangeFrom;
+
 /// Current state of the PubGrub algorithm.
 #[derive(Clone)]
 pub struct State<P: Package, V: Version> {
@@ -61,6 +64,30 @@ impl<P: Package, V: Version> State<P, V> {
             Rc::make_mut(&mut self.incompatibilities),
         );
     }
+
+    /// Add an incompatibility to the state.
+    pub fn add_incompatibility_from_dependencies(
+        &mut self,
+        package: P,
+        version: V,
+        deps: &DependencyConstraints<P, V>,
+    ) -> RangeFrom<IncompId<P, V>> {
+        let out = self.incompatibility_store.next_id()..;
+        let incompatibilities = Rc::make_mut(&mut self.incompatibilities);
+        for dep in deps {
+            Incompatibility::merge_into(
+                self.incompatibility_store
+                    .alloc(Incompatibility::from_dependency(
+                        package.clone(),
+                        version.clone(),
+                        dep,
+                    )),
+                incompatibilities,
+            );
+        }
+        out
+    }
+
     /// Check if an incompatibility is terminal.
     pub fn is_terminal(&self, incompatibility: &Incompatibility<P, V>) -> bool {
         incompatibility.is_terminal(&self.root_package, &self.root_version)
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -117,47 +117,47 @@ impl<P: Package, V: Version> State<P, V> {
         &mut self,
         incompatibility: IncompId<P, V>,
     ) -> Result<(P, IncompId<P, V>), PubGrubError<P, V>> {
-        let mut current_incompat = incompatibility;
+        let mut current_incompat_id = incompatibility;
         let mut current_incompat_changed = false;
         loop {
-            if self.incompatibility_store[current_incompat]
+            if self.incompatibility_store[current_incompat_id]
                 .is_terminal(&self.root_package, &self.root_version)
             {
                 return Err(PubGrubError::NoSolution(
-                    self.build_derivation_tree(current_incompat),
+                    self.build_derivation_tree(current_incompat_id),
                 ));
             } else {
                 let (satisfier, satisfier_level, previous_satisfier_level) = self
                     .partial_solution
                     .find_satisfier_and_previous_satisfier_level(
-                        &self.incompatibility_store[current_incompat],
+                        &self.incompatibility_store[current_incompat_id],
                         &self.incompatibility_store,
                     );
                 match satisfier {
                     Decision { package, .. } => {
                         self.backtrack(
-                            current_incompat,
+                            current_incompat_id,
                             current_incompat_changed,
                             previous_satisfier_level,
                         );
-                        return Ok((package, current_incompat));
+                        return Ok((package, current_incompat_id));
                     }
                     Derivation { cause, package } => {
                         if previous_satisfier_level != satisfier_level {
                             self.backtrack(
-                                current_incompat,
+                                current_incompat_id,
                                 current_incompat_changed,
                                 previous_satisfier_level,
                             );
-                            return Ok((package, current_incompat));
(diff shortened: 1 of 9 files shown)


mpizenberg:
I didn't take the time to look at this one this weekend sorry, I'll definitely have a look next weekend!

Eh2406:
This has a big diff. So I would like us to decide what to do with it before I work on other PRs. As rebasing over this tends to be almost as much work as the original commits.

mpizenberg:
Hi, just to make sure I'm reviewing the correct thing, this PR is the one you'd put forward right, not the code in the no-ids branch?

**Edit**: ok, reread the comments on the associated issue. Indeed I like better the typed-ids approach

Eh2406:
Yes, I think we should go with the typed-ids branch from this PR.
The no-ids is just for comparison.

mpizenberg:
On the performance front, I'm seeing 10% improvements on the elm_packages benchmark, and 2.5% improvements on large_cases, so roughly the same as you for elm and a bit lower for the synthetic data.

mpizenberg (approved):
Looks good to me! Thank you for implementing that change! I've made some tiny refactors and doc changes. I think we could still slightly improve with two changes:

1. `Incompatibility::build_derivation_tree` is now an associated function which should be moved up directly into the core module
2. `Incompatibility::from_dependencies` is allocating a `Vec` that I think we can avoid entirely (by removing that function) if we provide a method in `Arena` to get a slice or an iterator over items

Do you want to have a look at that here? we can also do that after the PR.

src/internal/incompatibility.rs
```diff
@@ -169,19 +169,12 @@ impl<P: Package, V: Version> Incompatibility<P, V> {
         let kind = Kind::DerivedFrom(incompat, satisfier_cause);
         let mut package_terms = incompatibility_store[incompat].package_terms.clone();
         let t1 = package_terms.remove(package).unwrap();
+        let statisfier_cause_terms = &incompatibility_store[satisfier_cause].package_terms;
```
Eh2406: Typo: statisfier vs satisfier

mpizenberg: Thx, fixed

Eh2406:
> `Incompatibility::build_derivation_tree` is now an associated function which should be moved up directly into the core module

That makes sense, but when giving it a try a bunch of `Incompatibility`'s internals need to be exposed. The entire function is a match on `kind`, and `Derived` requires `package_terms`. So, unless I am missing something, that feels like enough of a look for now. Not objecting to trying it again in a follow up.

mpizenberg:
> a bunch of Incompatibility's internals need to be exposed

Right! let leave it as is then

Eh2406:
> `Incompatibility::from_dependencies` is allocating a `Vec` that I think we can avoid entirely (by removing that function) if we provide a method in `Arena` to get a slice or an iterator over items

I gave it a try, and I think it worked well.

src/internal/core.rs
```diff
@@ -61,6 +64,30 @@ impl<P: Package, V: Version> State<P, V> {
...
     }
+
+    /// Add an incompatibility to the state.
+    pub fn add_incompatibility_from_dependencies(
+        &mut self,
+        package: P,
+        version: V,
+        deps: &DependencyConstraints<P, V>,
+    ) -> RangeFrom<IncompId<P, V>> {
+        let out = self.incompatibility_store.next_id()..;
```
mpizenberg: Isn't this going to be perilous if the list of dependencies is empty? The `next_id()` function doesn't seem very safe to me.

mpizenberg (commented):
I have one concern about the next_id function. I'll try another thing and let you know

src/internal/core.rs
mpizenberg: It seems to be fine actually returning an empty slice
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=89f3e3c237bc23df3e501804e08bc7f7

mpizenberg:
What do you think of this variant? I've added an `alloc_iter` function returning an `ops::Range` and removed the `next_id` method that felt like it could be used in ways leading to a crash if used as index.

Eh2406:
Yes, that is better.

mpizenberg:
It seems like removing those allocations improved again the performances. Comparing e665471 and the latest commit, I have 7.5 % improvement with both benchmarks :)

Eh2406:
Sorry I lost track of this. I like the code as is. Are we good to merge? or do we need to give it 8 more days for feedback?

mpizenberg:
I'm good with it, @aleksator seems to not have been available at all these days. @aleksator Do you mind if we merge things before the 10 days notice and you tell us when you're ready to re-enter the review process? I'm fine either way, since I'm not particularly active either currently.

aleksator:
There's never a better time than now, isn't there? Give me an hour or two to have a look.

aleksator (approved):

PR #79, opened 2021-03-09
refactor: inline find_satisfier_helper and try to simplify

Eh2406:
This PR is intended to refactor the find_satisfier code to make sure I understood it. There are bigger redesigns to data structures (`history`) that should be considered, but I did not want to try them until I was sure I understand this part. I am seeing consistent perf improvements, but that is not intended. The "unrelated" refactor to add_assignment, is salvaged from an attempt at one of the possible redesigns.

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -157,7 +157,7 @@ impl<P: Package, V: Version> State<P, V> {
                         &self.incompatibility_store[current_incompat_id],
                         &self.incompatibility_store,
                     );
-                match satisfier {
+                match satisfier.clone() {
                     Decision { package, .. } => {
                         self.backtrack(
                             current_incompat_id,
--- a/src/internal/incompatibility.rs
+++ b/src/internal/incompatibility.rs
@@ -221,6 +221,11 @@ impl<P: Package, V: Version> Incompatibility<P, V> {
         self.package_terms.iter()
     }
 
+    // The number of packages.
+    pub fn len(&self) -> usize {
+        self.package_terms.len()
+    }
+
     // Reporting ###############################################################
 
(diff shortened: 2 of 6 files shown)


mpizenberg:
In terms of performance, I see no change on the elm benchmark and a 25% improvement on the synthetic large case.

mpizenberg:
So if I'm understanding things correctly, the main algorithmic difference is the introduction of `satisfier_map` which records the index of when in history each package term of the incompatibility was satisfied. The index of the satisfier is then naturally the biggest of those indices. And the search of the previous satisfier is simplified since now we can just inspect terms that correspond to the same package as the one of the satisfier and update `satisfier_map` with its new index. The previous satisfier index is then the biggest of those updated indices.

mpizenberg (approved):
That seems like a very good change to me! Thanks @Eh2406 !

Eh2406:
We long had a discussion that this had a change in behavior.  Some relevant back links are
- https://rust-lang.zulipchat.com/#narrow/stream/260232-t-cargo.2FPubGrub/topic/partial_solution/near/240328481
- https://rust-lang.zulipchat.com/#narrow/stream/260232-t-cargo.2FPubGrub/topic/random/near/303891131
- https://rust-lang.zulipchat.com/#narrow/stream/260232-t-cargo.2FPubGrub/topic/Progress.20report/near/465855304

I want to make sure I exactly understood where the change came from, and so made a branch where all the old functions are renamed `old_`  and the new functions are renamed `new_`  so that we could assert that they were equal.  The first commit on https://github.com/Eh2406/pubgrub/pull/new/check_79_find_previous_satisfier  shows that code. Of course they were shown not to be equal. The second commit shows the very small change required to make them equal.

PR #81, opened 2021-04-03
refactor: restart the unit propagation loop on Backtrack

Eh2406:
This removes the RC added in #14. This version ends the loop early to allow us to call `self.conflict_resolution`. This had no effect on test nor on perf and a tiny improvement on mem usage. I think it is worth doing for the simplification of the code.

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -3,7 +3,7 @@
 //! Core model and functions
 //! to write a functional PubGrub algorithm.
 
-use std::{collections::HashSet as Set, rc::Rc};
+use std::collections::HashSet as Set;
 
 use crate::error::PubGrubError;
 use crate::internal::arena::Arena;
@@ -23,7 +23,7 @@ pub struct State<P: Package, V: Version> {
     root_version: V,
 
     /// TODO: remove pub.
-    pub incompatibilities: Rc<Vec<IncompId<P, V>>>,
+    pub incompatibilities: Vec<IncompId<P, V>>,
 
     /// Partial solution.
     /// TODO: remove pub.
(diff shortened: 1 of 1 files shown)


src/internal/core.rs
```diff
@@ -130,6 +123,17 @@ impl<P: Package, V: Version> State<P, V> {
                     _ => {}
                 }
             }
+            if let Some(incompat_id) = conflict_id {
+                let (package_almost, root_cause) = self.conflict_resolution(incompat_id)?;
+                self.unit_propagation_buffer.clear();
+                self.unit_propagation_buffer.push(package_almost.clone());
```
Eh2406, line 129: I originally added a `self.unit_propagation_buffer.push(current_package);`, after this line to have it look to see if there are other incompatibilities that are still relevant related to `current_package`. That would better match what the old code did. But all tests passed without that line, including comparing the `elm` file to the sat solver, so it seams not to be needed.

mpizenberg:
Sorry been busy this weekend, I'll try to review this and the other PR when I can but it may need to wait for upcoming weekend :(

src/internal/core.rs
mpizenberg: With the old code, if another incompatibility related to the current package were to satisfy the partial solution, it would clear the buffer and thus replacing the package we identified first. So I believe your approach is both correct and better reflects the note "Iterate over incompatibilities in reverse order to evaluate first the newest incompatibilities.".

mpizenberg (approved):
The key difference in this PR isn't really the RC removal (though that is nice) but the fact that we restart the unit propagation loop with a new package at the first conflict found instead of continuing evaluating the rest of incompatibilities for current package.

This is a good change! thanks!

PR #83, opened 2021-04-19
refactor: add minor code clean-up

tranzystorekk:
Highlights:

- Add `Deref` and `IntoIterator` impls for `SmallVec` references for convenience
- Shorten code where deref coercion can be applied
- Replace `&*` with explicit `.deref()` calls for readability

### DIFF
--- a/src/range.rs
+++ b/src/range.rs
@@ -93,7 +93,7 @@ impl<V: Version> Range<V> {
 
     /// Compute the complement set of versions.
     pub fn negate(&self) -> Self {
-        match self.segments.as_slice().first() {
+        match self.segments.first() {
             None => Self::any(), // Complement of ∅  is *
 
             // First high bound is +∞
@@ -110,9 +110,9 @@ impl<V: Version> Range<V> {
             // First high bound is not +∞
             Some((v1, Some(v2))) => {
                 if v1 == &V::lowest() {
-                    Self::negate_segments(v2.clone(), &self.segments.as_slice()[1..])
+                    Self::negate_segments(v2.clone(), &self.segments[1..])
                 } else {
-                    Self::negate_segments(V::lowest(), &self.segments.as_slice())
+                    Self::negate_segments(V::lowest(), &self.segments)
                 }
             }
         }
@@ -125,7 +125,7 @@ impl<V: Version> Range<V> {
     fn negate_segments(start: V, segments: &[Interval<V>]) -> Range<V> {
         let mut complement_segments = SmallVec::empty();
         let mut start = Some(start);
-        for (v1, some_v2) in segments.iter() {
+        for (v1, some_v2) in segments {
             // start.unwrap() is fine because `segments` is not exposed,
             // and our usage guaranties that only the last segment may contain a None.
             complement_segments.push((start.unwrap(), Some(v1.to_owned())));
@@ -144,7 +144,7 @@ impl<V: Version> Range<V> {
 
     /// Compute the union of two sets of versions.
     pub fn union(&self, other: &Self) -> Self {
-        (self.negate().intersection(&other.negate())).negate()
+        self.negate().intersection(&other.negate()).negate()
     }
 
     /// Compute the intersection of two sets of versions.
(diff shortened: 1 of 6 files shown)


Eh2406:
Thanks for the PR! 

Just a few questions:
I like the use of `Deref` on `SmallVec`! Can we double down on that by inlining and removing `as_slice`?
I would have expected `IntoIterator` to have `Item = T;`. Now that we have `Deref`, can we just call `smallvec.iter()` to get an iterator of `&T`? That seems less surprising; if a little more verbose.
Can you elaborate on why you fined `.deref()` better to read then `&*`? Not that I always like Clippys opinions, but it has a [(pedantic) lint](https://rust-lang.github.io/rust-clippy/master/#explicit_deref_methods) to go the other way. 

Over all I like the direction!

tranzystorekk:
> Can you elaborate on why you fined `.deref()` better to read then `&*`? Not that I always like Clippys opinions, but it has a [(pedantic) lint](https://rust-lang.github.io/rust-clippy/master/#explicit_deref_methods) to go the other way.

Huh, I didn't know clippy had that lint, but I guess that's why it's pedantic 😄 I know it's a matter of opinion, but I think the deref coercion, although slick, is a very confusing and not well understood mechanism. The verbose `.deref()` call clearly says what is done when we use it in the `match` expressions.

> Can we double down on that by inlining and removing as_slice?

We can remove the `SmallVec::iter()` method and delegate to `slice::iter()` with deref, but `IntoIterator` is mostly for all those places where I replaced `for x in smallvec.iter() { ... }` with `for x in &smallvec { ... }`. Do you think those should be reverted to the verbose version?

Also, `SmallVec::as_slice()` cannot be removed, because deref coercion doesn't work in `match` expressions, i.e. all those places where we do

```rs
match smallvec.as_slice() {
    [x] => ...
    ...
}
```

EDIT: I guess we could call `.deref()` in such `match`es, but in this case I think `as_slice()` conveys what we want to match against more clearly.

Eh2406 (approved):
`as_slice` and `deref` both have context where they are clearer. I am convinced, have them both.
As to `IntoIterator`, I miss read your PR. It is on `&'a SmallVec<T>`, so `Item = &'a T` makes sense. All good, I like it now.
For `.deref()` vs `&*`, I don't have a strong opinion. Ether way is fine. I will leave it to @mpizenberg.

So sorry for doing to fast a review earlier. I think this is good.

Our normal prosses on this project is that before we merge we give 10 days for less active project members to put in a comment. So please be a little patient.

src/range.rs
```diff
@@ -288,11 +284,10 @@ impl<V: Version> fmt::Display for Range<V> {
     }
 }
 
-fn interval_to_string<V: Version>(interval: &Interval<V>) -> String {
-    match interval {
-        (start, Some(end)) => format!("[ {}, {} [", start, end),
-        (start, None) => format!("[ {}, ∞ [", start),
-    }
+fn interval_to_string<V: Version>((start, end): &Interval<V>) -> String {
```
mpizenberg: The destructuring of `(start, end)` is nice but for the body of the function, I'd prefer if we keep a match. We can match on just `end` though.

src/report.rs
```diff
@@ -3,7 +3,10 @@
 //! Build a report as clear as possible as to why
 //! dependency solving failed.
 
-use std::fmt;
+use std::{
```
mpizenberg: I'd prefer if we keep two lines here to keep the style consistent with the rest of the crate.
```rust
use std::fmt;
use std::ops::{Deref, DerefMut};
```

mpizenberg (changes requested):
Thanks! I don't have any strong opinion on the deref style, just as long as we keep clippy happy :) I generally like the changes! Have a few nitpicks though if you can address them.

Also don't hesitate to simply push new commits as it's easier to review after an initial review instead of force push. We squash merge anyway so there will be only one commit left.

src/range.rs
tranzystorekk: Done ✅

src/report.rs
tranzystorekk: Done ✅

mpizenberg (approved):

mpizenberg (approved):

Eh2406:
Thank you for the clean up!

PR #84, opened 2021-04-23
perf: index incompatibilities by package

Eh2406:
This optimization is documented in the original description of the algorithm.

> While we could iterate over every incompatibility over and over until we can't find any more derivations, this isn't efficient when many of them represent dependencies of packages that are currently irrelevant. Instead, we index them by the names of the packages they refer to ...

https://github.com/dart-lang/pub/blob/master/doc/solver.md

Thanks to #81 it is now a small change. Unfortunately, it does not make a difference for the `large_case` and `elm` benchmarks. The [`zuse`](https://github.com/pubgrub-rs/pubgrub/issues/44#issuecomment-730721051) benchmark does improved by a lot ~40%, as does a synthetic benchmark where `A->B->C->E->F->G` where the first thing that is tried works. So I think it will be useful on bigger use cases. But I understand that I don't have strong evidence.

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -61,9 +61,7 @@ impl<P: Package, V: Version> State<P, V> {
     /// Add an incompatibility to the state.
     pub fn add_incompatibility(&mut self, incompat: Incompatibility<P, V>) {
         let id = self.incompatibility_store.alloc(incompat);
-        for (a, _) in self.incompatibility_store[id].iter() {
-            Incompatibility::merge_into(id, self.incompatibilities.entry(a.clone()).or_default());
-        }
+        self.merge_into(id);
     }
 
     /// Add an incompatibility to the state.
@@ -81,12 +79,7 @@ impl<P: Package, V: Version> State<P, V> {
             }));
         // Merge the newly created incompatibilities with the older ones.
         for id in IncompId::range_to_iter(new_incompats_id_range.clone()) {
-            for (a, _) in self.incompatibility_store[id].iter() {
-                Incompatibility::merge_into(
-                    id,
-                    self.incompatibilities.entry(a.clone()).or_default(),
-                );
-            }
+            self.merge_into(id);
         }
         new_incompats_id_range
     }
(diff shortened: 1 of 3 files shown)


src/internal/core.rs
```diff
@@ -22,8 +23,7 @@ pub struct State<P: Package, V: Version> {
     root_package: P,
     root_version: V,
 
-    /// TODO: remove pub.
-    pub incompatibilities: Vec<IncompId<P, V>>,
+    incompatibilities: Map<P, Vec<IncompId<P, V>>>,
```
mpizenberg, line 26: Alright so incompatibility ids are now present in as many Map entries as its number of terms. So this means slightly more memory (likely 2x for `incompatibilities`) but an accelerated access. This accelerated access is likely proportional to the total number of different packages. So this is likely to improve performances in cases where packages have many dependencies.

mpizenberg: However this also means that updating `incompatibilities` is now more complex since instead of one array push, we have one array push per key that needs to be modified in the hashmap.

src/internal/core.rs
```diff
@@ -79,7 +81,12 @@ impl<P: Package, V: Version> State<P, V> {
             }));
         // Merge the newly created incompatibilities with the older ones.
         for id in IncompId::range_to_iter(new_incompats_id_range.clone()) {
-            Incompatibility::merge_into(id, &mut self.incompatibilities);
+            for (a, _) in self.incompatibility_store[id].iter() {
```
mpizenberg: We might want to dedup this with the code inside `add_incompatibility()` method.

src/internal/core.rs
```diff
@@ -98,12 +105,9 @@ impl<P: Package, V: Version> State<P, V> {
             // Iterate over incompatibilities in reverse order
             // to evaluate first the newest incompatibilities.
             let mut conflict_id = None;
-            for &incompat_id in self.incompatibilities.iter().rev() {
+            // We only care about incompatibilities if it contains the current package.
+            for &incompat_id in self.incompatibilities[&current_package].iter().rev() {
```
mpizenberg, line 102: Alright the big advantage is here!

src/internal/core.rs
```diff
@@ -204,7 +208,12 @@ impl<P: Package, V: Version> State<P, V> {
         self.partial_solution
             .backtrack(decision_level, &self.incompatibility_store);
         if incompat_changed {
-            Incompatibility::merge_into(incompat, &mut self.incompatibilities);
+            for (a, _) in self.incompatibility_store[incompat].iter() {
```
mpizenberg: same here regarding dedup with the code inside `add_incompatibility()` method

src/internal/incompatibility.rs
```diff
@@ -52,19 +52,16 @@ enum Kind<P: Package, V: Version> {
...
 #[derive(Eq, PartialEq)]
-pub enum Relation<P: Package, V: Version> {
+pub enum Relation<P: Package> {
     /// We say that a set of terms S satisfies an incompatibility I
     /// if S satisfies every term in I.
     Satisfied,
     /// We say that S contradicts I
     /// if S contradicts at least one term in I.
-    Contradicted(PackageTerm<P, V>),
+    Contradicted(P),
```
mpizenberg, line 64: This change is unrelated right? I suppose it's fine if we actually don't need the `Term` in that variant. We might want to verify though that it isn't the source of the perf improvement instead of the hash map changes. Seems very unlikely but things are sometimes surprising.

mpizenberg (approved):
I made a few remarks that are in separate comments but I think this is a good change. It should degrade peak performances due to the added complexity of the Hashmap but it should also largely improve worst case scenarios where there are many different packages in incompatibilities.

src/internal/core.rs
Eh2406: Yes this is where the savings happens, on the `zuse` benchmark a lot of time was spend in `current_incompat.get`. Bye keeping them indexed we can jump straight to the relevant ones without having to check them all.

src/internal/incompatibility.rs
Eh2406: Oops yes this is unrelated. I can back it out if you don't want it, or don't want it in this PR. I have bean running all tests using this commit as a baseline, just incase the removal of the one `Range<V>.clone()` mattered. But it does not make a perf difference from my measurements.

src/internal/core.rs
Eh2406: Good point, I will do it next time I sit down to code.

src/internal/incompatibility.rs
mpizenberg: alright we can keep the change in then

mpizenberg:
I just did some small renaming. `merge_into` was adequate when `self` was an incompatibility and the argument was the store, but now that `self` is the algorithm state and the argument is the incompatibility id, I though `merge_incompatibility` makes more sense.

Also renamed `for (a, _) in ...` into `for (pkg, _term)` because I needed a reminder of what those things were to better grasp what this does at a glance.

mpizenberg (approved):

Eh2406:
After this I think I am out of things that I want to see happen before a `0.2.1`. What do you think about releasing after this merges?

mpizenberg:
> After this I think I am out of things that I want to see happen before a `0.2.1`. What do you think about releasing after this merges?

Yes sounds like a good plan to me! The only painful thing will be updating the changelog. I used to make it a habit to add things to the unreleased section with each PR but we forgot to do it for all the recent PRs. I hope the PR discussions will make it easier.

Eh2406:
I just stumbled across another ~10% improvement. So when that and #87 land, then a release. Sorry for the delays.

As to the change log, we can go into more depth if we want, but as there is no change to the api I think a perf run comparing `0.2.0` to `0.2.1` would be sufficient.

Issue #33, opened 2020-10-11
perf: keep directly the intersection of derivations in Memory


mpizenberg:
Memory stores a `Vec` of derivations per package. If we are to evaluate potential packages in a smarter way #32 , we need the intersection of derivation terms.

There are two methods in `Memory` that can return iterators to individual assignment terms, `terms_for_package` and `potential_packages`. The latter is the use case we are discussing for performance improvements in #32 . And the former is to compute relations in incompatibilities that eventually ends up computing the intersection of those terms. Considering that we seem to spend much time in relation computation #27 , this could be a non-negligible speedup.

Eh2406:
One thing I was wondering about is how this works with backtracking. If we backtrack to before a decision that introduced that derivation then does that mean that some terms need to come out of that intersection? If so how do we do that? Even if this is a problem, it is not insurmountable, we can keep the `Vec` of derivations and a pre computed intersection. In the case of backtracking, the vec gets smaller and the intersection needs to be recomputed.

mpizenberg:
That's a very relevant question! Backtracking is currently operated [thanks to `history`](https://github.com/mpizenberg/pubgrub-rs/blob/02a85b53607b3f4c62c0922f24b0708851ff5f40/src/internal/partial_solution.rs#L71) which is a chronological vec of all the assignments. Once we backtrack to a given level, we [iterate on assignments with a lower level and rebuild the memory](https://github.com/mpizenberg/pubgrub-rs/blob/02a85b53607b3f4c62c0922f24b0708851ff5f40/src/internal/partial_solution.rs#L87).

Conflict resolution is not common I think in a normal setting since newer versions tend to be compatible with older ones, and we choose newer versions by default. Also, once we've found the root cause for a conflict, I don't see a reason we should have another conflict soon. ~But backtracking happens once per prior cause until the root cause is found so it could typically happen multiple times until we've found the root cause~. **<- woops, that is wrong, there is 1 backtrack per conflict resolution**

It's probably worth having two things for memory derivations, a precomputed intersection, and terms that have not been intersected yet in a complementary vec. Something like this in the `Memory.assignments`.
```rust
struct PackageAssignments<V: Version> {
    decision: Option<(V, Term<V>)>,
    derivations_intersected: Term,
    derivations_not_intersected_yet: Vec<Term<V>>,
}
```
[...]

PR #87, opened 2021-05-01
perf: remove not_intersected_yet

Eh2406:
This includes #84, the only new part is the last commit. It is intended to be merged after it.

For a Range an intersection is almost as fast as a clone. By removing the not_intersected_yet optimization we do more intersections and less clones and allocations. This simplifies the code for `term_intersection_for_package` witch is in the hotest part of the of `unit_propagation`.

I was trying this because of how much it simplified the code. It would be worth doing even if it was perf neutral, but I am consistently seeing small perf improvements!

```
Benchmarking large_cases/elm-packages_str_SemanticVersion.ron: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 20.0s. You may wish to increase target time to 28.2s, or reduce sample count to 70.
                        time:   [279.06 ms 279.37 ms 279.67 ms]
                        change: [-6.6904% -6.5466% -6.4031%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
                        time:   [10.604 ms 10.634 ms 10.670 ms]
                        change: [-10.819% -10.537% -10.273%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high severe
Benchmarking large_cases/zuse_str_SemanticVersion.ron: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 20.0s. You may wish to increase target time to 38.2s, or reduce sample count to 50.
                        time:   [380.02 ms 380.46 ms 380.92 ms]
                        change: [-8.5621% -8.3803% -8.2179%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) low mild
  5 (5.00%) high mild
  2 (2.00%) high severe
```

### DIFF
--- a/src/internal/incompatibility.rs
+++ b/src/internal/incompatibility.rs
@@ -143,32 +143,6 @@ impl<P: Package, V: Version> Incompatibility<P, V> {
         }
     }
 
-    /// CF definition of Relation enum.
-    pub fn relation(&self, mut terms: impl FnMut(&P) -> Option<Term<V>>) -> Relation<P> {
-        let mut relation = Relation::Satisfied;
-        for (package, incompat_term) in self.package_terms.iter() {
-            match terms(package).map(|term| incompat_term.relation_with(&term)) {
-                Some(term::Relation::Satisfied) => {}
-                Some(term::Relation::Contradicted) => {
-                    return Relation::Contradicted(package.clone());
-                }
-                None | Some(term::Relation::Inconclusive) => {
-                    // If a package is not present, the intersection is the same as [Term::any].
-                    // According to the rules of satisfactions, the relation would be inconclusive.
-                    // It could also be satisfied if the incompatibility term was also [Term::any],
-                    // but we systematically remove those from incompatibilities
-                    // so we're safe on that front.
-                    if relation == Relation::Satisfied {
-                        relation = Relation::AlmostSatisfied(package.clone());
-                    } else {
-                        relation = Relation::Inconclusive;
-                    }
-                }
-            }
-        }
-        relation
-    }
-
     /// Check if an incompatibility should mark the end of the algorithm
     /// because it satisfies the root package.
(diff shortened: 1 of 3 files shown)


mpizenberg:
So this seems like a follow up to #33 that was done quite a while ago.

I feel like this could go even further and simplify some of the parameters around the `relation` function. Currently, we have `pub fn relation(&self, terms: impl Fn(&P) -> Option<&'a Term<V>>)`. If I remember correctly, the name `terms` comes from the fact that it used to be a function returning an iterator or a slice of terms. Then we decided to precompute terms intersection, but only when necessary, while the rest of the time terms are accumulated in a vector `not_intersected_yet`, awaiting to be useful. And when we decided that, we only had the resulting intersection to pass to the function, thus `Option<Term<V>>`. But now that intersection are systematically computed, I feel like this could be turned into `Fn(&P) -> &Term<V>` without the option. And when there is no term, the intersection is anything (was it `Term::any()?` I forgot). The point is we could have that defined as a static reference to a constant term.

All that is just gut feeling based on what I remember of all this. I think you have all this in mind better than me now so I hope you'll see what I'm trying to say ^^.

Eh2406:
You are correct. `Option<&'a Term<V>>` can be replaced with `&'a Term<V>` by `unwrap_or(&Term::any())`. ( Borrow checker makes thinks a little more complicated, but it is not relevant. ) And I was about to push it, but let me be thorough and run new perf numbers ... and that is not good. By returning `None` we skip calling `relation_with` witch is important for performance.

mpizenberg (approved):

mpizenberg:
You're right! I forgot about the cost of that call. Seems great to me then!

PR #92, opened 2021-05-23
Unify history and memory in partial_solution

mpizenberg:
This is a cleaned up version of PR #91.

The main point of this PR that the `history` and `memory` fields of the partial solution have been merged into a single hashmap. As a result, we don't have a single `Vec` history of all assignments, but one `Vec` per package. Among the changes, that vec now only store derivations since we know for certain that a decision can only be in the last position. It is thus automatically deduced by the terms intersection which is now a new enum, either a decision or the intersection of derivations.

This change was introduced because I thought it would improve conflict resolution performances by avoiding to browse the whole history. In practice, I don't see a performance change in the elm, synthetic and zuse benchmarks so either there is not enough conflict to measure it, or it doesn't improve much. Might be interesting to see if it helps reduce worst case scenarios though.


Eh2406:
I tried using `retain` and `SmallVec` and I think the perf is better. What do you think?
Also we should try after rebasing on dev.

mpizenberg:
Retain is much nicer! and smallvec usage should be better since those vec tend to be small. In practice, for the benchs I have setup (elm, synthetic, zuse) I don't see a performance change bigger than the variations I observe for each run. But I like those changes!

Eh2406:
This makes backtracking `O(packages)` instead of `O(assignments)`, that is a big step foreword. Ideally we would get to something like `O(min(decision_level, current_decision_level - decision_level))` witch may bring back some kind of `history` like construct. But this seams like a far more flexible base to start from, I can think of several interesting ways to augment this structure.

So I am +1 when I have had a chance to review.

src/internal/partial_solution.rs
```diff
@@ -1,103 +1,201 @@
...
+#[derive(Clone, Debug)]
+enum AssignmentsIntersection<V: Version> {
+    Decision((u32, V, Term<V>)),
+    Derivations(Term<V>),
+}
+
+/// An assignment is either a decision: a chosen version for a package,
+/// or a derivation : a term specifying compatible versions for a package.
+/// We also record the incompatibility at the origin of a derivation, called its cause.
+/// TODO: Remove. I only added that to not change the code in core.rs (except reference &).
```
Eh2406, line 62: I don't understand this TODO.

Eh2406 (approved):
Looking it over carefully I don't see anything wrong. So Lets go with it. If we find a problem we can always change it then.

src/internal/partial_solution.rs
mpizenberg: That comment was here because I wanted to reduce the size of the git diff for now but not forget to improve the code by removing that type later, because actually, we don't really need that type. With the history/memory unification, it's only useful in one place now, as the return type of the `find_satisfier_and_previous_satisfier_level` function. And by keeping the same return type, we don't have to modify the code using it in the `core` module.

But in fact, I think it would be cleaner to remove that `Assignment` type and change the return type of that function to something like `(&P, SatisfierSearch)` where that type would be defined with something like
```rust
enum SatisfierSearch<P, V> {
    DifferentDecisionLevels { previous_satisfier_level: DecisionLevel },
    SameDecisionLevels { satisfier_cause: IncompId<P, V> },
}
```

Eh2406: looks to me like we need it to get passed to `prior_cause`, but we can leave it for latter.

PR #93, opened 2021-05-24
A tiny cleanup change

mpizenberg:


Eh2406 (approved):
I did some premature optimization, and reduced the calls to `clone` to only happen once and only in the `None` case.

It looks good to me!

PR #88, opened 2021-05-04
perf: don't reuse the same incompatibility repeatedly

Eh2406:
This includes #87, the only new part is the last commit. It is intended to be merged after it.

Most of the time `relation` for an incompatibility (like `A` depends on `B`) is called at least twice, once when `A` gets added and when `B` gets added. After the first call we get `AlmostSatisfied` and change the `partial_solution` so that it will be contradicted the next time. This adds a HashSet to keep track of all the `IncompId`s we know will stay contradicted until the next backtrack. `IncompId` are just a small integer so hash very quickly.

It is possible to figure out witch `IncompId`s are still used after a backtrack, but so far nothing that is pulls its weight. 

```
Warning: Unable to complete 100 samples in 20.0s. You may wish to increase target time to 23.6s, or reduce sample count to 80.
large_cases/elm_str_SemanticVersion.ron
                        time:   [244.05 ms 244.40 ms 244.79 ms]
                        change: [-9.0431% -8.8780% -8.7172%] (p = 0.00 < 0.05)
                        Performance has improved.
large_cases/large_case_u16_NumberVersion.ron
                        time:   [9.3124 ms 9.3213 ms 9.3302 ms]
                        change: [-13.284% -13.164% -13.049%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking large_cases/zuse_str_SemanticVersion.ron: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 20.0s. You may wish to increase target time to 35.2s, or reduce sample count to 50.
large_cases/zuse_str_SemanticVersion.ron
                        time:   [351.46 ms 351.77 ms 352.08 ms]
                        change: [-12.531% -12.418% -12.310%] (p = 0.00 < 0.05)
                        Performance has improved.
```

PR #94, opened 2021-05-24
Remove contradicted incompatibilities

mpizenberg:
Inspired by the `contradicted_incompatibilities` change that was recently introduced in #88 I wanted to see if instead of checking for each incompat if it is in `contradicted_incompatibilities`, we could simply remove those contradicted incompats and only add them again when backtracking.

It works, but it's slower. I think because of few downsides:
1. I changed `self.incompatibilities` from a `Map<P, Vec<Id>>` to a `Map<P, Set<Id>>` to be able to remove contradicted incompats
2. We need to transfer incompats from `self.contradicted...` to `self.incompat...` when backtracking.
3. I changed `self.contradicted...` type from `Set<id>` to `Map<P, Set<Id>>` and so we don't know in Id was already contradicted when evaluating the other packages of the same incompat.

Regardless of the fact that it's slower, I thought it was cool enough to share (but not to merge)

### DIFF
--- a/src/internal/arena.rs
+++ b/src/internal/arena.rs
@@ -91,6 +91,10 @@ impl<T> Arena<T> {
         Arena { data: Vec::new() }
     }
 
+    pub fn len(&self) -> usize {
+        self.data.len()
+    }
+
     pub fn alloc(&mut self, value: T) -> Id<T> {
         let raw = self.data.len();
         self.data.push(value);
--- a/src/internal/arena.rs
+++ b/src/internal/arena.rs
@@ -1,4 +1,5 @@
 use std::{
+    cmp::Ordering,
     fmt,
     hash::{Hash, Hasher},
     marker::PhantomData,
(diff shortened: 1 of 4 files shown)


Eh2406:
I think there is something like this that can work. ("filter out items faster than a `HashSet.contains`" in https://github.com/pubgrub-rs/pubgrub/pull/88#issuecomment-833780223) but I think the big thing is your 3. The big savings if from incompats like `a depends on b`, where we only have to check it for `a`, as when we get to `b` we already know it is contradicted. Definitely worth continued exploration. Thanks for sharing!

Eh2406:
I had some time to experiment. I found some variations that broke even, but nothing that was an overall improvement. As other parts get faster, the opportunity's here may become clearer. Or we may think of other structures with even less overhead. Definitely worth continued exploration.

mpizenberg:
Indeed, let's leave this type of change for later, if it still makes sense at that time.

mpizenberg:
Another idea that turned out to be not as performant as expected. The code base has changed quite a bit since then. So let's just discard this and as you say, we'll likely identify clearer opportunities later.

PR #107, opened 2021-08-07
Logging to help debugging

mpizenberg:
From my experiments with the interval design, I realized that being able to log the important steps of the solver was quite helpful to locate the origin of bugs. In this PR, I've introduced the `log` crate and added some calls to `log::info!()` in the solver. I've also added `env_logger` to the dev dependencies to be able to display logs in failing tests.

After running the benchmarks, I think this has an impact on the zuse bench where I saw roughly 10% slower performances. I did not notice performance changes on the other benchmarks. I wonder what it is though that could impact performances. I thought the `log` crate did not impact performances if it is not instantiated with a logger. Could this be because the few `Display` implementations that I added on some internal types?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ include = ["Cargo.toml", "LICENSE", "README.md", "src/**", "tests/**", "examples
 thiserror = "1.0"
 rustc-hash = "1.1.0"
 serde = { version = "1.0", features = ["derive"], optional = true }
-log = { version = "0.4.14", default-features = false } # for debug logs in tests
+log = "0.4.14" # for debug logs in tests
 
 [dev-dependencies]
 proptest = "0.10.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,12 +23,14 @@ include = ["Cargo.toml", "LICENSE", "README.md", "src/**", "tests/**", "examples
 thiserror = "1.0"
 rustc-hash = "1.1.0"
 serde = { version = "1.0", features = ["derive"], optional = true }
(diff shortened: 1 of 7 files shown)


Eh2406:
We should try to track down the perf, but the logging is important and it is worth the change.

src/internal/partial_solution.rs
```diff
@@ -32,6 +35,24 @@ pub struct PartialSolution<P: Package, V: Version> {
...
 }
 
+impl<P: Package, V: Version> Display for PartialSolution<P, V> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let assignments: BTreeSet<String> = self
+            .package_assignments
+            .iter()
+            .map(|(p, pa)| format!("{}: {}", p, pa))
+            .collect();
+        let assignments: Vec<_> = assignments.into_iter().collect();
```
Eh2406: I think it would be clearer to collect into a vec and then call sort.

Cargo.toml
```diff
@@ -23,12 +23,14 @@ include = ["Cargo.toml", "LICENSE", "README.md", "src/**", "tests/**", "examples
 thiserror = "1.0"
 rustc-hash = "1.1.0"
 serde = { version = "1.0", features = ["derive"], optional = true }
+log = { version = "0.4.14", default-features = false } # for debug logs in tests
```
Eh2406: No objection, just wondering:
What features are we trying to opted out of? Why `default-features = false`?

Eh2406:
I did not see any pref impact. So I addressed my nits, and made clippy happy. I will merge when CI is green.

PR #109, opened 2021-08-23
Use arrays to provide dependency lists

tranzystorekk:
Use arrays instead of vecs in examples, docs and tests (IntoIterator implemented since Rust 1.53)


mpizenberg (approved):
Looks nicer indeed! Thanks for the change!

Issue #120, opened 2022-10-13
Additional Constraints


baszalmstra:
Im (still) working on using PubGrub to solve Conda packages. I've made a lot of progress but currently run into the problem of "constraints".

Conda packages can have two types of dependencies: regular dependencies and constraints. Regular dependencies are packages that need to be included in the solution but constraints are dependencies that don't have to be included, but if they are (via the dependency of another package), have to adhere to the constraints. I think this is similar to peer dependencies in NodeJS. 

I initially modeled these as regular dependencies which I could strip from the solution after solving. However, sometimes these constraints are used to ensure a certain other package is *not* selected. For instance, python packages that are incompatible with `pypi` often have the constraint: `pypy <0a0`. There is no version compatible with this since it denotes the lowest possible version. Using this as a dependency results in an error `DependencyOnTheEmptySet` which makes sense: that's exactly what the user intends.

Is this something that could be supported with pubgrub? Is this something I could potentially add?

Eh2406:
For some context #36 especially [this comment](https://github.com/pubgrub-rs/pubgrub/issues/36#issuecomment-709400796) and the PR #55.

Our surface level interface does not have a way to express this. It can only say "A depends on B" meaning "do not build the solution where a version of A is selected and a version of B is not also selected". However this is just a limitation of our interface. The internal [Incompatibility](https://github.com/pubgrub-rs/pubgrub/blob/release/src/internal/incompatibility.rs#L33) data structure can happily express more complicated constraints like "do not build the solution where a version of A is selected and a version of B is selected". (Along with arbitrarily more exotic things.)

So far we've done a pretty good job of having a user-friendly public interface and documenting it at the module level, package level, and book level. I would be open to designing a more flexible interface as long as it continued to be easy to use. Alternatively we could design a lower level interface as long as we have a good pattern for how straightforward use cases can be done easily.
As you've noticed no one has had time to work on this project given its existing level of polish, so maybe we should lower our standards to let you get unstuck.

mpizenberg:
PS, there is also an optimisation on the conflict resolution backtracking
algo introduced by Jacob that is not exactly what we should do but is
faster and equivalent considering our interface to build constraints. If we
allow more expressive constraints. We need check that this optimisation is
still correct. And we also need to adapt the generators for the property
tests.

mpizenberg:
This comment and the few following it are what I had in mind in my previous comment:
[...]

PR #124, opened 2022-10-28
feat: constraining dependencies

baszalmstra:
This is an implementation of "constraining" dependencies or sometimes called "peer dependencies" as described in https://github.com/pubgrub-rs/pubgrub/issues/120.

The implementation is fully functioning. I've added some regular tests as well as modified the proptests to handle constraints.

I did have to modify the API of `DependencyProvider` and of `OfflineDependencyProvider`. I'm not really happy with the names and stuff but I'm curious to get your feedback.

I've also already used this branch in my Conda package solver and it works like a charm there! 🥳

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -15,6 +15,7 @@ use crate::internal::partial_solution::{DecisionLevel, PartialSolution};
 use crate::internal::small_vec::SmallVec;
 use crate::package::Package;
 use crate::report::DerivationTree;
+use crate::solver::Requirement;
 use crate::type_aliases::{DependencyConstraints, Map};
 use crate::version_set::VersionSet;
 
@@ -75,14 +76,25 @@ impl<P: Package, VS: VersionSet> State<P, VS> {
         &mut self,
         package: P,
         version: VS::V,
-        deps: &DependencyConstraints<P, VS>,
+        deps: &DependencyConstraints<P, Requirement<VS>>,
     ) -> std::ops::Range<IncompId<P, VS>> {
         // Create incompatibilities and allocate them in the store.
-        let new_incompats_id_range = self
-            .incompatibility_store
-            .alloc_iter(deps.iter().map(|dep| {
-                Incompatibility::from_dependency(package.clone(), version.clone(), dep)
-            }));
+        let new_incompats_id_range =
+            self.incompatibility_store
+                .alloc_iter(deps.iter().map(|(dep, requirement)| match requirement {
+                    Requirement::Required(range) => Incompatibility::from_required_dependency(
+                        package.clone(),
+                        version.clone(),
+                        (dep, range),
+                    ),
+                    Requirement::Constrained(range) => {
+                        Incompatibility::from_constrained_dependency(
+                            package.clone(),
+                            version.clone(),
+                            (dep, range),
+                        )
+                    }
+                }));
         // Merge the newly created incompatibilities with the older ones.
         for id in IncompId::range_to_iter(new_incompats_id_range.clone()) {
             self.merge_incompatibility(id);
--- a/src/internal/incompatibility.rs
+++ b/src/internal/incompatibility.rs
@@ -105,13 +105,26 @@ impl<P: Package, VS: VersionSet> Incompatibility<P, VS> {
     }
 
     /// Build an incompatibility from a given dependency.
-    pub fn from_dependency(package: P, version: VS::V, dep: (&P, &VS)) -> Self {
+    pub fn from_required_dependency(package: P, version: VS::V, dep: (&P, &VS)) -> Self {
         let set1 = VS::singleton(version);
         let (p2, set2) = dep;
         Self {
             package_terms: SmallMap::Two([
                 (package.clone(), Term::Positive(set1.clone())),
                 (p2.clone(), Term::Negative(set2.clone())),
+            ]),
+            kind: Kind::FromDependencyOf(package, set1, p2.clone(), set2.clone()),
+        }
+    }
+
+    /// Build an incompatibility from a given constraint.
+    pub fn from_constrained_dependency(package: P, version: VS::V, dep: (&P, &VS)) -> Self {
+        let set1 = VS::singleton(version);
+        let (p2, set2) = dep;
+        Self {
+            package_terms: SmallMap::Two([
+                (package.clone(), Term::Positive(set1.clone())),
+                (p2.clone(), Term::Positive(set2.complement())),
             ]),
             kind: Kind::FromDependencyOf(package, set1, p2.clone(), set2.clone()),
         }
--- a/src/solver.rs
+++ b/src/solver.rs
@@ -166,7 +166,10 @@ pub fn resolve<P: Package, VS: VersionSet>(
                             version: v.clone(),
                         });
                     }
-                    if let Some((dependent, _)) = x.iter().find(|(_, r)| r == &&VS::empty()) {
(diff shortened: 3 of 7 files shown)


mpizenberg:
> I've also already used this branch in my Conda package solver and it works like a charm there!

Awesome! I can't review this weekend but I'll have some time during the week and will happily give you some feedback

src/internal/incompatibility.rs
```diff
@@ -117,6 +117,19 @@ impl<P: Package, VS: VersionSet> Incompatibility<P, VS> {
...
+    /// Build an incompatibility from a given constraint.
+    pub fn from_constrained_dependency(package: P, version: VS::V, dep: (&P, &VS)) -> Self {
+        let set1 = VS::singleton(version);
+        let (p2, set2) = dep;
+        Self {
+            package_terms: SmallMap::Two([
+                (package.clone(), Term::Positive(set1.clone())),
+                (p2.clone(), Term::Positive(set2.complement())),
+            ]),
+            kind: Kind::FromDependencyOf(package, set1, p2.clone(), set2.clone()),
```
Eh2406, line 129: Currently weather it is constrained or required does not appear in the kind at all. This does not affect the correctness of the algorithm, as kind is only used for error messages. Do we think it's important to have this distinction available when generating error messages?

- Yes, you can't understand the proof argument created by this algorithm unless you know what the inputs are. Therefore we should add a boolean or enum to `Kind::FromDependency` to distinguish which kind of requirement it was.
- No, they are not significantly different. If a resolver uses both of them and wants to distinguish it can look out this dependency in its source of truth and figure out which kind it had generated.

Allso, is `FromDependencyOf` still the correct terminology, Is `FromRequirementOf` better?

src/solver.rs
```diff
@@ -215,7 +218,17 @@ pub enum Dependencies<P: Package, VS: VersionSet> {
     /// Package dependencies are unavailable.
     Unknown,
     /// Container for all available package versions.
-    Known(DependencyConstraints<P, VS>),
+    Known(DependencyConstraints<P, Requirement<VS>>),
+}
+
+/// An enum that defines the type of dependency.
+#[derive(Clone, Debug)]
+pub enum Requirement<VS: VersionSet> {
```
Eh2406, line 226: this will need to `#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]`, and it's going to need to be even a little more complicated than that. because it needs to deserialize the existing test files and benchmarks, which do not have a marker for which kind it is.

tests/sat_dependency_provider.rs
```diff
@@ -79,19 +79,37 @@ impl<P: Package, VS: VersionSet> SatResolve<P, VS> {
...
+
+                        // ^ `p` is not active
+                        cnf.add_clause(&matches);
+                    }
+                    Requirement::Constrained(range) => {
+                        for (p1_version, p1_version_var) in
+                            all_versions_by_p.get(p1).unwrap_or(&empty_vec).iter()
+                        {
+                            if !range.contains(p1_version) {
+                                // Either this specific variant is not selected or non-matching dependency versions are not selected.
```
Eh2406, line 107: On my third time reading it. Yes, I think this is correct.

src/solver.rs
```diff
@@ -338,14 +352,76 @@ impl<P: Package, VS: VersionSet> OfflineDependencyProvider<P, VS> {
...
+            entries.insert(package, Requirement::Constrained(constraint));
+        }
+    }
+
+    /// Registers the dependencies of a package and version pair.
+    /// Dependencies must be added with a single call to
+    /// [add_dependencies](OfflineDependencyProvider::add_dependencies),
+    /// [add_constraints](OfflineDependencyProvider::add_constraints), or
+    /// [add_requirements](OfflineDependencyProvider::add_requirements).
+    /// All subsequent calls to any of those functions for a given package version pair will replace the
```
Eh2406, line 403: This is no longer true. At a minimum these comments need to be updated. These calls now append items to an existing list.
@mpizenberg, do you have a strong preference for the all at once API?

Eh2406 (changes requested):
Thank you! This is looking good!
I had a few comments and concerns. Allso tests and clippy are unhappy.
But we are very close.

src/internal/core.rs
```diff
@@ -75,14 +76,25 @@ impl<P: Package, VS: VersionSet> State<P, VS> {
...
     ) -> std::ops::Range<IncompId<P, VS>> {
         // Create incompatibilities and allocate them in the store.
-        let new_incompats_id_range = self
-            .incompatibility_store
-            .alloc_iter(deps.iter().map(|dep| {
-                Incompatibility::from_dependency(package.clone(), version.clone(), dep)
-            }));
+        let new_incompats_id_range =
+            self.incompatibility_store
+                .alloc_iter(deps.iter().map(|(dep, requirement)| match requirement {
```
mpizenberg, line 84: I'd suggest rename `dep` into `p` or `pkg` here to fit with the type annotation

src/solver.rs
mpizenberg: Or we could decide to break compatibility of serialized data? or start defining a serialized format more in line with the internal representation of incompatiblities (list of pairs of pkg, and +/- version set) which is not going to change soon as this would not be a "pubgrub" algorithm anymore? Then will have manually written serialize and deserializers.

src/solver.rs
mpizenberg: Indeed, I'm not a big fan of making it easy for people to make mistakes. I'd say we probably make `add_dependency` and `add_constraint` private or remove them alltogether, and only keep the `add_requirements` function. And the new `add_requirements` function should use `=` instead of `insert`. It doesn't make sense to have both a required dep and a constraint for the same dependency package.

mpizenberg (commented):
I haven't looked at the testing part but added some comments on the rest. I'm very happy that adding constraints managed to unblock you @baszalmstra! I think the changes make sense for your needs.

My personal position is that we should make dependency definition and retrieval our focus for pubgrub v0.4 (after 0.3 which focuses on version sets). I'd like that we explore many different ways of changing the `get_dependencies` part of pubgrub API, all the way from what's currently possible to having full flexibility of providing incompatibilities directly. Evaluate that whole spectrum of solutions, including intermediate ones like this one with "constraints".

My reasoning is that

- yesterday we needed to transform `(p1, v1) depends on (p2, range2)` into `{ p1: +singleton(v1), p2: -range2 }`
- today we also need to transform `(p1, v1) is constrained (optionally) by (p2, range2)` into `{ p1: +singleton(v1), p2: +(complement(range2) }`
- tomorrow we might need to transform `(p1, v1) might be used with either (p2, range2) or (p3, range3)` into ...
- ...

The list goes on, with all things considered valid incompatibilities by pubgrub. By the way, for example, your use case of saying "my package p1 is incompatible with p2" would be better served by a custom type transforming `(p1, v1) incompat with (p2, *)` into `{ p1: +singleton(v1), p2: +*}` instead of the twisting exercise consisting in saying `(p1, v1) is constrained by (p2, v<0)` which is effectively the same, since `complement(v<0)` is `*`.

All that is to say, I think this PR and your use case is a very important step in an API and ergonomics work which could be the focus of pubgrub v0.4. But I don't see this as being merged before we do similar experiments on other designs.

konstin:
I think this PR is stalled? At least it's heavy conflicting with current main.

baszalmstra:
Jep this was a very early experiment. Ill close. Thanks for the ping.

Issue #110, opened 2021-12-30
Async API?


mpizenberg:
I've been playing with making a [WebAssembly solver for the Elm ecosystem](https://github.com/mpizenberg/elm-solve-deps-wasm), with the aim of it being easily usable for tooling written in JavaScript. The wasm API I've settled on for the time being looks like this:
```js
let wasm = require("elm-solve-deps-wasm");
let solution = wasm.solve_deps(
  rootPackage,
  fetchDependencies, // function
  listVersions // function
);
```
It works fine but requires that the `fetchDependencies` and `listVersions` functions passed as argument are sync. However, in the NodeJS ecosystem, reading files and making http requests is mainly done with async. So this API forces the user to use sync versions of file IO, which is fine, and sync versions of http requests which is a bit more annoying to do.

In order to be able to write those two functions with async, it means we also need an async version of the `DependencyProvider` trait, where `getDependencies` and `choosePackageVersion` would return futures instead of normal values. I'm not sure if it's easily done, because async and traits are quite painful together from some of my readings, but I wanted to raise this use case.

Eh2406:
Thank you for giving this a try! I look forward to reading over your elm example! It is related to https://github.com/pubgrub-rs/advanced_dependency_providers/issues/6.

The first problem is that we don't fit well into the "just create futures and process them when they're done" paradigm. While `listVersions` is off doing network stuff, it's entirely possible that `unit_propagation` has discovered the parent cannot be used due to an unrelated conflict.
We also don't fit well in the underlying "cooperative multi-threading" paradigm. `unit_propagation` is synchronous computational work that is sometimes fast enough to do on the main thread, and other times without warning very blocking. That makes it very uncooperative with respect to "cooperative multi-threading".
However it is a real user need, so whatever "the best we can do" is we need to implement or document it.

What I had in mind was:
- an async `DependencyProvider` has a cache of `package -> promise<listVersions result>`.
- `choosePackageVersion` looks over the packages and asynchronously queues any items not in the cash. (Which will get processed by some other thread.)
- `choosePackageVersion` it then filters too only items whose result is available and pics one.
- if none are available then it blocks for one of them to be available.

Of course this may fall apart when we actually try and implement it.

Eh2406:
CC #138

konstin:
[...]

PR #126, opened 2023-05-04
Adds optional async interface

xfbs:
Hey!

In reference to #110, this adds an optional `async` feature which, when enabled, enables the `AsyncDependencyProvider` trait.

This PR also cleans up the `resolver` code somewhat (I like it when code is not nested too deeply, as it makes it easier for me to follow).

### DIFF
--- a/src/solver.rs
+++ b/src/solver.rs
@@ -74,8 +74,12 @@ use crate::internal::core::State;
 use crate::internal::incompatibility::Incompatibility;
 use crate::package::Package;
 use crate::range::Range;
+use crate::term::Term;
 use crate::type_aliases::{Map, SelectedDependencies};
 use crate::version::Version;
+
+#[cfg(feature = "async")]
+pub use async_trait::async_trait;
 
 /// Main function of the library.
 /// Finds a set of packages satisfying dependency bounds for a given package + version pair.
@@ -94,20 +98,19 @@ pub fn resolve<P: Package, V: Version>(
 
         state.unit_propagation(next)?;
 
-        let potential_packages = state.partial_solution.potential_packages();
-        if potential_packages.is_none() {
-            drop(potential_packages);
-            // The borrow checker did not like using a match on potential_packages.
-            // This `if ... is_none ... drop` is a workaround.
-            // I believe this is a case where Polonius could help, when and if it lands in rustc.
-            return state.partial_solution.extract_solution().ok_or_else(|| {
-                PubGrubError::Failure(
-                    "How did we end up with no package to choose but no solution?".into(),
-                )
-            });
-        }
+        let potential_packages = match state.partial_solution.potential_packages() {
+            Some(packages) => packages,
+            None => {
+                return state.partial_solution.extract_solution().ok_or_else(|| {
+                    PubGrubError::Failure(
+                        "How did we end up with no package to choose but no solution?".into(),
+                    )
+                });
+            }
+        };
+
         let decision = dependency_provider
-            .choose_package_version(potential_packages.unwrap())
+            .choose_package_version(potential_packages)
             .map_err(PubGrubError::ErrorChoosingPackageVersion)?;
         next = decision.0.clone();
 
@@ -116,6 +119,7 @@ pub fn resolve<P: Package, V: Version>(
             .partial_solution
             .term_intersection_for_package(&next)
             .expect("a package was chosen but we don't have a term.");
+
         let v = match decision.1 {
             None => {
                 let inc = Incompatibility::no_versions(next.clone(), term_intersection.clone());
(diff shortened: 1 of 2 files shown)


mpizenberg:
Cool! I didn’t know async trait was a thing yet. Would you mind splitting this PR in 2? So we can easily review first the code cleanup, then the async one?

xfbs:
For sure! Sounds like a good idea. I'll pull the refactoring out into a separate PR. Will also add some tests for the async stuff in here.

mpizenberg:
Also, this would need to PR against the `dev` branch, not the `release` one.

xfbs:
Ha! I did not even know there was a dev branch (used to PRing against `master` with a `stable` branch for the release, hehe).

src/solver.rs
```diff
@@ -124,79 +128,215 @@ pub fn resolve<P: Package, V: Version>(
...
+            dep_incompats,
+            &state.incompatibility_store,
+        );
+    }
+}
+
+/// Main function of the library.
+/// Finds a set of packages satisfying dependency bounds for a given package + version pair.
+#[cfg(feature = "async")]
+pub async fn resolve_async<P: Package, V: Version>(
```
Eh2406, line 215: It seems unfortunate for us to need to completely duplicate all of this code.

src/solver.rs
```diff
@@ -124,79 +128,215 @@ pub fn resolve<P: Package, V: Version>(
...
+) -> Result<SelectedDependencies<P, V>, PubGrubError<P, V>> {
+    let mut state = State::init(package.clone(), version.into());
+    let mut added_dependencies: Map<P, Set<V>> = Map::default();
+    let mut next = package;
+    loop {
+        dependency_provider
+            .should_cancel()
+            .map_err(|err| PubGrubError::ErrorInShouldCancel(err))?;
+
+        state.unit_propagation(next)?;
```
Eh2406, line 228: This function call is usually cheap, but sometimes it will need to do a lot of work (potentially exponential amount of work) and will therefore block the asynchronous runner. On the other hand, if we do some kind of `spawn_blocking` call here the overhead will be significant as the work is almost always trivial.

I think this is a profound and serious issue we are going to need to explore before we can merge something like this.

src/solver.rs
```diff
@@ -124,79 +128,215 @@ pub fn resolve<P: Package, V: Version>(
...
+                return state.partial_solution.extract_solution().ok_or_else(|| {
+                    PubGrubError::Failure(
+                        "How did we end up with no package to choose but no solution?".into(),
+                    )
+                });
+            }
+        };
+
+        let decision = dependency_provider
+            .choose_package_version(potential_packages)
```
Eh2406, line 242: It would be really nice if the asynchronous calls to get package information can be run in parallel allowing this await to return when the first of them returns. Of course this would make reproducibility much more complicated, and would need to be documented carefully. But it should make many real-world cases much more efficient.

Eh2406 (commented):
A couple of interesting things to consider before the next version of this.

src/solver.rs
Eh2406: I was very lucky to get @carllerche's opinion on this. The two alternatives he suggested were:

1. Have `unit_propagation` to a fixed amount of work and if there's more work to be done do the rest in a `spon_blocking`. In theory, this is the best solution. In practice I'm not sure that we can clearly identify how much work is being done by each sub call within `unit_propagation`, so I don't know if it's practical. We would also need extensive tests to verify that we don't accidentally have a loop occurring before we call `spon_blocking`.
2. Do not provide an async interface. Callers can `block_in_place` before calling resolve, and call `tokio::runtime::Handle::block_on` within the callbacks.

I'm not sure which solution is more unpleasant/inefficient, but it's comforting to know that it is a real problem.

src/solver.rs
xfbs: I definitely agree.

My rough idea is to refactor this method into smaller methods, such that there is no duplication, only the async parts need to be duplicated. 

I will wait for the refactoring PR (#127) to go through before starting this however.

src/solver.rs
xfbs: To be honest, I did not know of the `block_in_place` approach, but that sounds like something that might be easy to use and still provide the ability to interface with async code.

Maybe one idea is to add this to the documentation with some code examples? This should be a good starting point for other people that interface with async stuff.

src/solver.rs
Eh2406: Taking long methods and splitting them up seems like good value all on its own. I look forward to seeing your work on it.

src/solver.rs
Eh2406: My main concern with that approach is how it works with other executors. It can work well for Tokyo, but are there equivalents in the others? What about when running as Wasm in a browser?

Anyway, 110% on documentation. We should have a section in the book about how to shoehorn asynchronous communication into the current API. That, to the best of our ability, documents the available options and their advantages and disadvantages. That unblocks people without waiting for us to release a new version of the crate.

PR #127, opened 2023-05-04
Cleanup src/solver.rs

xfbs:
This PR does some little refactoring of the `resolve` method.

- I rearranged the code so that it reduces nesting. I do that by rearranging conditionals so that the error case returns early and then paves the way for the (non-indented) success path.
- I give booleans a name (I tried, hope the names make sense) so that if statements "make more sense"
- I turn an "expect" into an error and remove an `unwrap()` by using a match statement.

### DIFF
--- a/src/solver.rs
+++ b/src/solver.rs
@@ -78,6 +78,7 @@ use crate::internal::incompatibility::Incompatibility;
 use crate::package::Package;
 use crate::type_aliases::{DependencyConstraints, Map, SelectedDependencies};
 use crate::version_set::VersionSet;
+use log::{debug, info};
 
 /// Main function of the library.
 /// Finds a set of packages satisfying dependency bounds for a given package + version pair.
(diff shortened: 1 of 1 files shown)


src/solver.rs
```diff
@@ -94,36 +95,38 @@ pub fn resolve<P: Package, VS: VersionSet>(
...
-            // The borrow checker did not like using a match on potential_packages.
-            // This `if ... is_none ... drop` is a workaround.
-            // I believe this is a case where Polonius could help, when and if it lands in rustc.
-            return state.partial_solution.extract_solution().ok_or_else(|| {
-                PubGrubError::Failure(
-                    "How did we end up with no package to choose but no solution?".into(),
-                )
-            });
-        }
+        let potential_packages = match state.partial_solution.potential_packages() {
```
Eh2406: Looking at the code as you have it here, it seems like a perfect place for the new `let ... else` syntax.

Eh2406 (approved):
Thanks for the improvements!

src/solver.rs
xfbs: Ha! Never used that one before. Pretty neat. I don't like how it makes the initial line quite long, but I do like how it removes one level of indentation from the body.

Eh2406: Talking about syntax is so confusing. No there is a newer and even more concise syntax https://rust-lang.github.io/rfcs/3137-let-else.html

Something like
```rust
        let Some(packages) = state.partial_solution.potential_packages() else {
            return state.partial_solution.extract_solution().ok_or_else(|| {
                PubGrubError::Failure(
                    "How did we end up with no package to choose but no solution?".into(),
                )
            });
        };
```

mpizenberg (commented):
Thanks for the PR! At first read I like most of it though I have a few remarks. We tend to be as specific as possible and favor reader over writer, so I'm not a super fan of unqualified import (for the `log` module).
Another nitpick, I'd prefer avoiding shadowing of variable (`dependencies`).

src/solver.rs
xfbs: Sweet, I've just re-pushed it using that syntax :) That is definitely more concise.

xfbs:
> Thanks for the PR! 

Happy to help, awesome crate :) We use this in production btw!

> At first read I like most of it though I have a few remarks. We tend to be as specific as possible and favor reader over writer, so I'm not a super fan of unqualified import (for the log module).

I definitely agree with you! I tend to always explicitly include anything I use, with only two exceptions: `log::*` and `tracing::*` (never both, just either one of them ofc), just because these crates are so useful everywhere and it's nice to be able to just plop a `debug!()` somewhere. But this is a personal preference, so I changed to explicitly importing them (I hope that is what you prefer).

> Another nitpick, I'd prefer avoiding shadowing of variable (dependencies).

That is one feature of Rust I love :D but I get you, so I changed it, I hope my naming makes sense (I'm not too familiar with the codebase).

Feel free to also just take some part of my PR and throw out other parts! Style is always very subjective, I don't want to enforce my ideas, just sharing something that might be useful.

Cheers!

Eh2406:
@mpizenberg are we good to merge now?

mpizenberg:
If it's good for you, let me final check Sunday and I'll merge. If you
don't here from me on Monday, consider it's good to merge :)

mpizenberg:
Feels good to get rid of another `unwrap` and `expect` :)
Thanks again for the improvements.

PR #129, opened 2023-06-10
test: prop test for error report and refactor

Eh2406:
Removing a dependency from a dependency provider can only make more resolves succeed. Therefore, an interesting test is to look at our error reporting for some resolve, and we resolve with a dependency provider that only has dependency requirements mentioned in our report. If this subsequent run were to pass, then we have demonstrated that the error report did not include enough information to constitute a proof of unsatisfiability.

While working on the code I moved some code out of the proptest macro, as auto complete and formatting works better on non-macro code.

It would be nice to be able to test our proof of unsatisfiability against the failure information from the SAT solver. I was not able to figure out how to do this. I did find several small things that could be cleaned up about the implementation for SAT embedding.

### DIFF
--- a/src/solver.rs
+++ b/src/solver.rs
@@ -175,7 +175,7 @@ pub fn resolve<P: Package, VS: VersionSet>(
             Dependencies::Known(x) if x.contains_key(p) => {
                 return Err(PubGrubError::SelfDependency {
                     package: p.clone(),
-                    version: v.clone(),
+                    version: v,
                 });
             }
             Dependencies::Known(x) => x,
--- a/tests/proptest.rs
+++ b/tests/proptest.rs
@@ -5,11 +5,12 @@ use std::{collections::BTreeSet as Set, error::Error};
 use pubgrub::error::PubGrubError;
 use pubgrub::package::Package;
 use pubgrub::range::Range;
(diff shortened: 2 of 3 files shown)


Eh2406:
Rebased + happy clippy + fmt fix

Eh2406:
Pull out some redundant code for checking with sat

Eh2406:
This just affects test code, so hopefully we can keep it moving.

mpizenberg:
Nice!
I have an end-of-quarter thing at work to prepare for monday, but after that I’ll fully review here.
Also getting a bit hyped up by packaging-con 2023 being announced so I’ll boost my time for pubgrub :)

Eh2406:
ping?

Eh2406:
Added a profile override to make (sat) tests run faster. Which let me run `large_case` on more files I had laying around, some of which had errors that were not `NoSolution`.

Eh2406:
Backed out the override changes, even with the override the test was still too slow to run without `--release`, and with `--release` dependencies are already optimized.

Eh2406:
Rebased. And this just affects test code, so starting a 10 day timer.

Eh2406:
10 day timer is over with no objection. I will muck with settings so that I can self approve.

PR #133, opened 2023-10-04
Allow DependencyOnTheEmptySet

### LINKED ISSUE #125: DependencyOnTheEmptySet
For the Conda solver I build I implemented a `VersionSet` that builds a bitset of applicable version from a Conda specific range specifier. I can do this because I can know upfront exactly which versions are available of any given package. This makes implementing `complement`, `intersection`, etc very easy and fast. An `Empty` can also be very easily expressed if all bits are set to zero.

However, sometimes I encounter a package version (lets call it `A`) that has a dependency range that doesn't select any package version. This is fine and should indicate that package version `A` should also not be selected. However, because the range equals to empty I get the `DependencyOnTheEmptySet` error and the solve process aborts.

I was wondering if that behavior is an optimization or if that's something that is required by the algorithm. In my case, due to the way my `VersionSet` implementation behaves it doesn't make sense. Is it something that I could optionally turn off?

Eh2406:
This closes #125

I started by adding a panic to the sat solver when it is had a `DependencyOnTheEmptySet`, all tests passed because proptest  was not configured to generate them. I modified proptest and confirmed that this test now failed. `meta_test_deep_trees_from_strategy` then informed me that I had too many ways for package to be invalid. Removing `allow_deps` and reducing the probability of generating `DependencyOnTheEmptySet` made it happy. Then I could get to the meat of the experiment. I removed the code that generates `DependencyOnTheEmptySet` and all test pass. Well prop test passing once don't mean much, so I increased the `cases` in the `ProptestConfig`. It has been running for a long time without finding a problem.

Excitement over this progress is a little premature. This library is surprisingly robust against violations of its internal assumptions. In a number of places we filter out and assume that no term of an incompatibility is `Term::any()`, which turns out to be exactly what's generated when `DependencyOnTheEmptySet`. Adding a check for that property in `merge_incompatibility` makes everything unhappy. This is easy enough, when we generate the incompatibility for the dependency just leave off that term. Unfortunately, this causes another problem if the root package has one of these dependencies then it is automatically terminal, leading to the unhelpful error message "Root package depends on itself at a different version?". That check is always been suspect, it even has a comment about removing it. If we do remove it, everything passes. The next call is to `add_version` which does a more complicated check and decides not to add the package. So the loop continues to `unit_propagation` which correctly generates a proof of unSAT.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -32,21 +32,6 @@ pub enum PubGrubError<P: Package, VS: VersionSet> {
 
     /// Error arising when the implementer of
     /// [DependencyProvider](crate::solver::DependencyProvider)
-    /// returned a dependency on an empty range.
-    /// This technically means that the package can not be selected,
-    /// but is clearly some kind of mistake.
-    #[error("Package {dependent} required by {package} {version} depends on the empty set")]
-    DependencyOnTheEmptySet {
-        /// Package whose dependencies we want.
-        package: P,
-        /// Version of the package for which we want the dependencies.
-        version: VS::V,
-        /// The dependent package that requires us to pick from the empty set.
-        dependent: P,
-    },
-
-    /// Error arising when the implementer of
-    /// [DependencyProvider](crate::solver::DependencyProvider)
     /// returned a dependency on the requested package.
     /// This technically means that the package directly depends on itself,
     /// and is clearly some kind of mistake.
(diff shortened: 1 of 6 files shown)


Eh2406:
I think this is correct and a way to add an important use case. But there are lots of subtle interactions. @mpizenberg, do you see something I missed?

mpizenberg (approved):
Thanks for the call and explanations

Eh2406:
~In retrospect this should probably have non-prop tests as well. I can make a PR soonish.~ There where existing tests.

Issue #116, opened 2022-05-27
More compatible error type


Eh2406:
At a minimum we should switch to `dyn Error + Send + Sync` to be compatible with `anyhow`. Even better would be to have an arbitrary type associated with a dependency provider that we return in our errors. We don't actually care what the properties of a user's error are, we just need to be able to return them to the user.

https://rust-lang.zulipchat.com/#narrow/stream/260232-t-cargo.2FPubGrub/topic/Error.20type

charliermarsh:
@Eh2406 - Are you open to a PR changing to `dyn Error + Send + Sync`?

Eh2406:
I think a fully generic type would be better, but let's not let the perfect be the enemy of good. Yes I would accept such a PR.

PR #136, opened 2023-10-17
Extend error type to `Send + Sync`

charliermarsh:
## Summary

Though we may want to adopt a fully generic error type eventually, this at least makes the error type compatible with `anyhow` (and makes it async-friendly, which is both good and bad, since it requires that user errors are thread safe regardless of whether it's important for them to be).

See: https://github.com/pubgrub-rs/pubgrub/issues/116


Eh2406 (approved):
This looks like a good step forward!

I will unilaterally merge in no objections in the next 10 days.

mpizenberg:
I trust you on this one.

PR #137, opened 2023-10-25
More verbose debug assert

konstin:
Accidentally caused this error, added a better debug message

### DIFF
--- a/src/internal/partial_solution.rs
+++ b/src/internal/partial_solution.rs
@@ -141,7 +141,13 @@ impl<P: Package, VS: VersionSet> PartialSolution<P, VS> {
                     AssignmentsIntersection::Decision(_) => panic!("Already existing decision"),
                     // Cannot be called if the versions is not contained in the terms intersection.
                     AssignmentsIntersection::Derivations(term) => {
-                        debug_assert!(term.contains(&version))
+                        debug_assert!(
+                            term.contains(&version),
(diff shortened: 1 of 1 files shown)


Eh2406:
Thank you! This looks good. When CI is green I will be happy to merge.

Can you elaborate on how you hit this error? I'm surprised a higher level / user facing check didn't catch the bug.

Eh2406:
(The commit message needs to follow "conventional commit" before I can merge.)

mpizenberg:
> (The commit message needs to follow "conventional commit" before I can merge.)

Nah, it's fine you rename the commit anyway when merging so you can conventionalize it @Eh2406

Eh2406:
@mpizenberg how can I check that the new message will make CI happy. I have messed up before and ended merging with a bad name and had red CI on dev, witch required doing a force push to fix, witch required turning off the branch protection. Witch is a mess I don't want to repeat.

mpizenberg:
Ahah, ok makes sense ^^

konstin:
commitlint passes locally, though i don't know if it matches the CI one.

> Can you elaborate on how you hit this error? I'm surprised a higher level / user facing check didn't catch the bug.

iirc i was midway into refactoring that made more versions available (those with only a source distribution and lacking a wheel) in on place but not in another.

Eh2406 (approved):
Thank you!

PR #142, opened 2023-11-02
Move to edition 2021

### LINKED ISSUE #140: Move to edition 2021

0xAtticus:
Fix #140

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ authors = [
     "Alex Tokarev <aleksator@gmail.com>",
     "Jacob Finkelman <Eh2406@wayne.edu>",
 ]
-edition = "2018"
+edition = "2021"
 description = "PubGrub version solving algorithm"
 readme = "README.md"
 repository = "https://github.com/pubgrub-rs/pubgrub"
--- a/src/internal/small_vec.rs
+++ b/src/internal/small_vec.rs
@@ -166,9 +166,9 @@ impl<T> IntoIterator for SmallVec<T> {
     fn into_iter(self) -> Self::IntoIter {
         match self {
             SmallVec::Empty => SmallVecIntoIter::Empty,
-            SmallVec::One(a) => SmallVecIntoIter::One(IntoIterator::into_iter(a)),
-            SmallVec::Two(a) => SmallVecIntoIter::Two(IntoIterator::into_iter(a)),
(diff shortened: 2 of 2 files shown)


0xAtticus:
First of all, running:
`cargo fix --edition` does not trigger any change.

## Prelude changes
_No changes required._
The following traits were added to the prelude: 
- std::convert::TryInto
- std::convert::TryFrom
- std::iter::FromIterator
This may lead to some conflicts if we have custom traits defined with the same method names. However, we have no custom traits which have method named `try_into`, `try_from` or `from_iter`.  (this would most certainly not compile anyway)

## New default resolver.
_No changes required_
Simply an improvement of dependencies
```bash
When building the following dependencies, the given features will no longer be used:

  log v0.4.20 removed features: std
  syn v2.0.38 (as host dependency) removed features: full, visit

The following differences only apply when building with dev-dependencies:

  libc v0.2.149 removed features: extra_traits
  regex v1.10.2 removed features: default, unicode, unicode-age, unicode-bool, unicode-case, unicode-gencat, unicode-perl, unicode-script, unicode-segment
  regex-automata v0.4.3 removed features: unicode, unicode-age, unicode-bool, unicode-case, unicode-gencat, unicode-perl, unicode-script, unicode-segment, unicode-word-boundary
  regex-syntax v0.8.2 removed features: default, unicode, unicode-age, unicode-bool, unicode-case, unicode-gencat, unicode-perl, unicode-script, unicode-segment
  ```
## IntoIterator for arrays
This may have impact for `SmallVec`. No lint `array_into_iter` were triggered, because we were using the `IntoIterator::into_iter` syntax. According to the migration guide, this means that we _should already be ready to go for Rust 2021!_. 
An optional migration is to replace `IntoIterator::into_iter` with `.into_iter`, which has been done in the pull request.

## Disjoint capture in closures
_No changes required_
Could not find any impact manually, and no `rust_2021_incompatible_closure_captures` lint are triggered.

## Panic macro consistency
_No changes required_
No panic message is concerned by the change.

## Reserving syntax
_No changes required_
We were not using any of the now reserved syntax.

## Warnings promoted to errors
_No changes required_
There are no lint warning nor errors.

## Or patterns in macro-rules
_No changes required_
If I am not mistaken, there are no macro defined in the project.

Eh2406 (approved):
This looks good! Thanks for the PR and wright up!

mpizenberg:
Thanks for the analysis and write up @0xAtticus !

PR #146, opened 2023-11-05
Use Merge Queue to Fix CI For Conventional Commits

### LINKED ISSUE #139: Fix CI for conventional commits
This project insists on using conventional commits for everything on the `dev` and `release` branches. We attempt to check this in CI. The current CI code checks that every commit to a PR or on `dev` or `release` follow the style. We also generally use a "squash and merge" merging strategy.

This doesn't work for at least three reasons:
- PR's have red CI, if there commits don't follow the style, even though those commits will be squash away after merge.
- When doing a squash and merge, the maintainer gets to pick a new name. But nothing checks the new name follows conventional commits.
- After squash and merge, CI flags nonconventional names. But at that point the only thing that can be done is force pushing to `dev`, which is worse than the problem it's trying to solve.

I think this can be fixed by the following:
1. Turn on merge queues. This feature is intended to allow testing multiple PR's at the same time in the exact state they will be after merging. We don't need the additional parallelism, but we do need CI to run after the commit has been squashed.
2. Reconfigure CI to run on pushes to PR and being added to merge queue.
3. Set the conventional commits to only run in a merge queue.

It would be wonderful to have someone test the proposed configuration in some other repo to make sure it works, and then submit a PR to change CI here.

If we don't find a good solution, I would like to remove the conventional commits checking. Force pushing to `dev` is worse than having to build the first draft of the change log manually.

svanderbleek:
fixes #139

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: CI
 on:
   pull_request:
+  merge_group:
   push:
     branches: [ release, dev ]
   schedule: [ cron: "0 6 * * 4" ]
@@ -103,6 +104,7 @@ jobs:
   check_commit_conventions:
     name: Commit messages follow project guidelines
     runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request' || github.event.action == 'enqueued'
     steps:
       - uses: actions/checkout@v2
         with:


aleksator (approved):

aleksator:
Thanks @svanderbleek!
It worked wonders: https://github.com/pubgrub-rs/pubgrub/actions/runs/6765092186/job/18384272644

Issue #155, opened 2023-11-20
Why do error messages start with no version?


Eh2406:
I was looking at the error message created by the test case for #135. The case I was looking at was:
```rust
    let mut dependency_provider = OfflineDependencyProvider::<&str, NumVS>::new();

    // root depends on foo...
    dependency_provider.add_dependencies("root", 1, vec![("foo", Range::full())]);

    for i in 1..6 {
        // foo depends on bar...
        dependency_provider.add_dependencies("foo", i as u32, vec![("bar", Range::full())]);
    }
```
which gives the error message:
```
Because there is no available version for bar and foo 1 depends on bar, foo 1 is forbidden.
And because there is no version of foo in <1 | >1, <2 | >2, <3 | >3, <4 | >4, <5 | >5, foo <2 | >2, <3 | >3, <4 | >4, <5 | >5 is forbidden. (1)

Because there is no available version for bar and foo 2 depends on bar, foo 2 is forbidden.
And because foo <2 | >2, <3 | >3, <4 | >4, <5 | >5 is forbidden (1), foo <3 | >3, <4 | >4, <5 | >5 is forbidden. (2)

Because there is no available version for bar and foo 3 depends on bar, foo 3 is forbidden.
And because foo <3 | >3, <4 | >4, <5 | >5 is forbidden (2), foo <4 | >4, <5 | >5 is forbidden. (3)

Because there is no available version for bar and foo 4 depends on bar, foo 4 is forbidden.
And because foo <4 | >4, <5 | >5 is forbidden (3), foo <5 | >5 is forbidden. (4)

Because there is no available version for bar and foo 5 depends on bar, foo 5 is forbidden.
And because foo <5 | >5 is forbidden (4), foo * is forbidden.
And because root 1 depends on foo, root 1 is forbidden.
```
The logical argument made by the current error message is correct, it's just confusing to my meat brain.
The second point in the argument is that `there is no version of foo in <1 | >1, <2 | >2, <3 | >3, <4 | >4, <5 | >5`. Which is correct, but the error messages give me know context for why I would be interested in this range with all the holes in it. The actual answer is the each of those holes is going to be filled in by a argument further down, but I haven't read ahead to find that out yet. It seems like the argument would be much clearer if it started by the versions that it tried, and then explained that there were no versions left in the gaps. Something like:
```
Because there is no available version for bar and foo 5 depends on bar, foo 5 is forbidden. (1)

[...]

Issue #135, opened 2023-10-17
Solving slows down dramatically when testing hundreds of versions


charliermarsh:
When testing many consecutive versions of a given package, the cost of computing compatible versions seems to increase non-linearly over time.

Concretely, running this example slows down noticeably for me at around 100 iterations, and grinds to a ~halt in the 200s:

```rust
use pubgrub::error::PubGrubError;
use pubgrub::range::Range;
use pubgrub::report::{DefaultStringReporter, Reporter};
use pubgrub::solver::{OfflineDependencyProvider, resolve};
use pubgrub::version::SemanticVersion;

fn main() {
    let mut dependency_provider = OfflineDependencyProvider::<&str, SemanticVersion>::new();

    // root depends on foo...
    dependency_provider.add_dependencies(
        "root", (1, 0, 0),
        vec![
            ("foo", Range::any()),
        ],
    );

    for i in 1..1000 {
        dbg!(i);

        // foo depends on bar...
        dependency_provider.add_dependencies(
            "foo", (i, 0, 0),
            vec![
                ("bar", Range::any()),
            ],
        );

        match resolve(&dependency_provider, "root", (1, 0, 0)) {
            Ok(sol) => println!("{:?}", sol),
            Err(PubGrubError::NoSolution(mut derivation_tree)) => {
                derivation_tree.collapse_no_versions();
                eprintln!("{}", DefaultStringReporter::report(&derivation_tree));
            }
            Err(err) => panic!("{:?}", err),
        };
    }
}
```

I suspect that this may have to do with the increased size of the intersected versions. For example, after testing and rejecting Django 5.0a1, we produce a term like:

```
[ 0a0.dev0, 5.0a1 [  [ 5.0a1.post0.dev0, ∞ [
```

We then test Django 4.2.6, which gives us:

```
[ 0a0.dev0, 4.2.6 [  [ 4.2.6.post0.dev0, ∞ [
```

We then take the intersection of these terms, which gives us:

```
[ 0a0.dev0, 4.2.6 [  [ 4.2.6.post0.dev0, 5.0a1 [  [ 5.0a1.post0.dev0, ∞ [
```

This continues until we have a range like:

```
[...]

PR #156, opened 2023-11-22
feat: add a `simplify` for error messages

Eh2406:
Ranges generated by PubGrub often end up being verbose and pedantic about versions that do not matter. To take an example from #155 `2 | 3 | 4 | 5` could be more concisely stated as `>=2, <=5` given that it's not possible to have versions in between the integers. More generally it could be expressed as `>=2`, if the list of available versions were taken into account.

The logic for simplifying a VS given a complete set of versions feels like it should be simple. But it is trickier to implement than I expected. Especially if you want to guarantee `O(len(VS) + len(versions))` time and `O(1)` allocations.

While working through the logic I had to implement a check which versions match from a list in `O(len(VS) + len(versions))` time, which felt useful enough to be worth including in the API. In implementing that I noticed that our implementation of `contains`, did not aggressively short-circuit. The short-circuiting implementation is just as easy to read, so I decided to include that as well.

### DIFF
--- a/src/range.rs
+++ b/src/range.rs
@@ -51,6 +51,7 @@
 //! If we do not see practical bugs, or we get a formal proof that the code cannot lead to error states, then we may remove this warning.
 
 use crate::{internal::small_vec::SmallVec, version_set::VersionSet};
+use std::cmp::Ordering;
 use std::ops::RangeBounds;
 use std::{
     fmt::{Debug, Display, Formatter},
@@ -202,21 +203,44 @@ impl<V: Ord> Range<V> {
     /// Returns true if the this Range contains the specified value.
     pub fn contains(&self, v: &V) -> bool {
         for segment in self.segments.iter() {
-            if match segment {
-                (Unbounded, Unbounded) => true,
-                (Unbounded, Included(end)) => v <= end,
-                (Unbounded, Excluded(end)) => v < end,
-                (Included(start), Unbounded) => v >= start,
-                (Included(start), Included(end)) => v >= start && v <= end,
-                (Included(start), Excluded(end)) => v >= start && v < end,
-                (Excluded(start), Unbounded) => v > start,
-                (Excluded(start), Included(end)) => v > start && v <= end,
-                (Excluded(start), Excluded(end)) => v > start && v < end,
-            } {
-                return true;
+            match within_bounds(v, segment) {
+                Ordering::Less => return false,
+                Ordering::Equal => return true,
+                Ordering::Greater => (),
             }
         }
         false
+    }
+
+    /// Returns true if the this Range contains the specified values.
+    ///
+    /// The `versions` iterator must be sorted.
+    /// Functionally equivalent to `versions.map(|v| self.contains(v))`.
+    /// Except it runs in `O(size_of_range + len_of_versions)` not `O(size_of_range * len_of_versions)`
+    pub fn contains_many<'s, I>(&'s self, versions: I) -> impl Iterator<Item = bool> + 's
+    where
+        I: Iterator<Item = &'s V> + 's,
+        V: 's,
+    {
+        self.locate_versions(versions).map(|m| m.is_some())
+    }
+
+    /// Return the segment index in the range for each version in the range, None otherwise
+    fn locate_versions<'s, I>(&'s self, versions: I) -> impl Iterator<Item = Option<usize>> + 's
+    where
+        I: Iterator<Item = &'s V> + 's,
+        V: 's,
+    {
+        versions.scan(0, |i, v| {
+            while let Some(segment) = self.segments.get(*i) {
+                match within_bounds(v, segment) {
+                    Ordering::Less => return Some(None),
+                    Ordering::Equal => return Some(Some(*i)),
+                    Ordering::Greater => *i += 1,
+                }
+            }
+            Some(None)
+        })
     }
 
     /// Construct a simple range from anything that impls [RangeBounds] like `v1..v2`.
@@ -264,6 +288,26 @@ impl<V: Ord> Range<V> {
     }
 }
 
+fn within_bounds<V: PartialOrd>(v: &V, segment: &Interval<V>) -> Ordering {
+    let below_lower_bound = match segment {
+        (Excluded(start), _) => v <= start,
+        (Included(start), _) => v < start,
+        (Unbounded, _) => false,
+    };
+    if below_lower_bound {
+        return Ordering::Less;
+    }
+    let below_upper_bound = match segment {
+        (_, Unbounded) => true,
+        (_, Included(end)) => v <= end,
+        (_, Excluded(end)) => v < end,
+    };
+    if below_upper_bound {
+        return Ordering::Equal;
+    }
+    Ordering::Greater
+}
+
 fn valid_segment<T: PartialOrd>(start: &Bound<T>, end: &Bound<T>) -> bool {
     match (start, end) {
         (Included(s), Included(e)) => s <= e,
@@ -272,6 +316,33 @@ fn valid_segment<T: PartialOrd>(start: &Bound<T>, end: &Bound<T>) -> bool {
         (Excluded(s), Excluded(e)) => s < e,
         (Unbounded, _) | (_, Unbounded) => true,
     }
+}
+
+/// group adjacent versions locations
+/// [None, 3, 6, 7, None] -> [(3, 7)]
+/// [3, 6, 7, None] -> [(None, 7)]
+/// [3, 6, 7] -> [(None, None)]
+/// [None, 1, 4, 7, None, None, None, 8, None, 9] -> [(1, 7), (8, 8), (9, None)]
+fn group_adjacent_locations(
+    mut locations: impl Iterator<Item = Option<usize>>,
+) -> impl Iterator<Item = (Option<usize>, Option<usize>)> {
+    // If the first version matched, then the lower bound of that segment is not needed
+    let mut seg = locations.next().flatten().map(|ver| (None, Some(ver)));
+    std::iter::from_fn(move || {
+        for ver in locations.by_ref() {
+            if let Some(ver) = ver {
+                // As long as were still matching versions, we keep merging into the currently matching segment
+                seg = Some((seg.map_or(Some(ver), |(s, _)| s), Some(ver)));
+            } else {
+                // If we have found a version that doesn't match, then right the merge segment and prepare for a new one.
+                if seg.is_some() {
+                    return seg.take();
+                }
+            }
+        }
+        // If the last version matched, then write out the merged segment but the upper bound is not needed.
+        seg.take().map(|(s, _)| (s, None))
+    })
 }
 
 impl<V: Ord + Clone> Range<V> {
@@ -319,6 +390,41 @@ impl<V: Ord + Clone> Range<V> {
             }
         }
 
+        Self { segments }.check_invariants()
+    }
+
+    /// Returns a simpler Range that contains the same versions
+    ///
+    /// For every one of the Versions provided in versions the existing range and
+    /// the simplified range will agree on whether it is contained.
+    /// The simplified version may include or exclude versions that are not in versions as the implementation wishes.
+    /// For example:
+    ///  - If all the versions are contained in the original than the range will be simplified to `full`.
+    ///  - If none of the versions are contained in the original than the range will be simplified to `empty`.
+    ///
+    /// If versions are not sorted the correctness of this function is not guaranteed.
+    pub fn simplify<'s, I>(&'s self, versions: I) -> Self
+    where
+        I: Iterator<Item = &'s V> + 's,
+        V: 's,
+    {
+        let version_locations = self.locate_versions(versions);
+        let kept_segments = group_adjacent_locations(version_locations);
+        self.keep_segments(kept_segments)
+    }
+
+    /// simplify range with segments at given location bounds.
+    fn keep_segments(
+        &self,
+        kept_segments: impl Iterator<Item = (Option<usize>, Option<usize>)>,
+    ) -> Range<V> {
+        let mut segments = SmallVec::Empty;
+        for (s, e) in kept_segments {
+            segments.push((
+                s.map_or(Unbounded, |s| self.segments[s].0.clone()),
+                e.map_or(Unbounded, |e| self.segments[e].1.clone()),
+            ));
+        }
         Self { segments }.check_invariants()
     }
 }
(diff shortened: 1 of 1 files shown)


Eh2406:
cc @zanieb, I think this can be used to dramatically improve your error messages.

cc @baszalmstra https://github.com/mamba-org/resolvo/issues/2#issuecomment-1744436726 I don't know if these are improvements you'd be interested in include in your copy of this type.

A future possibility is figuring out when it is safe for PubGrub to ask for a range to be simplified during processing. This would be enormously helpful for #135. But needs to be done with extreme care. Basic set properties do not hold if `simplify` is added to the equation: `A.itersection(A.negate()) == empty` is true, but `simplify(A).itersection(simplify(A.negate())) == empty` is not.

mpizenberg:
It feels like a very good idea for error reporting where solver logic is not important anymore, and readability is much more useful. Also at this point the potential costs are likely negligible compared to the solving costs?

src/range.rs
```diff
@@ -203,22 +203,55 @@ impl<V: Ord> Range<V> {
...
-                (Unbounded, Excluded(end)) => v < end,
-                (Included(start), Unbounded) => v >= start,
-                (Included(start), Included(end)) => v >= start && v <= end,
-                (Included(start), Excluded(end)) => v >= start && v < end,
-                (Excluded(start), Unbounded) => v > start,
-                (Excluded(start), Included(end)) => v > start && v <= end,
-                (Excluded(start), Excluded(end)) => v > start && v < end,
+                (Excluded(start), _) => v <= start,
+                (Included(start), _) => v < start,
+                (Unbounded, _) => false,
```
mpizenberg: Isn't this wrong if the very first segment is `(Unbounded, Unbounded)`?

mpizenberg: This whole short-circuit change feels shady ...

Eh2406: The `false` here means that it fails to short-circuit and falls through to the check below which correctly returns `true`. That being said I need to reread it twice to figure out what was going on here, and I wrote the code yesterday. There has to be a clear way to write this.

Similarly, we can probably improve the testing to ensure that the code are correct, even when were not paying attention. I will need to look into what is worth the effort. (One day, I would love to use kani or creusot to prove the code correct, not that I have time for it now.)

Eh2406:
I tinkered with the code to make it more "obviously correct". While I was at it I noticed that STD provide some helpful methods which simplified a few things.

I looked at our test code generation, and I'm pretty confident it will cover all the corner cases.

src/range.rs
```diff
@@ -202,23 +202,40 @@ impl<V: Ord> Range<V> {
...
-                (Unbounded, Included(end)) => v <= end,
-                (Unbounded, Excluded(end)) => v < end,
-                (Included(start), Unbounded) => v >= start,
-                (Included(start), Included(end)) => v >= start && v <= end,
-                (Included(start), Excluded(end)) => v >= start && v < end,
-                (Excluded(start), Unbounded) => v > start,
-                (Excluded(start), Included(end)) => v > start && v <= end,
-                (Excluded(start), Excluded(end)) => v > start && v < end,
-            } {
+            if !within_lower_bound(segment, v) {
```
mpizenberg: Ah ok, I I'm getting it now. The naming sounds a bit weird? "within bounds" makes sense, but "within lower bound" feels odd. I'd say something like "above lower bound", "below upper bound". Or if we want to avoid the negation, use "below" in both cases, like this:
```rust
if version_below_lower_bound( v, segment ) {
  return false;
} else if version_below_upper_bound( v, segment ) {
  return true;
}
```

PS, there is a typo in `within_uppern_bound` with an "n" at the end of "upper".

mpizenberg: Or alternatively, a method on the bounded range maybe:
```rust
if segment.is_above(v) {
  return false;
} else if !segment.is_below(v) {
  return true;
}
```
I'm not a super fan of the not `!` in there though.

**EDIT**: forget that, we don't have a type for the segment

Eh2406: How about a `within_bounds` that returns a https://doc.rust-lang.org/std/cmp/enum.Ordering.html ?

mpizenberg: So taking `Ordering` as a slightly richer `Bool`. It's tempting. In that case it's important to have the version argument before the segment argument (`within_bounds(v, segment)`) because ordering can be confusing.

mpizenberg:
What do you think of splitting a bit the simplify function, to make it easier to follow? I'm thinking something like this (pseudo code):

```rust
// return the segment index in the range for each version in the range, None otherwise
locate_versions(&self, versions: Iter<V>) -> Iter<Option<usize>>

// group adjacent versions locations
// [None, 3, 6, 7, None] -> [(3, 7)]
// [3, 6, 7, None] -> [(None, 7)]
// [3, 6, 7] -> [(None, None)]
// [None, 1, 4, 7, None, None, None, 8, None, 9] -> [(1, 7), (8, 8), (9, None)]
group_adjacent_locations(locations: Iter<Option<usize>>) -> Iter<(Option<usize>, Option<usize>)>

// simplify range with segments at given location bounds.
keep_segments(&self, kept_segments: Iter<(Option<usize>, Option<usize>)) -> Self

simplify(&self, versions: Iter<V>) -> Self {
  let version_locations = self.locate_versions(versions);
  let kept_segments = group_adjacent_locations(version_locations);
  self.keep_segments(kept_segments)
}
```

Eh2406:
`group_adjacent_locations` was harder than I was expecting. It's hard to build out of normal iterator methods, because it may need to return one more result after the underlying iterator has been exhausted. It would be straightforward to implement using generators, if they were stable. But eventually I made it work.

zanieb:
I'll try to use this today!

mpizenberg:
> `group_adjacent_locations` was harder than I was expecting.

If you feel it's harder than worth it maybe it's better to not split group_adjacent_locations and keep_segments. I'm not finding more straightforward ways to write it either. Your call.

I don't see your previous version to compare complexity. I guess you forced push rewrote it. I haven't followed the CI merging changes. Are PRs still squashed merged? or are they "normal" merge? Because if they are still squashed merge there is no need to overwrite your commit while working on the PR. And if they aren't well it's slightly annoying to make sure all a PR history is clean. I don't want to derail this PR, I just thought it was relevant as it makes my reviewing of the PR harder.

Eh2406:
I think they're worth keeping. I wish they had been easier to come up with or more direct code, but they do make things clear.

I did force push, and will try and avoid doing so while you're reviewing. I got into the habit because it simplifies the history which is helpful when dealing with several long-lived branches. Hopefully those will become less common. If you scroll through conversation here on GitHub it provides links for every time I force push, like 
image
the compare button can show you what changed between each force push.

src/range.rs
```diff
@@ -274,6 +318,33 @@ fn valid_segment<T: PartialOrd>(start: &Bound<T>, end: &Bound<T>) -> bool {
     }
 }
 
+/// group adjacent versions locations
+/// [None, 3, 6, 7, None] -> [(3, 7)]
+/// [3, 6, 7, None] -> [(None, 7)]
+/// [3, 6, 7] -> [(None, None)]
+/// [None, 1, 4, 7, None, None, None, 8, None, 9] -> [(1, 7), (8, 8), (9, None)]
```
konstin, line 318:
```suggestion
/// ```text
/// [None, 3, 6, 7, None] -> [(3, 7)]
/// [3, 6, 7, None] -> [(None, 7)]
/// [3, 6, 7] -> [(None, None)]
/// [None, 1, 4, 7, None, None, None, 8, None, 9] -> [(1, 7), (8, 8), (9, None)]
/// ```
```
(otherwise they are (broken) markdown links)

konstin:
> But needs to be done with extreme care. Basic set properties do not hold if `simplify` is added to the equation: `A.itersection(A.negate()) == empty` is true, but `simplify(A).itersection(simplify(A.negate())) == empty` is not.

Could you give an example where this doesn't hold? I tried but couldn't find one.

konstin:
I made a POC for simplifying all terms: https://github.com/zanieb/pubgrub/compare/main...zanieb:pubgrub:know-thy-versions-rebase. The main problem is that this breaks `accum_term.subset_of(&incompat_term)`, any ideas? I've just inserted the simplification at all the places where we regularly intersect, though i feels like we shouldn't need to build up all the intersections every time in the first place.

zanieb:
Hm so I gave this a try by hacking it into the reporter and testing a simple holes case.

https://github.com/pubgrub-rs/pubgrub/commit/b9b0e1d2a13248e76b26241da4bfb6b6fa637aaa

Before:
```
Because there is no available version for bar and foo 1.0.0 depends on bar, foo 1.0.0 is forbidden.
And because there is no version of foo in <1.0.0 | >1.0.0, <2.0.0 | >2.0.0, <3.0.0 | >3.0.0, <4.0.0 | >4.0.0, foo <2.0.0 | >2.0.0, <3.0.0 | >3.0.0, <4.0.0 | >4.0.0 is forbidden. (1)

Because there is no available version for bar and foo 2.0.0 depends on bar, foo 2.0.0 is forbidden.
And because foo <2.0.0 | >2.0.0, <3.0.0 | >3.0.0, <4.0.0 | >4.0.0 is forbidden (1), foo <3.0.0 | >3.0.0, <4.0.0 | >4.0.0 is forbidden. (2)

Because there is no available version for bar and foo 3.0.0 depends on bar, foo 3.0.0 is forbidden.
And because foo <3.0.0 | >3.0.0, <4.0.0 | >4.0.0 is forbidden (2), foo <4.0.0 | >4.0.0 is forbidden. (3)

Because there is no available version for bar and foo 4.0.0 depends on bar, foo 4.0.0 is forbidden.
And because foo <4.0.0 | >4.0.0 is forbidden (3), foo * is forbidden.
And because root 1.0.0 depends on foo, root 1.0.0 is forbidden.
```

After:
```

Because there is no version of bar in ∅ and foo <=1.0.0 depends on bar ∅, foo 1.0.0 is forbidden.
And because there is no version of foo in ∅, foo <2.0.0 | >2.0.0, <3.0.0 | >3.0.0, <4.0.0 | >4.0.0 is forbidden. (1)

Because there is no version of bar in ∅ and foo 2.0.0 depends on bar ∅, foo 2.0.0 is forbidden.
And because foo <2.0.0 | >2.0.0, <3.0.0 | >3.0.0, <4.0.0 | >4.0.0 is forbidden (1), foo <3.0.0 | >3.0.0, <4.0.0 | >4.0.0 is forbidden. (2)

Because there is no version of bar in ∅ and foo 3.0.0 depends on bar ∅, foo 3.0.0 is forbidden.
And because foo <3.0.0 | >3.0.0, <4.0.0 | >4.0.0 is forbidden (2), foo <4.0.0 | >4.0.0 is forbidden. (3)

Because there is no version of bar in ∅ and foo >=4.0.0 depends on bar ∅, foo 4.0.0 is forbidden.
And because foo <4.0.0 | >4.0.0 is forbidden (3), foo * is forbidden.
And because root depends on foo, root 1.0.0 is forbidden.
```

While we can certainly do something to improve the null cases, I'm not seeing this as a drastic improvement. Perhaps I'm doing something wrong?

Eh2406:
> > But needs to be done with extreme care. Basic set properties do not hold if `simplify` is added to the equation: `A.itersection(A.negate()) == empty` is true, but `simplify(A).itersection(simplify(A.negate())) == empty` is not.
> 
> Could you give an example where this doesn't hold? I tried but couldn't find one.

You seem to be correct, this implementation of simplify does uphold this property. The point I was trying to make is defining what properties `simplify` must uphold and making sure that the rest of the algorithm doesn't rely on anything else requires some care.

> I made a POC for simplifying all terms: https://github.com/zanieb/pubgrub/compare/main...zanieb:pubgrub:know-thy-versions-rebase. The main problem is that this breaks `accum_term.subset_of(&incompat_term)`, any ideas? I've just inserted the simplification at all the places where we regularly intersect, 

There are clearly some properties that code is relying on that `simplify` does not uphold. What they are, I do not yet know. And I would like to separate/delay the conversation about how the algorithm can use `simplify` until after we merge making the freestanding method useful/available.

> though i feels like we shouldn't need to build up all the intersections every time in the first place.

You may want to look at the "accumulated_intersection" and the "fewer_intersections" branches. I be happy to discuss other changes to the algorithm to reduce intersections, either on zulip or in a issue.

> Hm so I gave this a try by hacking it into the reporter and testing a simple holes case.

This is exactly where I was hoping this freestanding method would be useful. Let's see how much it helped...

> While we can certainly do something to improve the null cases, I'm not seeing this as a drastic improvement. Perhaps I'm doing something wrong?

That is not as helpful as I was hoping :-( clearly `simplify` is not working as well as I'd like. Let me experiment with your branch and I will report back.

Eh2406:
Got it. Most of the error reporting is based on the `terms`, replacing `terms: derived.terms.clone(),` with 
```
terms: derived
                    .terms
                    .iter()
                    .map(|(p, t)| (p.clone(), t.simplify(versions.get(&p).unwrap_or(&Vec::new()).into_iter())))
                    .collect(),
```
and adding a `simplify` method to term that forwards along. I get:

```
Because there is no version of bar in ∅ and foo <=1.0.0 depends on bar ∅, foo <=1.0.0 is forbidden.
And because there is no version of foo in ∅, foo <2.0.0 is forbidden. (1)

Because there is no version of bar in ∅ and foo 2.0.0 depends on bar ∅, foo 2.0.0 is forbidden.
And because foo <2.0.0 is forbidden (1), foo <3.0.0 is forbidden. (2)

Because there is no version of bar in ∅ and foo 3.0.0 depends on bar ∅, foo 3.0.0 is forbidden.
And because foo <3.0.0 is forbidden (2), foo <4.0.0 is forbidden. (3)

Because there is no version of bar in ∅ and foo >=4.0.0 depends on bar ∅, foo >=4.0.0 is forbidden.
And because foo <4.0.0 is forbidden (3), foo * is forbidden.
And because root depends on foo, root * is forbidden.
```

zanieb:
@Eh2406 ah that makes a lot of sense! I'll explore the effect on more error messages then.

zanieb:
With https://github.com/pubgrub-rs/pubgrub/commit/53c9f6d089a6445bc491751766b891e566f4af11 we get better handling of the empty ranges

```
Because there is no available version for bar and foo <=1.0.0 depends on bar, foo <=1.0.0 is forbidden.
And because there is no available version for foo, foo <2.0.0 is forbidden. (1)

Because there is no available version for bar and foo 2.0.0 depends on bar, foo 2.0.0 is forbidden.
And because foo <2.0.0 is forbidden (1), foo <3.0.0 is forbidden. (2)

Because there is no available version for bar and foo 3.0.0 depends on bar, foo 3.0.0 is forbidden.
And because foo <3.0.0 is forbidden (2), foo <4.0.0 is forbidden. (3)

Because there is no available version for bar and foo >=4.0.0 depends on bar, foo >=4.0.0 is forbidden.
And because foo <4.0.0 is forbidden (3), foo * is forbidden.
And because root depends on foo, root * is forbidden.
```

There's a bit of a problem because `And because there is no available version for foo` should read `And because there is no available version for foo >1.00,<2.0.0` and `And because foo <2.0.0 is forbidden (1)` should read `And because foo <2.0.0 is forbidden (1) and there is no available version for foo >2.0.0,<3.0.0`

Eh2406:
Right. This simplification code assumes that information about versions that don't exist is unneeded. Which is not true when dealing with "NoVertions", or anything derived from them. Because of #155, everything in this example derives from a "NoVertions". I'm open to ideas on how to get us to a better place.

In the meantime, the open question in this PR is whether this code is useful and worth merging.

zanieb:
I think this is a pretty clear path to better error messages. We can either continue tackling it piecewise by merging or devote a new branch to error messaging and merge the whole thing to `dev` at once.

Eh2406:
This project has been plagued with long living branches, so I'm biased toward merging as often as is acceptable.

Furthermore there are at least two independent ways to build on this PR, incorporating it in the algorithm and using it on the output, which could be done in parallel and may each take several attempts.

mpizenberg:
That's a lot nicer error message @zanieb ! If this is useful as-is I'd agree with @Eh2406 that we can merge it.

src/range.rs
```diff
@@ -274,6 +309,35 @@ fn valid_segment<T: PartialOrd>(start: &Bound<T>, end: &Bound<T>) -> bool {
     }
 }
 
+/// group adjacent versions locations
```
zanieb:
```suggestion
/// Group adjacent versions locations.
///
```

src/range.rs
```diff
@@ -321,6 +385,51 @@ impl<V: Ord + Clone> Range<V> {
...
+                    Ordering::Greater => *i += 1,
+                }
+            }
+            Some(None)
+        });
+        let kept_segments = group_adjacent_locations(version_locations);
+        self.keep_segments(kept_segments)
+    }
+
+    /// simplify range with segments at given location bounds.
```
zanieb:
```suggestion
    /// Create a new range with a subset of segments at given location bounds.
    ///
    /// Each new segment is constructed from a pair of segments, taking the
    /// start of the first and the end of the second.
```

Is this right?

zanieb (approved):

Issue #150, opened 2023-11-10
Root package version should be optional


zanieb:
Example at https://github.com/zanieb/pubgrub/blob/zanie/examples/examples/unsat_root_messages.rs

In practice, the root package often does not have a real or known version number. To get around this, we just use a dummy minimum version for the root package but then the version appears in messaging e.g.

```
Because there is no version of foo in 1.0.0 and root 0.0.0 depends on foo 1.0.0, root 0.0.0 is forbidden.
```

To resolve this, a PubGrub user must:

- Define a `Package` type with  "root" and "not root" variants
- Implement a custom `Display` trait for `External` which requires copying the entire `Reporter` implementation

Ideally the root package version would be optional entirely or it would be an option to omit it from the report.

Eh2406:
Thank you for trying PubGrub! I appreciate the production feedback.

I am just giving first reactions here. I may disagree with the opinion stated upon further reflection or the arrival of evidence.

- We should definitely have an example, in the section in the guide, about having a `enum<P> {Root, Else(P)}`! (If nothing else that can show me how painful it really is in practice.)
- If getting the error messages to display correctly requires "copying the entire `Reporter` implementation" than something is very wrong. We definitely need to improve the API so that this is not required! Suggestions are welcome.
- Adding a way to take a `derivation_tree` and strip out mentions of the root seems like a very helpful thing for us to provide. Once we have that implemented, we should see if there are other generalizations worth making.
- The dart implementation has a micro optimization where all references to the `root` packages are stripped out of incompatibilities. This adds a lot of special cases that I don't think pulls its weight. In general, the more internal details that know what the root package is (even to strip it out) the less happy I am. That being said, alternative APIs can be made that mainly affect to code in `solver` I am open to exploring it.

Reading over your statement again and my response, I may be responding to "my repackage doesn't have a name" not "my repackage doesn't have a version". If I have misunderstood and the two topics are not deeply intertwined, my responses may be entirely irrelevant.

zanieb:
Thanks for the response!

`enum<P> {Root, Else(P)}` is not a problem itself. We actually need to do `Root(Option<String>)` to allow it to have a name in some cases but that's fine.

The greater problem is that the version of `root` is meaningless and propagated through the reporting system. Here's an example of the code I wrote to strip versions from the display of the root package

https://github.com/zanieb/pubgrub/blob/zanie/examples/examples/unsat_root_message_no_version.rs

[...]

PR #158, opened 2023-11-30
feat: add report formatter trait for custom reporter output

zanieb:
Adds a `ReportFormatter` trait and a `Reporter::report_with_formatter` method allowing customization of formatting the report without reimplementing reporter internals. Currently, we use `format_external` and `format_terms` but we can refine formatting in the future as needed e.g. all the `explain` methods in the reporter.

Adds a `DefaultStringReportFormatter` which implements the existing format for the `DefaultStringReporter`.

You can see how much easier this is to use in the example diff at https://github.com/pubgrub-rs/pubgrub/commit/725ec87c7ef7b1a7efbe90f91375b8d685318aa2

Some existing discussion at https://github.com/zanieb/pubgrub/pull/10

Addresses some of the concerns in https://github.com/pubgrub-rs/pubgrub/issues/150

### DIFF
--- a/examples/unsat_root_message_no_version.rs
+++ b/examples/unsat_root_message_no_version.rs
@@ -1,0 +1,469 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use pubgrub::error::PubGrubError;
+use pubgrub::range::Range;
+use pubgrub::report::Reporter;
+use pubgrub::solver::{resolve, OfflineDependencyProvider};
+use pubgrub::version::SemanticVersion;
+
+use std::fmt::{self, Display};
+
+use pubgrub::report::{DerivationTree, Derived, External};
+use pubgrub::term::Term;
+use pubgrub::type_aliases::Map;
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum Package {
+    Root,
+    Package(String),
+}
+
+impl Display for Package {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Package::Root => write!(f, "root"),
+            Package::Package(name) => write!(f, "{}", name),
+        }
+    }
+}
+
+/// Derivative of [`pubgrub::report::DefaultStringReporter`] for customized display
+/// of package resolution errors.
+pub struct CustomStringReporter {
+    /// Number of explanations already with a line reference.
+    ref_count: usize,
+    /// Shared nodes that have already been marked with a line reference.
+    /// The incompatibility ids are the keys, and the line references are the values.
+    shared_with_ref: Map<usize, usize>,
+    /// Accumulated lines of the report already generated.
+    lines: Vec<String>,
+}
+
+impl CustomStringReporter {
+    /// Initialize the reporter.
+    fn new() -> Self {
+        Self {
+            ref_count: 0,
+            shared_with_ref: Map::default(),
+            lines: Vec::new(),
+        }
+    }
+
+    fn build_recursive(&mut self, derived: &Derived<Package, Range<SemanticVersion>>) {
+        self.build_recursive_helper(derived);
+        if let Some(id) = derived.shared_id {
+            if self.shared_with_ref.get(&id).is_none() {
+                self.add_line_ref();
+                self.shared_with_ref.insert(id, self.ref_count);
+            }
+        };
+    }
+
+    fn build_recursive_helper(&mut self, current: &Derived<Package, Range<SemanticVersion>>) {
+        match (&*current.cause1, &*current.cause2) {
+            (DerivationTree::External(external1), DerivationTree::External(external2)) => {
+                // Simplest case, we just combine two external incompatibilities.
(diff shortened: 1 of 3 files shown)


src/report.rs
```diff
@@ -17,8 +17,15 @@ pub trait Reporter<P: Package, VS: VersionSet> {
     type Output;
```
Eh2406, line 17: I don't think this trait pulls its weight, I think users would be just as well served by freestanding functions. Do you want to do that in this PR or leave it for follow-up?

src/report.rs
```diff
@@ -173,6 +180,50 @@ impl<P: Package, VS: VersionSet> fmt::Display for External<P, VS> {
     }
 }
 
+/// Trait for formatting outputs in the reporter.
+pub trait ReportFormatter<P: Package, VS: VersionSet> {
+    /// Output type of the report.
+    type Output;
```
Eh2406, line 186: The current code only supports `Output = String`, eventually I prefer these functions return `impl Display` or take a `&mut Write`. Do you think `Output` pulls its weight? We could have the methods return `String` for now and fix it in a follow-up, or have the methods take `&mut Write` and return the bubbled up error. Thoughts?

src/report.rs
```diff
@@ -194,8 +245,12 @@ impl DefaultStringReporter {
         }
     }
 
-    fn build_recursive<P: Package, VS: VersionSet>(&mut self, derived: &Derived<P, VS>) {
-        self.build_recursive_helper(derived);
+    fn build_recursive<P: Package, VS: VersionSet>(
+        &mut self,
+        derived: &Derived<P, VS>,
+        formatter: &impl ReportFormatter<P, VS, Output = String>,
```
Eh2406: Should this use `impl` or should it be a generic? The `impl`, but generic allows using more contacts. I'm not sure which one make sense here. Thoughts?

examples/unsat_root_message_no_version.rs
```diff
@@ -0,0 +1,147 @@
...
+struct CustomReportFormatter;
+
+impl ReportFormatter<Package, Range<SemanticVersion>> for CustomReportFormatter {
+    type Output = String;
+
+    fn format_terms(&self, terms: &Map<Package, Term<Range<SemanticVersion>>>) -> String {
+        let terms_vec: Vec<_> = terms.iter().collect();
+        match terms_vec.as_slice() {
+            [] => "version solving failed".into(),
+            // TODO: special case when that unique package is root.
```
mpizenberg: I think that TODO comment is not relevant anymore since you did special-case root.

mpizenberg:
I like it a lot too!

src/report.rs
zanieb: The generic feels kind of complicated since it also requires P and VS, but I'm not sure about the trade-offs.

cc @burntsushi

src/report.rs
zanieb: I think it makes sense to mirror the `Reporter` trait for now. We can tackle the rest in a subsequent refactor? I think that design deserves some dedicated consideration.

src/report.rs
zanieb: I agree this trait is a little weird unless users are expected to be passing around generic reporters. It makes a bit more sense now that there's another method, but I agree it seems sensible to remove. I'd prefer to track this in a new issue.

examples/unsat_root_message_no_version.rs
zanieb:
```suggestion
```

src/report.rs
BurntSushi: Sorry, I don't think I quite grok the question here. Right now, this signature is, I believe, essentially equivalent to the following:

```rust
fn build_recursive<P: Package, VS: VersionSet, RF: ReportFormatter<P, VS, Ouptut = String>>(
    &mut self,
    derived: &Derived<P, VS>,
    formatter: &RF,
)
```

The only difference is that my signature lets you turbofish the type of `RF`. e.g., `build_recursive::<ConcreteP, ConcreteVS, ConcreteRF>()`. Since it's an internal routine, I don't think it matters much if you don't need to turbofish. If it were public, I probably would write the signature without `impl Trait` so that folks can use turbofish.

(I feel like I'm not answering the right question though.)

Eh2406: That seems like the question at hand. Which suggests that any public function/method should use  `formatter: &RF`. The private function/method can use whichever one is more convenient. If somebody needs to make the decision, I ever so slightly lean toward being consistent.

zanieb: Sounds like we should use a generic.

zanieb: https://github.com/pubgrub-rs/pubgrub/pull/158/commits/c0c749da47c7d6ce9686a424b85412e6e40fb7a6

Eh2406 (approved):
This is a huge step forward! There are more fixes to come, so let's merge this as is. Unless @mpizenberg have anything he wants done in this PR?

mpizenberg:
Nothing to add, this is great!

PR #166, opened 2023-12-01
ci: replace unmaintained action-rs

### LINKED ISSUE #164: Move off actions-rs
[actions-rs organization became unmaintained](https://www.reddit.com/r/rust/comments/vyx4oj/actionsrs_organization_became_unmaintained/) we should move our CI to no longer use it. I think everything were using it for can just as easily be done with GA provided rustup, but I am no CI expert.

mpizenberg:
This fixes #164 

[actions-rs organization became unmaintained](https://www.reddit.com/r/rust/comments/vyx4oj/actionsrs_organization_became_unmaintained/) we should move our CI to no longer use it.

In this PR, I opted for using dtolnay's action. It is straightforward to use.

I've removed caches for the `test` and `check_documentation` jobs which had caching. The docs job is now faster, 15s against 20s. And the test job is slightly slower (50s against 43s roughly). Let's avoid the added complexity of caching as long as we can keep CI under a minute or so.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,16 +64,11 @@ jobs:
     name: Source code is formatted
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - name: Install stable Rust
-        uses: actions-rs/toolchain@v1
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


Eh2406 (approved):
I think this looks good!

PR #168, opened 2023-12-14
feat!: allocation free errors

### LINKED ISSUE #116: More compatible error type

Eh2406:
This replaces the work in #136 and fixes #116 this allows the user to pick the error type that best suits their needs.

Working on this is a prerequisite for allowing the user to associate data with unavailable dependencies.

### DIFF
--- a/examples/caching_dependency_provider.rs
+++ b/examples/caching_dependency_provider.rs
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: MPL-2.0
 
 use std::cell::RefCell;
-use std::error::Error;
 
 use pubgrub::package::Package;
 use pubgrub::range::Range;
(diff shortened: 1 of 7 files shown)


mpizenberg (approved):
Seems like something unavoidable, looks good to me. Thanks @Eh2406 !

zanieb (approved):

PR #170, opened 2023-12-18
perf: invalidate less contradicted_incompatibilities

Eh2406:
This extends the work from #88. To quote:

> It is possible to figure out witch IncompIds are still used after a backtrack, but so far nothing that is pulls its weight.

Looking at our normal benchmarks this is still within the noise. Looking at `hobo` it is a 5% improvement. However it does reduce the number of times intersection is called, and our users are telling us that intersections in their code are more expensive than in our benchmarks.

There are more efficient alternatives to calling `retain(|_, dl| *dl <= decision_level)`. (The insertion order is also sorted by `dl`, so binary search and truncate is an option. Or ...) But I have not seen it in profiles so I don't think it is worth more complexity at this time.

**If** this is a big win for our users then we should merge soon. Otherwise, this can stick around until other optimizations make the background noise smaller.

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -27,8 +27,9 @@ pub struct State<P: Package, VS: VersionSet, Priority: Ord + Clone> {
 
     incompatibilities: Map<P, Vec<IncompId<P, VS>>>,
 
-    /// Store the ids of incompatibilities that are already contradicted
-    /// and will stay that way until the next conflict and backtrack is operated.
+    /// Store the ids of incompatibilities that are already contradicted.
+    /// For each one keep track of the decision level when it was found to be contradicted.
+    /// These will stay contradicted until we have backtracked beyond its associated decision level.
     contradicted_incompatibilities: Map<IncompId<P, VS>, DecisionLevel>,
 
     /// All incompatibilities expressing dependencies,
@@ -226,6 +227,7 @@ impl<P: Package, VS: VersionSet, Priority: Ord + Clone> State<P, VS, Priority> {
     ) {
         self.partial_solution
             .backtrack(decision_level, &self.incompatibility_store);
+        // Remove contradicted incompatibilities that depend on decisions we just backtracked away.
         self.contradicted_incompatibilities
             .retain(|_, dl| *dl <= decision_level);
         if incompat_changed {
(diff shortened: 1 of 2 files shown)


konstin:
Just cherry picking 976d03ed17ba64af56279b8844afe69892065383 onto our tagged branch speeds my test up from a good 4s to 1.7s-1.8s! We'll have to rebase onto the main to measure with the others speedups, once that is done i'll make a proper hyperfine benchmark and report back.

Eh2406:
That sounds like a big win! Hopefully this can get a review and get merged soon.

mpizenberg (approved):
So what this PR does if I've understood it correctly, is record the decision level for each contradicted incompatibility. Then when backtracking, instead of resetting all contradicted incompats, we keep those from decisions previous to the backtrack point.

Sounds logical and it's great that this improves real world use cases.

src/internal/core.rs
```diff
@@ -29,7 +29,7 @@ pub struct State<P: Package, VS: VersionSet, Priority: Ord + Clone> {
 
     /// Store the ids of incompatibilities that are already contradicted
     /// and will stay that way until the next conflict and backtrack is operated.
```
mpizenberg: Can you improve the comment here to add that we record decision levels to avoid a full reset when backtracking?
That's the only remark I have.

Eh2406: Good call. I will fix this tomorrow!

konstin:
When i cherry-pick 976d03ed17ba64af56279b8844afe69892065383 onto f64c499f83e5798a583f9850fac28d2e7bf1dfbf the speedup is lower but still there. For the `tf-models-nightly` benchmark, which includes running through a lot of incompatible versions:

```
Benchmark 1: main
  Time (mean ± σ):      5.235 s ±  0.084 s    [User: 4.932 s, System: 0.295 s]
  Range (min … max):    5.157 s …  5.417 s    10 runs

Benchmark 2: branch
  Time (mean ± σ):      4.875 s ±  0.066 s    [User: 4.578 s, System: 0.291 s]
  Range (min … max):    4.795 s …  4.970 s    10 runs

Summary
  branch ran
    1.07 ± 0.02 times faster than main
```

Eh2406:
That makes sense. When #163 applies, as it does in your case, it entirely removes the old incompatibilities from consideration. This PR keeps track of the fact that the old incompatibilities can be skipped, which is much less useful if they're not being considered at all. I'm glad to see that despite the redundancy this PR pulls its weight.

It is easy to imagine cases where #163 does not apply, where this PR could have a much larger impact. Closer to the one you originally measured.

PR #172, opened 2023-12-21
Avoid loss of data in Range.simplify

konstin:
Upstream port of https://github.com/zanieb/pubgrub/pull/12

### DIFF
--- a/src/range.rs
+++ b/src/range.rs
@@ -415,21 +415,26 @@ impl<V: Ord + Clone> Range<V> {
         Self { segments: output }.check_invariants()
     }
 
-    /// Returns a simpler Range that contains the same versions
+    /// Returns a simpler Range that contains the same versions.
     ///
-    /// For every one of the Versions provided in versions the existing range and
-    /// the simplified range will agree on whether it is contained.
+    /// For every one of the Versions provided in versions the existing range and the simplified range will agree on whether it is contained.
     /// The simplified version may include or exclude versions that are not in versions as the implementation wishes.
-    /// For example:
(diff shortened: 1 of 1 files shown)


tests/proptest.rs
```diff
@@ -115,7 +115,6 @@ fn timeout_resolve<P: Package, VS: VersionSet, DP: DependencyProvider<P, VS>>(
 }
 
 type NumVS = Range<NumberVersion>;
-type SemVS = Range<SemanticVersion>;
```
konstin: `cargo check` complained about this one.

Eh2406:
The existing `simplify` returns the smallest representation where all the versions get the same result. This feels like an important property to keep available. On the other hand the new `simplify` is probably better for error messages, and for some (not yet determined) reason works in contexts where the original doesn't.

Perhaps we should have a second function available:

```rust

pub fn bike_shed<'v, I>(&self, versions: I) -> Self
    where
        I: Iterator<Item = &'v V> + 'v,
        V: 'v,
    {
        // Do not simplify singletons
        if self.is_singleton() {
            return self.clone();
        }

        let simp = self.simplify(versions);
        
        if simp == Empty() {
            self.cone()
        } else {
            simp
        }
    }
```

@mpizenberg thoughts?

mpizenberg:
> for some (not yet determined) reason works in contexts where the original doesn't.

I might not have the enough context for this.

Personally, I'd not use the same simplify if one is for optimization and the other is for quality of errors. The issue is that the original simplify is lossy. So you can't really apply the alternative simplify if the data has already been simplified with loss of information.

PR #177, opened 2024-01-05
perf: specialize range operations for better performances

### LINKED ISSUE #135: Solving slows down dramatically when testing hundreds of versions

konstin:
In uv, we have a very slow case with bio_embeddings, which include boto3 and botocore. In the fully cached case, the great majority of time is spent in range operation with large ranges due to the many version. 
I replaced some range operations implemented by their mathematical definitions with optimized versions. This PR is based on profiling my test case, `bio_embeddings[all]` on python 3.12.

To reproduce, checkout `git clone https://github.com/astral-sh/uv` and replace the pubgrub dependency with https://github.com/astral-sh/pubgrub/pull/25, build with `cargo build --profile profiling` and run with `samply record`, i ran with `--rate 5000`.

**before** 12s

image

**after** 3s

image

The atomic/drop operations are clone and drop for our version type, which is an `Arc`.

This scenarios slow because we try a lot of versions:

```
Tried 101329 versions: pytest 2628, botocore 2627, blis 1314, boto3 1314, certifi 1314, charset-normalizer 1314, click 1314, conllu 1314, contourpy 1314, cycler 1314, cymem 1314, editdistance 1314, flaky 1314, flask 1314, flask-cors 1314, fonttools 1314, fsspec 1314, ftfy 1314, gevent 1314, huggingface-hub 1314, idna 1314, jinja2 1314, joblib 1314, jsonnet 1314, jsonpickle 1314, kiwisolver 1314, murmurhash 1314, networkx 1314, nltk 1314, numba 1314, numpydoc 1314, nvidia-cublas-cu12 1314, nvidia-cuda-cupti-cu12 1314, nvidia-cuda-nvrtc-cu12 1314, nvidia-cuda-runtime-cu12 1314, nvidia-cudnn-cu12 1314, nvidia-cufft-cu12 1314, nvidia-curand-cu12 1314, nvidia-cusolver-cu12 1314, nvidia-cusparse-cu12 1314, nvidia-nccl-cu12 1314, nvidia-nvtx-cu12 1314, overrides 1314, packaging 1314, parsimonious 1314, plac 1314, preshed 1314, protobuf 1314, pynndescent 1314, pyparsing 1314, python-dateutil 1314, pytorch-pretrained-bert 1314, pytorch-transformers 1314, pyyaml 1314, regex 1314, responses 1314, ruamel-yaml-clib 1314, safetensors 1314, sentencepiece 1314, six 1314, spacy 1314, sqlparse 1314, srsly 1314, sympy 1314, tenacity 1314, tensorboardx 1314, thinc 1314, threadpoolctl 1314, tokenizers 1314, typing-extensions 1314, tzdata 1314, unidecode 1314, wasabi 1314, word2number 1314, wrapt 1314, allennlp 27, torch 20, bio-embeddings[all] 10, bio-embeddings 4, biopython 4, gensim 4, h5py 4, matplotlib 3, numpy 3, pandas 3, plotly 3, ruamel-yaml 3, scikit-learn 3, scipy 3, Python 2, appdirs 2, importlib-metadata 2, jax-unirep 2, jaxlib 2, lock 2, alabaster 1, babel 1, bio-embeddings-allennlp 1, bio-embeddings-bepler 1, bio-embeddings-cpcprot 1, bio-embeddings-esm 1, bio-embeddings-plus 1, bio-embeddings-tape-proteins 1, blinker 1, docutils 1, filelock 1, greenlet 1, imagesize 1, iniconfig 1, itsdangerous 1, jmespath 1, llvmlite 1, markupsafe 1, mpmath 1, nvidia-nvjitlink-cu12 1, pillow 1, pluggy 1, pygments 1, pytz 1, requests 1, root 1, s3transfer 1, setuptools 1, smart-open 1, snowballstemmer 1, sphinx 1, sphinxcontrib-applehelp 1, sphinxcontrib-devhelp 1, sphinxcontrib-htmlhelp 1, sphinxcontrib-jsmath 1, sphinxcontrib-qthelp 1, sphinxcontrib-serializinghtml 1, tabulate 1, tqdm 1, transformers 1, umap-learn 1, urllib3 1, wcwidth 1, werkzeug 1, zope-event 1, zope-interface 1
```

(Taskset descreases the variance for a minimal slowdown, we're bound on single-threaded pubgrub in this benchmark and not on multithreaded io/parsing.)

```
$  taskset -c 0 hyperfine --warmup 1 "../uv/target/profiling/main-uv pip compile ../uv/scripts/requirements/bio_embeddings.in"  "../uv/target/profiling/branch-uv pip compile ../uv/scripts/requirements/bio_embeddings.in" 
Benchmark 1: ../uv/target/profiling/main-uv pip compile ../uv/scripts/requirements/bio_embeddings.in
  Time (mean ± σ):     12.321 s ±  0.064 s    [User: 12.014 s, System: 0.300 s]
  Range (min … max):   12.224 s … 12.406 s    10 runs
 
Benchmark 2: ../uv/target/profiling/branch-uv pip compile ../uv/scripts/requirements/bio_embeddings.in
  Time (mean ± σ):      3.109 s ±  0.004 s    [User: 2.782 s, System: 0.321 s]
  Range (min … max):    3.103 s …  3.116 s    10 runs
 
Summary
  ../uv/target/profiling/branch-uv pip compile ../uv/scripts/requirements/bio_embeddings.in ran
    3.96 ± 0.02 times faster than ../uv/target/profiling/main-uv pip compile ../uv/scripts/requirements/bio_embeddings.in
```

I'd say this closes #135.

---

The lack combination of lack of `Ord` on `Bound` and `Unbounded` being `-inf` or `inf` depending on the (untyped) context makes implementing these manually tedious, i wonder if we could add some abstraction?

---

The first two commits of this PR can be split out into separate PRs. From main over each of the three commits:

image

image

image

image

### DIFF
--- a/src/internal/partial_solution.rs
+++ b/src/internal/partial_solution.rs
@@ -501,7 +501,7 @@ impl<P: Package, VS: VersionSet> PackageAssignments<P, VS> {
         let idx = self
             .dated_derivations
             .as_slice()
-            .partition_point(|dd| dd.accumulated_intersection.intersection(start_term) != empty);
+            .partition_point(|dd| !dd.accumulated_intersection.is_disjoint(start_term));
         if let Some(dd) = self.dated_derivations.get(idx) {
             debug_assert_eq!(dd.accumulated_intersection.intersection(start_term), empty);
             return (Some(dd.cause), dd.global_index, dd.decision_level);
--- a/src/internal/small_map.rs
+++ b/src/internal/small_map.rs
@@ -99,6 +99,49 @@ impl<K: PartialEq + Eq + Hash, V> SmallMap<K, V> {
             }
         };
     }
+
+    /// Returns a reference to the value for one key and a copy of the map without the key.
+    ///
+    /// This is an optimization over the following, where we only need a reference to `t1`. It
+    /// avoids cloning and then drop the ranges in each `prior_cause` call.
+    /// ```ignore
+    /// let mut package_terms = package_terms.clone();
+    //  let t1 = package_terms.remove(package).unwrap();
+    /// ```
+    pub fn split_one(&self, key: &K) -> Option<(&V, Self)>
+    where
+        K: Clone,
+        V: Clone,
+    {
+        match self {
+            Self::Empty => None,
+            Self::One([(k, v)]) => {
+                if k == key {
+                    Some((v, Self::Empty))
+                } else {
+                    None
+                }
+            }
+            Self::Two([(k1, v1), (k2, v2)]) => {
+                if k1 == key {
+                    Some((v1, Self::One([(k2.clone(), v2.clone())])))
+                } else if k2 == key {
+                    Some((v2, Self::One([(k1.clone(), v1.clone())])))
+                } else {
+                    None
+                }
+            }
+            Self::Flexible(map) => {
+                if let Some(value) = map.get(key) {
+                    let mut map = map.clone();
+                    map.remove(key).unwrap();
+                    Some((value, Self::Flexible(map)))
+                } else {
+                    None
+                }
+            }
+        }
+    }
 }
 
 impl<K: Clone + PartialEq + Eq + Hash, V: Clone> SmallMap<K, V> {
--- a/src/range.rs
+++ b/src/range.rs
@@ -374,6 +374,19 @@ fn end_before_start_with_gap<V: PartialOrd>(end: &Bound<V>, start: &Bound<V>) ->
     }
 }
 
+/// The end of one interval is before the start of the next one, meaning they don't share any
+/// version.
+fn end_before_start<V: PartialOrd>(end: Bound<&V>, start: Bound<&V>) -> bool {
+    match (end, start) {
+        (_, Unbounded) => false,
+        (Unbounded, _) => false,
+        (Included(left), Included(right)) => left < right,
+        (Included(left), Excluded(right)) => left <= right,
+        (Excluded(left), Included(right)) => left <= right,
+        (Excluded(left), Excluded(right)) => left <= right,
+    }
+}
+
 /// Group adjacent versions locations.
 ///
 /// ```text
@@ -559,6 +572,60 @@ impl<V: Ord + Clone> Range<V> {
         }
 
         // The remaining element(s) can't intersect anymore
+        true
+    }
+
+    /// Return true if any `V` that is contained in `self` is also contained in `other`.
+    ///
+    /// Note that we don't know that set of all existing `V`s here, so we only check if all
+    /// segments `self` are contained in a segment of `other`.
+    fn subset_of(&self, other: &Self) -> bool {
+        let mut containing_iter = other.segments.iter();
+        let mut subset_iter = self.segments.iter();
+        let Some(mut containing_elem) = containing_iter.next() else {
+            // As long as we have subset elements, we need containing elements
+            return subset_iter.next().is_none();
+        };
+
+        for subset_elem in subset_iter {
+            // Check if the current containing element ends before the subset element.
+            // There needs to be another containing element for our subset element in this case.
+            while end_before_start(containing_elem.end_bound(), subset_elem.start_bound()) {
+                if let Some(containing_elem_) = containing_iter.next() {
+                    containing_elem = containing_elem_;
+                } else {
+                    return false;
+                };
+            }
+
+            let start_contained = match (containing_elem.start_bound(), subset_elem.start_bound()) {
+                (Unbounded, _) => true,
+                (Included(_) | Excluded(_), Unbounded) => false,
+                // This is the only case where the subset is bound is "wider" than containing bound ...
+                (Excluded(left), Included(right)) => left < right,
+                // ... while in the other cases they can share the point
+                (Included(left), Included(right)) => left <= right,
+                (Included(left), Excluded(right)) => left <= right,
+                (Excluded(left), Excluded(right)) => left <= right,
+            };
+
+            let end_contained = match (subset_elem.end_bound(), containing_elem.end_bound()) {
+                (_, Unbounded) => true,
+                (Unbounded, Included(_) | Excluded(_)) => false,
+                // This is the only case where the subset is bound is "wider" than containing bound ...
+                (Included(left), Excluded(right)) => left < right,
+                // ... while in the other cases they can share the point
+                (Included(left), Included(right)) => left <= right,
+                (Excluded(left), Included(right)) => left <= right,
+                (Excluded(left), Excluded(right)) => left <= right,
+            };
+
+            if !(start_contained && end_contained) {
+                // This subset element is not contained
+                return false;
+            }
+        }
+
         true
     }
 
--- a/src/range.rs
+++ b/src/range.rs
@@ -301,7 +301,7 @@ impl<V: Ord> Range<V> {
                 assert!(end_before_start_with_gap(&p[0].1, &p[1].0));
             }
             for (s, e) in self.segments.iter() {
-                assert!(valid_segment(&s, &e));
+                assert!(valid_segment(s, e));
             }
         }
         self
@@ -518,7 +518,7 @@ impl<V: Ord + Clone> Range<V> {
             // But, we already know that the segments in our input are valid.
             // So we do not need to check if the `start`  from the input `end` came from is smaller then `end`.
             // If the `other_start` is larger than end, then the intersection will be invalid.
-            if !valid_segment(&other_start, &end) {
+            if !valid_segment(other_start, end) {
                 // Note: We can call `this_iter.next_if(!valid_segment(other_start, this_end))` in a loop.
                 // But the checks make it slower for the benchmarked inputs.
                 continue;
--- a/src/range.rs
+++ b/src/range.rs
@@ -298,14 +298,7 @@ impl<V: Ord> Range<V> {
     fn check_invariants(self) -> Self {
         if cfg!(debug_assertions) {
             for p in self.segments.as_slice().windows(2) {
-                match (&p[0].1, &p[1].0) {
-                    (Included(l_end), Included(r_start)) => assert!(l_end < r_start),
-                    (Included(l_end), Excluded(r_start)) => assert!(l_end < r_start),
-                    (Excluded(l_end), Included(r_start)) => assert!(l_end < r_start),
-                    (Excluded(l_end), Excluded(r_start)) => assert!(l_end <= r_start),
-                    (_, Unbounded) => panic!(),
-                    (Unbounded, _) => panic!(),
-                }
+                assert!(end_before_start_with_gap(&p[0].1, &p[1].0));
             }
             for (s, e) in self.segments.iter() {
                 assert!(valid_segment(s, e));
@@ -457,7 +450,6 @@ impl<V: Ord + Clone> Range<V> {
                     let accumulator_end = match (accumulator_.1, smaller_interval.1) {
                         (_, Unbounded) | (Unbounded, _) => &Unbounded,
                         (Included(l), Excluded(r) | Included(r)) if l == r => accumulator_.1,
-                        (Excluded(l) | Included(l), Included(r)) if l == r => smaller_interval.1,
                         (Included(l) | Excluded(l), Included(r) | Excluded(r)) => {
                             if l > r {
                                 accumulator_.1
--- a/src/range.rs
+++ b/src/range.rs
@@ -221,19 +221,32 @@ impl<V: Ord> Range<V> {
         })
     }
 
-    /// Returns true if the this Range contains the specified value.
+    /// Returns true if this Range contains the specified value.
     pub fn contains(&self, v: &V) -> bool {
-        for segment in self.segments.iter() {
-            match within_bounds(v, segment) {
-                Ordering::Less => return false,
-                Ordering::Equal => return true,
-                Ordering::Greater => (),
-            }
-        }
-        false
+        self.segments
+            .binary_search_by(|segment| {
+                let lower_bound_greater = match segment {
+                    (Excluded(start), _) => v <= start,
+                    (Included(start), _) => v < start,
+                    (Unbounded, _) => false,
+                };
+                if lower_bound_greater {
+                    return Ordering::Greater;
+                }
+                let upper_bound_greater = match segment {
+                    (_, Unbounded) => true,
+                    (_, Included(end)) => v <= end,
+                    (_, Excluded(end)) => v < end,
+                };
+                if upper_bound_greater {
+                    return Ordering::Equal;
+                }
+                Ordering::Less
+            })
+            .is_ok()
     }
 
-    /// Returns true if the this Range contains the specified values.
+    /// Returns true if this Range contains the specified values.
     ///
     /// The `versions` iterator must be sorted.
     /// Functionally equivalent to `versions.map(|v| self.contains(v))`.
--- a/src/range.rs
+++ b/src/range.rs
@@ -222,27 +222,14 @@ impl<V: Ord> Range<V> {
     }
 
     /// Returns true if this Range contains the specified value.
-    pub fn contains(&self, v: &V) -> bool {
+    pub fn contains(&self, version: &V) -> bool {
         self.segments
             .binary_search_by(|segment| {
-                let lower_bound_greater = match segment {
-                    (Excluded(start), _) => v <= start,
-                    (Included(start), _) => v < start,
-                    (Unbounded, _) => false,
-                };
-                if lower_bound_greater {
-                    return Ordering::Greater;
-                }
-                let upper_bound_greater = match segment {
-                    (_, Unbounded) => true,
-                    (_, Included(end)) => v <= end,
-                    (_, Excluded(end)) => v < end,
-                };
-                if upper_bound_greater {
-                    return Ordering::Equal;
-                }
-                Ordering::Less
+                // We have to reverse because we need the segment wrt to the version, while
+                // within bounds tells us the version wrt to the segment.
+                within_bounds(version, segment).reverse()
             })
+            // An equal interval is one that contains the version
             .is_ok()
     }
 
@@ -328,10 +315,16 @@ impl<V: Ord> Range<V> {
     }
 }
 
-fn within_bounds<V: PartialOrd>(v: &V, segment: &Interval<V>) -> Ordering {
+/// The ordering of the version wrt to the interval.
+/// ```text
+///      |-------|
+///   ^      ^      ^
+///   less   equal  greater
+/// ```
+fn within_bounds<V: PartialOrd>(version: &V, segment: &Interval<V>) -> Ordering {
     let below_lower_bound = match segment {
-        (Excluded(start), _) => v <= start,
-        (Included(start), _) => v < start,
+        (Excluded(start), _) => version <= start,
+        (Included(start), _) => version < start,
         (Unbounded, _) => false,
     };
     if below_lower_bound {
@@ -339,8 +332,8 @@ fn within_bounds<V: PartialOrd>(v: &V, segment: &Interval<V>) -> Ordering {
     }
     let below_upper_bound = match segment {
         (_, Unbounded) => true,
-        (_, Included(end)) => v <= end,
-        (_, Excluded(end)) => v < end,
+        (_, Included(end)) => version <= end,
+        (_, Excluded(end)) => version < end,
     };
     if below_upper_bound {
         return Ordering::Equal;
--- a/src/range.rs
+++ b/src/range.rs
@@ -341,13 +341,43 @@ fn within_bounds<V: PartialOrd>(version: &V, segment: &Interval<V>) -> Ordering 
     Ordering::Greater
 }
 
+/// A valid segment is one where at least one version fits between start and end
 fn valid_segment<T: PartialOrd>(start: &Bound<T>, end: &Bound<T>) -> bool {
     match (start, end) {
+        // Singleton interval are allowed
         (Included(s), Included(e)) => s <= e,
         (Included(s), Excluded(e)) => s < e,
         (Excluded(s), Included(e)) => s < e,
         (Excluded(s), Excluded(e)) => s < e,
         (Unbounded, _) | (_, Unbounded) => true,
+    }
+}
+
+/// The end of one interval is before the start of the next one, so they can't be concatenated
+/// into a single interval. The `union` method calling with both intervals and then the intervals
+/// switched. If either is true, the intervals are separate in the union and if both are false, they
+/// are merged.
+/// ```text
+/// True for these two:
+///  |----|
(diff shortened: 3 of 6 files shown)


src/term.rs
```diff
@@ -19,6 +19,19 @@ pub enum Term<VS: VersionSet> {
     Negative(VS),
 }
 
+impl<VS: VersionSet> PartialEq for Term<VS> {
```
konstin: I think this is currently wrong on dev?

src/term.rs
```diff
@@ -98,10 +111,47 @@ impl<VS: VersionSet> Term<VS> {
...
+        match (self, other) {
+            (Self::Positive(r1), Self::Positive(r2)) => Self::Positive(r1.union(r2)),
+            (Self::Positive(r1), Self::Negative(r2)) => Self::Positive(r1.union(&r2.complement())),
+            (Self::Negative(r1), Self::Positive(r2)) => Self::Positive(r1.complement().union(r2)),
+            (Self::Negative(r1), Self::Negative(r2)) => Self::Negative(r1.intersection(r2)),
+        }
+    }
+
+    /// TODO
+    pub(crate) fn completes(&self, other: &Self) -> bool {
```
konstin: Not sure if it's worth it, but we could check if we even need to compute the union

src/term.rs
Eh2406: Thank you for opening a [issue](https://github.com/pubgrub-rs/pubgrub/issues/178) to discuss this.

src/range.rs
```diff
@@ -342,10 +342,79 @@ fn group_adjacent_locations(
 impl<V: Ord + Clone> Range<V> {
     /// Computes the union of this `Range` and another.
     pub fn union(&self, other: &Self) -> Self {
-        self.complement()
-            .intersection(&other.complement())
-            .complement()
-            .check_invariants()
+        let end_before_start_exclusive = |end: &Bound<_>, start: &Bound<_>| match (end, start) {
```
Eh2406: This looks similar to the `valid_segment` helper. We should probably put the helpers together and document how they differ.

Eh2406:
Just dropping a back link to https://github.com/pubgrub-rs/pubgrub/pull/59

src/term.rs
```diff
@@ -98,10 +111,47 @@ impl<VS: VersionSet> Term<VS> {
...
+            // term excludes. Inversely, if there is a region outside the negative, they overlap in this region.
+            (Self::Positive(r1), Self::Negative(r2)) => r1.subset_of(r2),
+            (Self::Negative(r1), Self::Positive(r2)) => r2.subset_of(r1),
+        }
+    }
+
     /// Compute the union of two terms.
     /// If at least one term is negative, the union is also negative.
     pub(crate) fn union(&self, other: &Self) -> Self {
-        (self.negate().intersection(&other.negate())).negate()
```
Eh2406, line 104: Inlining this code may be a micro optimization, saving a clone or two, but definitely seems worth doing.

konstin:
The new version is slightly slower but passes the tests:
```
$ hyperfine --warmup 1 --runs 3 "target/profiling/main bio_embeddings.in"  "target/profiling/branch bio_embeddings.in"
Benchmark 1: target/profiling/main bio_embeddings.in
  Time (mean ± σ):     12.007 s ±  0.035 s    [User: 11.654 s, System: 0.301 s]
  Range (min … max):   11.977 s … 12.045 s    3 runs

Benchmark 2: target/profiling/branch bio_embeddings.in
  Time (mean ± σ):      6.961 s ±  0.017 s    [User: 6.565 s, System: 0.323 s]
  Range (min … max):    6.942 s …  6.974 s    3 runs

Summary
  target/profiling/branch bio_embeddings.in ran
    1.72 ± 0.01 times faster than target/profiling/main bio_embeddings.in
```

It spends 45% in union, mostly `Arc` drop, add and vec pushes.

src/range.rs
```diff
@@ -494,6 +570,93 @@ impl<T: Debug + Display + Clone + Eq + Ord> VersionSet for Range<T> {
     fn union(&self, other: &Self) -> Self {
         Range::union(self, other)
     }
+
+    fn is_disjoint(&self, other: &Self) -> bool {
```
Eh2406, line 713: This is a very elegant implementation!

src/range.rs
```diff
@@ -494,6 +570,93 @@ impl<T: Debug + Display + Clone + Eq + Ord> VersionSet for Range<T> {
...
+        let end_before_start = |end, start| match (end, start) {
```
Eh2406: This is `!valid_segment` I think the file stays easier to read if we reduce the number of lookup table helpers, and document how they differ.

src/version_set.rs
```diff
@@ -57,4 +57,14 @@ pub trait VersionSet: Debug + Display + Clone + Eq {
...
+
+    /// Whether the range have no overlapping segmets
+    fn is_disjoint(&self, other: &Self) -> bool {
+        self.intersection(other) == Self::empty()
+    }
+
+    /// Whether all range of `self` are contained in `other`
+    fn subset_of(&self, other: &Self) -> bool {
+        self == &self.intersection(other)
+    }
```
Eh2406, line 69: In https://github.com/pubgrub-rs/pubgrub/pull/59 @mpizenberg was concerned how much complexity this adds to the api. So I would appreciate getting his opinion now.

src/internal/partial_solution.rs
```diff
@@ -501,7 +501,7 @@ impl<P: Package, VS: VersionSet> PackageAssignments<P, VS> {
         let idx = self
             .dated_derivations
             .as_slice()
-            .partition_point(|dd| dd.accumulated_intersection.intersection(start_term) != empty);
+            .partition_point(|dd| !dd.accumulated_intersection.is_disjoint(start_term));
```
Eh2406, line 504: The other place where we could use these new methods is in `relation` in that context we want to know `is_disjoint` and `subset_of` for the same operands. The question becomes can the combination be computed faster if we asked them at the same time? If so does it justify having all three methods? @konstin, you spent the most time looking at optimizing the methods recently, what are your thoughts?

src/term.rs
Eh2406: Not sure I understand, how would this be used?

src/version_set.rs
konstin: Personally, i'm not too worried about adding methods which have default implementations. It doesn't add additional effort for the implementer but allows extra control where performance matters.

src/internal/partial_solution.rs
konstin: I don't see `relation_with` in the traces and `relation` seems to be almost exclusively map lookups:

image

src/range.rs
```diff
@@ -201,19 +201,32 @@ impl<V: Ord> Range<V> {
...
-            match within_bounds(v, segment) {
-                Ordering::Less => return false,
-                Ordering::Equal => return true,
-                Ordering::Greater => (),
-            }
-        }
-        false
+        self.segments
+            .binary_search_by(|segment| {
+                let lower_bound_greater = match segment {
```
Eh2406: Can this just be 

```
self.segments.binary_search_by(|segment| within_bounds(v, segment))
```

(`within_bounds` may have been added since you opened this PR.)

src/range.rs
```diff
@@ -201,19 +201,32 @@ impl<V: Ord> Range<V> {
...
+                Ordering::Less
+            })
+            .is_ok()
     }
 
-    /// Returns true if the this Range contains the specified values.
+    /// Returns true if this Range contains the specified values.
     ///
     /// The `versions` iterator must be sorted.
     /// Functionally equivalent to `versions.map(|v| self.contains(v))`.
```
Eh2406, line 239: `contains_many` can probably also start with a `binary_search`

konstin: I'd skip on modifying `contains_many`

src/internal/partial_solution.rs
Eh2406: If map lookups are a significant bottleneck I would love to see what performance you getting with:
https://github.com/pubgrub-rs/pubgrub/tree/p_id

konstin: It's not package ids, it's incompatibility ids in `self.contradicted_incompatibilities.contains_key(&incompat_id)` (5.8% in this run)

image

Eh2406: Yes, that is an expensive call, but it is not the hashmap lookup that is slowing `relation`.
`terms_intersection_for_package` is a lookup by package ID.

konstin: I tried to cherry pick the commit but it unfortunately has to many merge conflicts with our main for me to try

Eh2406: Okay, we will experiment with rebasing that branch once we have this merged. Sorry for getting distracted.

Eh2406:
I had trouble thinking useful thoughts looking at the full diff all at once. So I pulled it into my editor and started a new branch adding each new piece of functionality one at a time in separate commits. Then adding the tests and cleaning up the nits I saw before moving onto the next piece of functionality. The result is https://github.com/pubgrub-rs/pubgrub/tree/jf/177-review consider it a "yes and" to the wonderful work in this PR. But I'm not sure on a practical level how to take the next step. Do you want to pull those changes into this PR? Or should I start a new one?

Anyway, I'm gonna go run some benchmarks.

konstin:
Thanks! I forced pushed your commits to this branch (there's backup branch just in case also) and added a clippy commit on top.

Eh2406:
time|Dev|PR|%
---|----|---|---
sudoku|44s|30s|32%
elm| 274ms | 277ms| ---
large| 11ms | 10ms| 9%
slow_135 | 19ms | 6.3ms| 66%
zuse | 311 ms | 297ms| 4.5%
hobo*| 1045s | 851s|18%

Eh2406 (approved):
I rewrote the things I didn't like, and so like what's left.

I'm going to give it a few days in case other people want to respond. Specifically for @mpizenberg about the public API.

If I don't hear back I will merge later this week.

src/internal/small_map.rs
```diff
@@ -99,6 +99,49 @@ impl<K: PartialEq + Eq + Hash, V> SmallMap<K, V> {
...
+                    Some((v2, Self::One([(k1.clone(), v1.clone())])))
+                } else {
+                    None
+                }
+            }
+            Self::Flexible(map) => {
+                if let Some(value) = map.get(key) {
+                    let mut map = map.clone();
+                    map.remove(key).unwrap();
+                    Some((value, Self::Flexible(map)))
```
mpizenberg, line 138: Do we make any assumption anywhere that `Flexible` must have at least 3 elements? or not? If not then fine. Maybe that was already not the case and I'm just realizing it now. How can we make sure we don't make that assumption in the rest of the code base?

src/range.rs
```diff
@@ -345,6 +346,56 @@ fn valid_segment<T: PartialOrd>(start: &Bound<T>, end: &Bound<T>) -> bool {
...
+///       ^ end    ^ start
+/// False for these two:
+///  |----|
+///     |-----|
+/// Here it depends: If they both exclude the position they share, there is a version in between
+/// them that blocks concatenation
+///  |----|
+///       |-----|
+/// ```
+fn end_before_start_with_gap<V: PartialOrd>(end: &Bound<V>, start: &Bound<V>) -> bool {
```
mpizenberg, line 366: what do you think of `non_contiguous_bounds`? or `contiguous_bounds` with the bool flipped?

src/range.rs
```diff
@@ -378,10 +429,60 @@ fn group_adjacent_locations(
 impl<V: Ord + Clone> Range<V> {
     /// Computes the union of this `Range` and another.
     pub fn union(&self, other: &Self) -> Self {
```
mpizenberg, line 431: Could we keep a private canonical implementation? and maybe check the optimized one against the canonical one when not compiling in release? Or maybe just property tests are enough?

src/range.rs
```diff
@@ -451,6 +544,72 @@ impl<V: Ord + Clone> Range<V> {
         Self { segments: output }.check_invariants()
     }
 
+    /// Return true if there can be no `V` so that `V` is contained in both `self` and `other`.
+    ///
+    /// Note that we don't know that set of all existing `V`s here, so we only check if the segments
+    /// are disjoint, not if no version is contained in both.
+    pub fn is_disjoint(&self, other: &Self) -> bool {
+        // The operation is symmetric
```
mpizenberg, line 552: Sounds like a nice property test to have regarding symmetry of the operator

src/term.rs
```diff
@@ -98,18 +106,46 @@ impl<VS: VersionSet> Term<VS> {
...
 
     /// Indicate if this term is a subset of another term.
     /// Just like for sets, we say that t1 is a subset of t2
     /// if and only if t1 ∩ t2 = t1.
-    #[cfg(test)]
     pub(crate) fn subset_of(&self, other: &Self) -> bool {
-        self == &self.intersection(other)
+        match (self, other) {
+            (Self::Positive(r1), Self::Positive(r2)) => r1.subset_of(r2),
+            (Self::Positive(r1), Self::Negative(r2)) => r1.subset_of(&r2.complement()),
```
Eh2406: `r1.subset_of(&r2.complement())` is `r1.is_disjoint(r2)`. This saves `complement` with its associated allocations and drop. On the Sudoku case I see an additional 30% drop to 20s!

mpizenberg (approved):
Thanks for doing this performance work! As suggested by @Eh2406 I was previously skeptic of this kind of inlining degrading the expressiveness of the computations. According to Jacob benchmarks, I'm still doubting it's the best course of action. On sub-second runs, it does not seem to fundamentally change speed, and for very long run, it has a non-negligeable impact, but I feel like this order of magnitude (multiple tens of seconds) is something that should be approached on a computation complexity level (solving the big-O bottlenecks) instead.

However I can also understand that 30% of something that takes a very long time is a non-negligible gain so I'm in favor if you want to merge this.

I've made a few suggestions, feel free to address or ignore them and go through with the merge.

zanieb:
@mpizenberg does the "real world" improvement of 12s to 3s on the `bio_embeddings` requirements in `uv` make this feel more worthwhile? Does that suggest that there is a missing niche in the benchmarks for the project (i.e. that Jacob ran)?

mpizenberg:
> @mpizenberg does the "real world" improvement of 12s to 3s on the `bio_embeddings` requirements in `uv` make this feel more worthwhile? Does that suggest that there is a missing niche in the benchmarks for the project (i.e. that Jacob ran)?

Indeed that's a substantial difference!

mpizenberg:
> @mpizenberg does the "real world" improvement of 12s to 3s on the bio_embeddings 

Do you know if some of the inlinings produce most of the improvements, or if they all make small improvements that total to the big difference?

src/internal/small_map.rs
konstin: We don't make this assumption, specifically remove on the flexible variant will not switch over to one or two if it gets smaller than three (https://github.com/astral-sh/pubgrub/blob/88d8d8c41a966ce6f225ec66c41fd12f2676ea8a/src/internal/small_map.rs#L65-L68)

src/range.rs
konstin: Yes, we have both: https://github.com/astral-sh/pubgrub/blob/6715dc75d24a5f4f68513f53db76221d1f6bdc91/src/version_set.rs#L52-L59 and https://github.com/astral-sh/pubgrub/blob/74f61fb9cef608d8475f1e00c8d0158ada12e009/src/range.rs#L954-L962

src/range.rs
konstin: Good idea, done

konstin:
> Do you know if some of the inlinings produce most of the improvements, or if they all make small improvements that total to the big difference?

This work is entirely profiling and benchmark driven. I started off with two smaller problems, `Range::contains` being slower than needed and excessive clone/drop in `prior_cause`. After the first two commits you can see in the flamegraphs in the PR description it's almost entirely `union` and `intersection` calls, being slow due to excessive clone/drop (`cor...` in the flamegraphs). We had previously seen the slow drop/clone and put our version in an `Arc`, so all you see now are clone/drop on `Arc`s.

`satisfier` was calling intersection a lot, but it wasn't interested in the intersection itself, it just wanted to know if the terms are disjoint, so i special cased this into `is_disjoint`. The union calls were calling intersection once and complement three times, so i unrolled union into a version that clones as few as possible. I don't have numbers on hand for individual changes, but each made a noticeable difference.

src/range.rs
mpizenberg: Yes, the property test is great. What I was wondering is if we should have both impls in the type, something like
```rs
// inside impl Range
pub fn union(&self, other: &Self) -> Self {
  // optimized one
  ...
  // when not in release, check against union_canonical
  ...
}
fn union_canonical(&self, other: &Self) -> Self {
  // canonical impl
}
```
And in prop tests, we can use union_canonical directly for comparisons.
But that's probably not pragmatic at all to have the check on each optimized function call when not compiling in release, and the prop test is largely enough.
Yeah I think so. Forget about this one.

Eh2406:
> I was previously skeptic of this kind of inlining degrading the expressiveness of the computations.

Yesterday I went back and looked at my code from when I originally proposed this. You are right to reject that code. It was inscrutable. The code Konstin has now is much better.

> does the "real world" improvement of 12s to 3s on the `bio_embeddings` requirements in `uv` make this feel more worthwhile? Does that suggest that there is a missing niche in the benchmarks for the project (i.e. that Jacob ran)?

There is definitely a gap in the benchmarking infrastructure I have available. I can't seem to get a benchmark that has the performance characteristics of your real world data. Help on this would definitely be appreciated. The closest I have is `slow_135`, which is not nearly as slow as your original report, but has a similar shaped profile. The 66% improvement for that benchmark is nicely similar to the 75% your reporting.

> We had previously seen the slow drop/clone and put our version in an `Arc`, so all you see now are clone/drop on `Arc`s.

I should be able to run all benchmarks with version wrapped in `Arc`s. I wonder if that will give me graphs that looked more like yours.

Eh2406:
I added two more small optimizations on top for your review  https://github.com/pubgrub-rs/pubgrub/tree/jf/177-review

konstin:
No difference in performance when cherry-picking the three commits on top:

```
$ taskset -c 0 hyperfine --warmup 1 --runs 3 "./main-uv pip compile scripts/requirements/bio_embeddings.in"  "./branch-uv pip compile scripts/requirements/bio_embeddings.in" 
Benchmark 1: ./main-uv pip compile scripts/requirements/bio_embeddings.in
  Time (mean ± σ):      2.821 s ±  0.007 s    [User: 2.510 s, System: 0.303 s]
  Range (min … max):    2.813 s …  2.827 s    3 runs
 
Benchmark 2: ./branch-uv pip compile scripts/requirements/bio_embeddings.in
  Time (mean ± σ):      2.826 s ±  0.001 s    [User: 2.527 s, System: 0.291 s]
  Range (min … max):    2.825 s …  2.827 s    3 runs
 
Summary
  ./main-uv pip compile scripts/requirements/bio_embeddings.in ran
    1.00 ± 0.00 times faster than ./branch-uv pip compile scripts/requirements/bio_embeddings.in
```

Eh2406:
time|Dev|PR|%
---|----|---|---
sudoku| 44s | 22s | -50%
elm| 274ms | 249.07 ms | -9%
large| 11ms | 9.8821 ms | -7%
slow_135 | 19ms | 6.6884 ms | -63%
zuse | 311 ms | 285.94 ms | -6%
hobo*| 985.2s | 803.6s | -18%
arc_elm| 381.19 ms | 318.25 ms | -16%
arc_large| 15.674 ms | 14.214 ms | -9%
arc_slow_135 | 36.262 ms | 14.885 ms | -59%
arc_zuse | 387.93 ms | 340.78 ms | -12%

PR #179, opened 2024-01-12
feat: contains_many and simplify works with iter of V not just &V

Eh2406:
`contains` takes a `&V`, because if you haven't owned `V` (even behind the smart pointer `Box`/`Arc`/`Cow`) it is easy to add a `&`. So `contains_many` copied that interface and takes an iterator of `Item = &V`. Unfortunately, if you have an iterator of `Item = V` (or `Item =` `Box`/`Arc`/`Cow` of `V`) it is not straightforward to get a reference.

This came up in the context of [semver-pubgrub](https://github.com/pubgrub-rs/semver-pubgrub), whose implementation of [contains](https://github.com/pubgrub-rs/semver-pubgrub/blob/c42083a380d8280873dbbb6b3a1c9876824c7208/src/lib.rs#L73) needs to ignore the `BuildMetadata` part of the semver version. I will be investigating if there are ways to work around this by expressing `=x.y.z.pre` as `>=x.y.z.pre, <x.y.z.pre.0`. But even if `semver-pubgrub` does not **need** this functionality it doesn't hurt to make `contains_many` more flexible.

### DIFF
--- a/src/range.rs
+++ b/src/range.rs
@@ -441,29 +441,34 @@ impl<V: Ord + Clone> Range<V> {
     ///  - If none of the versions are contained in the original than the range will be simplified to `empty`.
     ///
     /// If versions are not sorted the correctness of this function is not guaranteed.
-    pub fn simplify<'v, I>(&self, versions: I) -> Self
+    pub fn simplify<'s, I, BV>(&self, versions: I) -> Self
     where
-        I: Iterator<Item = &'v V> + 'v,
-        V: 'v,
+        I: Iterator<Item = BV> + 's,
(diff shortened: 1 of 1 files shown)


zanieb (approved):
Looks good to me. Maybe @konsti would be interested too.

konstin (approved):

Eh2406:
I rebased which required some intervention due to #183. So while I was looking at it I noticed that we had not made the symmetric change for simplify. So I added that as well. I will give some time for re-review due to the substantive changes.

zanieb (approved):

PR #180, opened 2024-01-12
feat!: ues u32 insted of NumberVersion

Eh2406:
We no longer use a custom `Version` trait, and when we get rid of it there's nothing special about `NumberVersion`. `u32` works exactly the same. So let's get rid of a bunch of boilerplate.

Also apparently our documentation has not been updated to talk about VS, I should open an issue for that, or just fix it. For now it is a TODO.


src/lib.rs
```diff
@@ -19,6 +19,7 @@
 //! [Clone] + [Eq] + [Hash] + [Debug] + [Display](std::fmt::Display).
 //! So things like [String] will work out of the box.
 //!
+//! TODO! This is all wrong. Need to talk about VS, not Version.
```
zanieb, line 22: Can we open an issue for this?

zanieb (approved):

src/lib.rs
Eh2406: Done #185

Eh2406:
oops... this is blocked on me. I will get back to this!

PR #182, opened 2024-01-22
impl: centralize use of hash sets/maps

BurntSushi:
Previously, pubgrub used an alias for HashMap so that everything used
the same implementation. But it didn't do the same for HashSet, and
indeed, there were some uses of FxHashSet and std::collections::HashSet.
Like for HashMap, we standard on the use of FxHashSet for everything.

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -3,7 +3,6 @@
 //! Core model and functions
 //! to write a functional PubGrub algorithm.
 
-use std::collections::HashSet as Set;
 use std::error::Error;
 
 use crate::error::PubGrubError;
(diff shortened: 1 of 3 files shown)


Eh2406:
Do we ever iterate over those Sets? I don't think this is the source of the bugs you are trying to track down. Unless I missed something.

On the other hand, standardizing doesn't seem to hurt anything.

Eh2406 (approved):
I am happy to merge if one of the other maintainers agrees.

BurntSushi:
Right, I have no idea whether these are the source of the non-determinism we're seeing. But as you say, I figured it'd be good to use the same hasher consistently everywhere (unless of course there's a good reason to do otherwise).

Eh2406:
Yes. On second thought I'm convinced. The fact that `Set` means different things in different modules is confusing.

weihanglo:
Just a drive-by: [`clippy::disallowed-types`](https://rust-lang.github.io/rust-clippy/master/index.html#/disallowed_types) can be configured to avoid misuse of `HashSet`/`HashMap` again.

zanieb (approved):
Makes sense to me

PR #183, opened 2024-01-29
test: check for unsorted iteration

Eh2406:
I had a very difficult day debugging why `contains_many` did not match `contains`, turns out my input was not correctly sorted. Similarly, when debugging problems with simplify it would have been nice to easily reject "incorrectly sorted input". (Although that did not turn out to be relevant to the underlying issue.)

This should hopefully be an easy step in the right direction.


zanieb (approved):
Makes sense to me!

Eh2406:
Rebased and `cargo fmt`.

PR #184, opened 2024-01-30
feat!: no recursion and Rc for causes

Eh2406:
The existing code for generating a derivation tree had two performance pathologies.

Because derivations have reused subtrees (i.e. shared IDs) they are really more like DAGs then proper trees. Nonetheless the current API ignores the sharing and built a separate (`Box`) for each of these repeated sections of the output. By using shared ownership (`Rc`) for the shared IDs we can reduce the generation time and memory use for these complicated cases.

Because we generated derivation trees recursively, the stack memory usage was `O(tree_depth)`. When dealing with complicated test cases that use larger `V`s, it is quite possible to overflow the stack. There are probably ways to reduce the constant factors to make this problem even harder to come across, and I will need to investigate them if this PR is not accepted. But this PR solved the problem by simply removing the recursion.

This is a breaking change because the switch from `Box` to `Rc` is visible in our public API.

This is split up into several commits, each of one of which should be reviewable on its own.

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -4,7 +4,7 @@
 //! to write a functional PubGrub algorithm.
 
 use std::error::Error;
-use std::rc::Rc;
+use std::sync::Arc;
 
 use crate::error::PubGrubError;
 use crate::internal::arena::Arena;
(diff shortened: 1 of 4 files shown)


zanieb (approved):
This makes sense to me, although I would defer to @burntsushi if you need a more discerning eye.

BurntSushi (commented):
Nice! I don't have a ton of context on this library's internals and how everything works, but one question that comes to mind is: could you use `Arc` instead of `Rc` here? `Rc` will make the derivation tree non-`Send` which can be a pretty annoying constraint.

zanieb:
`Send` may actually may be important to use since we work in an async context

Eh2406:
Moved from Rc -> Arc. I doubt the perf will ever matter, and it helps you.

PR #186, opened 2024-03-11
Add `Range::as_singleton`

konstin:
We use this method for better error messages:

https://github.com/astral-sh/uv/blob/8d721830db8ad75b8b7ef38edc0e346696c52e3d/crates/uv-resolver/src/pubgrub/report.rs#L92-L96

https://github.com/astral-sh/uv/blob/8d721830db8ad75b8b7ef38edc0e346696c52e3d/crates/uv-resolver/src/pubgrub/report.rs#L509-L520

### DIFF
--- a/src/range.rs
+++ b/src/range.rs
@@ -186,6 +186,21 @@ impl<V: Clone> Range<V> {
 }
 
 impl<V: Ord> Range<V> {
+    /// If the range includes a single version, return it.
+    /// Otherwise, returns [None].
+    pub fn as_singleton(&self) -> Option<&V> {
+        match self.segments.as_slice() {
+            [(Included(v1), Included(v2))] => {
+                if v1 == v2 {
+                    Some(v1)
+                } else {
+                    None
+                }
+            }
+            _ => None,
+        }
+    }
+
     /// Convert to something that can be used with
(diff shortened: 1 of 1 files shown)


src/range.rs
```diff
@@ -186,6 +186,21 @@ impl<V: Clone> Range<V> {
...
+    /// Otherwise, returns [None].
+    pub fn as_singleton(&self) -> Option<&V> {
+        match self.segments.as_slice() {
+            [(Included(v1), Included(v2))] => {
+                if v1 == v2 {
+                    Some(v1)
+                } else {
+                    None
+                }
+            }^
```
Eh2406: You have a added `^` here.

Eh2406 (approved):
I have wanted this from time to time! Seems like a sensible thing to merge.

The commit name will need to match conventional commits (or there will need to be more than one commit and the PR name matches conventional commits) for CI to merge. When that's taking care of +1.

src/range.rs
konstin: oops thx

konstin:
> The commit name will need to match conventional commits (or there will need to be more than one commit and the PR name matches conventional commits) for CI to merge. When that's taking care of +1.

I'll fixup the PRs

PR #187, opened 2024-03-11
Add a `Range::iter()` method

konstin:
Otherwise, it's not possible to implement custom formatting of `Range` types. This seems generally useful to expose.

Example usages:

https://github.com/astral-sh/uv/blob/8d721830db8ad75b8b7ef38edc0e346696c52e3d/crates/uv-resolver/src/pubgrub/report.rs#L97-L112

https://github.com/astral-sh/uv/blob/8d721830db8ad75b8b7ef38edc0e346696c52e3d/crates/uv-resolver/src/pubgrub/report.rs#L549-L560

https://github.com/astral-sh/uv/blob/8d721830db8ad75b8b7ef38edc0e346696c52e3d/crates/uv-resolver/src/pubgrub/report.rs#L568-L605

Upstream port of https://github.com/astral-sh/pubgrub/pull/18

### DIFF
--- a/src/range.rs
+++ b/src/range.rs
@@ -493,6 +493,11 @@ impl<V: Ord + Clone> Range<V> {
         }
         Self { segments }.check_invariants()
     }
+
+    /// Iterate over the parts of the range.
+    pub fn iter(&self) -> impl Iterator<Item = (&Bound<V>, &Bound<V>)> {
+        self.segments.iter().map(|(start, end)| (start, end))
+    }
 }
 
 impl<T: Debug + Display + Clone + Eq + Ord> VersionSet for Range<T> {


src/range.rs
```diff
@@ -493,6 +493,11 @@ impl<V: Ord + Clone> Range<V> {
         }
         Self { segments }.check_invariants()
     }
+
+    /// Iterate over the parts of the range.
+    pub fn iter(&self) -> impl Iterator<Item = &(Bound<V>, Bound<V>)> {
```
Eh2406: Can this be `Item = (&Bound<V>, &Bound<V>)`, so that our API does not require us to store this in a tuple?

Eh2406 (approved):
One nit otherwise I agree that this looks useful. It a little bit locks us into this representation, but the Range Trait can always add another implementation if we find a better one.

src/range.rs
konstin: I've checked with uv, we can :+1:

PR #188, opened 2024-03-11
Add `Range::is_empty()` and `Range::bounds()`

konstin:
We use `Range::is_empty()` for special casing the empty version range in error messages:

https://github.com/astral-sh/uv/blob/8d721830db8ad75b8b7ef38edc0e346696c52e3d/crates/uv-resolver/src/pubgrub/report.rs#L553

https://github.com/astral-sh/uv/blob/8d721830db8ad75b8b7ef38edc0e346696c52e3d/crates/uv-resolver/src/pubgrub/report.rs#L565

We use `Range::bounds()` to check for prerelease bounds and hint the user at our prerelease rules:

https://github.com/astral-sh/uv/blob/8d721830db8ad75b8b7ef38edc0e346696c52e3d/crates/uv-resolver/src/pubgrub/report.rs#L360

Upstream port of https://github.com/astral-sh/pubgrub/pull/16

### DIFF
--- a/src/range.rs
+++ b/src/range.rs
@@ -119,6 +119,29 @@ impl<V> Range<V> {
             segments: SmallVec::one((Included(v1.into()), Excluded(v2.into()))),
         }
     }
+
+    /// Whether the set is empty, i.e. it has not ranges
+    pub fn is_empty(&self) -> bool {
+        self.segments.is_empty()
+    }
+
+    /// Return all boundary versions of this range.
+    pub fn bounds(&self) -> impl Iterator<Item = &V> {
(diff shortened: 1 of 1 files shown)


Eh2406:
`Range::bounds` can be built on `Range::iter`, and it seems like a somewhat niche use case.

`Range::is_empty` is `== Range::empty()`, but might be clearer.

konstin:
Good point, i've split off `Range::is_empty()` (https://github.com/pubgrub-rs/pubgrub/pull/189) and removed our `Range::bounds()` usage (https://github.com/astral-sh/uv/pull/2363)

Closing in favor of #189

PR #189, opened 2024-03-11
Add `Range::is_empty()`

konstin:
We use `Range::is_empty()` for special casing the empty version range in error messages. This mirrors std, which has `is_empty` methods on all of its collections: https://doc.rust-lang.org/nightly/std/?search=is_empty

Example usages:

https://github.com/astral-sh/uv/blob/8d721830db8ad75b8b7ef38edc0e346696c52e3d/crates/uv-resolver/src/pubgrub/report.rs#L553

https://github.com/astral-sh/uv/blob/8d721830db8ad75b8b7ef38edc0e346696c52e3d/crates/uv-resolver/src/pubgrub/report.rs#L565

### DIFF
--- a/src/range.rs
+++ b/src/range.rs
@@ -119,6 +119,11 @@ impl<V> Range<V> {
             segments: SmallVec::one((Included(v1.into()), Excluded(v2.into()))),
         }
     }
+
+    /// Whether the set is empty, i.e. it has not ranges
+    pub fn is_empty(&self) -> bool {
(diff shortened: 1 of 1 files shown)


Eh2406:
For consistency with STD +1. The real question here is should RangeSet treat also have a `is_empty` or should it continue to use `== VS::Empty()`

Eh2406 (approved):
Looks good. We can always add it to the Trate later if we decide to do so.

PR #190, opened 2024-03-12
feat!: associated types instead of generics for P and VS

Eh2406:
Dependency provider was inconsistent about its types, P and VS were generics but Err and Priority were associated types. If/as we had more configuration the benefits of associated types become even more valuable. This PR goes to the breaking change of making everything consistently associated types.

This was in preparation for adding user controlled metadata for `UnavailableDependencies`. But this was sufficiently complicated to deserve its own PR before we get into the additional complications of that extension.

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -3,55 +3,57 @@
 //! Core model and functions
 //! to write a functional PubGrub algorithm.
 
-use std::error::Error;
+use std::collections::HashSet as Set;
 use std::sync::Arc;
 
 use crate::error::PubGrubError;
 use crate::internal::arena::Arena;
-use crate::internal::incompatibility::{IncompId, Incompatibility, Relation};
+use crate::internal::incompatibility::{Incompatibility, Relation};
 use crate::internal::partial_solution::SatisfierSearch::{
     DifferentDecisionLevels, SameDecisionLevels,
 };
 use crate::internal::partial_solution::{DecisionLevel, PartialSolution};
 use crate::internal::small_vec::SmallVec;
-use crate::package::Package;
 use crate::report::DerivationTree;
-use crate::type_aliases::{DependencyConstraints, Map, Set};
+use crate::solver::DependencyProvider;
+use crate::type_aliases::{DependencyConstraints, IncompDpId, Map, V};
 use crate::version_set::VersionSet;
 
 /// Current state of the PubGrub algorithm.
 #[derive(Clone)]
-pub struct State<P: Package, VS: VersionSet, Priority: Ord + Clone> {
-    root_package: P,
-    root_version: VS::V,
+pub struct State<DP: DependencyProvider> {
+    root_package: DP::P,
+    root_version: V<DP>,
 
-    incompatibilities: Map<P, Vec<IncompId<P, VS>>>,
+    #[allow(clippy::type_complexity)]
+    incompatibilities: Map<DP::P, Vec<IncompDpId<DP>>>,
 
     /// Store the ids of incompatibilities that are already contradicted.
     /// For each one keep track of the decision level when it was found to be contradicted.
     /// These will stay contradicted until we have backtracked beyond its associated decision level.
-    contradicted_incompatibilities: Map<IncompId<P, VS>, DecisionLevel>,
+    contradicted_incompatibilities: Map<IncompDpId<DP>, DecisionLevel>,
 
     /// All incompatibilities expressing dependencies,
     /// with common dependents merged.
-    merged_dependencies: Map<(P, P), SmallVec<IncompId<P, VS>>>,
+    #[allow(clippy::type_complexity)]
+    merged_dependencies: Map<(DP::P, DP::P), SmallVec<IncompDpId<DP>>>,
 
     /// Partial solution.
     /// TODO: remove pub.
-    pub partial_solution: PartialSolution<P, VS, Priority>,
+    pub partial_solution: PartialSolution<DP>,
 
     /// The store is the reference storage for all incompatibilities.
-    pub incompatibility_store: Arena<Incompatibility<P, VS>>,
+    pub incompatibility_store: Arena<Incompatibility<DP::P, DP::VS>>,
 
     /// This is a stack of work to be done in `unit_propagation`.
     /// It can definitely be a local variable to that method, but
     /// this way we can reuse the same allocation for better performance.
-    unit_propagation_buffer: SmallVec<P>,
+    unit_propagation_buffer: SmallVec<DP::P>,
 }
 
-impl<P: Package, VS: VersionSet, Priority: Ord + Clone> State<P, VS, Priority> {
+impl<DP: DependencyProvider> State<DP> {
     /// Initialization of PubGrub state.
-    pub fn init(root_package: P, root_version: VS::V) -> Self {
+    pub fn init(root_package: DP::P, root_version: V<DP>) -> Self {
         let mut incompatibility_store = Arena::new();
         let not_root_id = incompatibility_store.alloc(Incompatibility::not_root(
             root_package.clone(),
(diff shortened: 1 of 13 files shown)


src/solver.rs
```diff
@@ -246,20 +251,24 @@ pub trait DependencyProvider<P: Package, VS: VersionSet> {
     /// The kind of error returned from these methods.
     ///
     /// Returning this signals that resolution should fail with this error.
-    type Err: Error;
+    type Err: Error + 'static;
```
mpizenberg, line 251: What are the implications of the static here?

src/type_aliases.rs
```diff
@@ -2,19 +2,29 @@
...
+};
+
 /// Map implementation used by the library.
 pub type Map<K, V> = rustc_hash::FxHashMap<K, V>;
 
 /// Set implementation used by the library.
 pub type Set<V> = rustc_hash::FxHashSet<V>;
 
+/// The version type associated with the version set associated with the dependency provider.
+pub type V<DP> = <<DP as DependencyProvider>::VS as VersionSet>::V;
```
mpizenberg: That alias makes my brain hurt ahah. Do we need the `<DP>` when the V is defined within a dependency provider?

mpizenberg (commented):
Sounds like a good idea. Couple things surprised me a bit.

src/type_aliases.rs
zanieb: Oof. I also found this a bit awkward but don't know if I have an alternative. I recall that putting `V` in the `DependencyProvider` is complicated too?

src/internal/core.rs
```diff
@@ -3,55 +3,57 @@
...
 
     /// This is a stack of work to be done in `unit_propagation`.
     /// It can definitely be a local variable to that method, but
     /// this way we can reuse the same allocation for better performance.
-    unit_propagation_buffer: SmallVec<P>,
+    unit_propagation_buffer: SmallVec<DP::P>,
 }
 
-impl<P: Package, VS: VersionSet, Priority: Ord + Clone> State<P, VS, Priority> {
+impl<DP: DependencyProvider> State<DP> {
```
konstin, line 54: We currently don't have a `DependencyProvider`, but we do have a `State<PubGrubPackage, Range<Version>, PubGrubPriority>` (https://github.com/astral-sh/uv/blob/3799862f5de388c49fe7a4a1417213861aff7746/crates/uv-resolver/src/resolver/mod.rs#L242)

src/type_aliases.rs
Eh2406: I had a commit where `V` was a associated type on `DP`. It made the use within the library a lot easier. But it required users to confirm that `type V = VS::V`, which ended up being ugly and tricky. I unfortunately drop that commit when rebasing. I will attempt to reconstruct it so we can decide which poison is worse.

src/solver.rs
Eh2406: That the error type must be able to outlive its dependency provider. It is less onerous than it might first appear. Nonetheless, I was surprised that I needed to add it. It's required by the `derive(Error)` generated code. I have not found a good way to look at the generated code to understand why it's needed. But I will give it another try.

src/internal/core.rs
Eh2406: I see two ugly ways to fix this. But I'm open to other suggestions:
- `pubgrub-rs` eats the ugliness, and we only use `DP` at the surface level. `State` and below have a large list of generic types, each of which is bound to match their `DP` associated type.
- `uv` eats the ugliness, you have some type that implements `DependencyProvider` with `unimplemented!()` for all required methods. As such, `State` can retrieve the associated types and no one needs to instantiate an instance.

I guess there's another option:
- we stabilize a `AsincDependencyProvider` and you switch to using our stable interface.

Thoughts?

mpizenberg: If uv's solver is heavily modified to fit the async needs, I wouldn't try to merge uv and pubgrub-rs solvers until more work is done on async in pubgrub-rs. Especially while uv is still in early days and more divergence may occur. Is this a priority of yours @konstin to re-depend on pubgrub-rs asap?

konstin: I like not using a dependency provider, running the main loop ourselves makes the code straightforward (no callbacks) and we can hook in at many places.

mpizenberg: That's good to know!

Eh2406: @konstin:
- Just to confirm my understanding `State` is not part of our public API. Is that correct? If we accidentally made it public I would very much like to know.
- Similarly, how painful would it be for you to make a fake `DependencyProvider` in `uv` for `State` to read its associated types out of? It would not have to implement any of the methods.

src/type_aliases.rs
Eh2406: `git reflog` did its magic and I was able to find the old commit! And looking it over maybe it's not so bad. What you all think?

src/internal/core.rs
konstin: Yes, we made `State` public (https://github.com/pubgrub-rs/pubgrub/compare/dev...astral-sh:pubgrub:main#diff-cbd16a2b1fdb7e2f39b676831dfaee5250000b403a9f1fcfcbfede424d7cbf73L78-R79)

If it's just for the typings it's fine to implement a `DependencyProvider`.

src/solver.rs
Eh2406: I used https://github.com/dtolnay/cargo-expand which worked really well. It would appear that `thiserror` decided that `Err` should be the [`source`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source) of a `ErrorRetrievingDependencies`. Since `source` returns `dyn Error + 'static` we have to be `'static`. 

Looks like [`thiserror`](https://docs.rs/thiserror/latest/thiserror/) has special handling for fields named `source`. So we can:
- Remove the ` + 'static` and rename the field.
- Leave the ` + 'static` so that we can mark more things as `source`s.

I will add a commit doing the latter, but could go either way.

src/internal/core.rs
konstin: That's probably a discussion to be had separate from this issue, but how hard would it be to make state or a state wrapper public, to officially support using pubgrub from your own solver loop instead of using pubgrub's? As in, making enough public that you can implement `resolve` in user code?

mpizenberg: From my point of view, I like having package APIs that offer different ease-of-use/control tradeoffs. First level is friendliest and easiest to use but don't offer much control. Up to now we only had 2 levels, first level is dependency provider already done, second level is you customize the dep provider. It seemed the right balance without sacrificing correctness. In your case, you need an additional level of control, where even more correctness is in the hand of the user. Anyway just my 2 cents. It might be worth exploring.

Eh2406:
@zanieb and @mpizenberg what are you thinking about the resent pushes?

mpizenberg:
Recent commits only add some docs right? or am I missing something. I haven't looked at it again since last week when commenting.

Eh2406:
No.
- [mabe V as well?](https://github.com/pubgrub-rs/pubgrub/pull/190/commits/58f786be15cd4b2cf583b4ba43a4e9b202c087bb) adds an associated type for `V` on `DP`.
- [more source](https://github.com/pubgrub-rs/pubgrub/pull/190/commits/91a7c9fbd2f651471cda09b852a86dd500ea2947) uses the new ` + 'static` by marking more errors as sources.

zanieb:
They seem reasonable to me but frankly generics in Rust APIs are not something I have a lot of experience with yet. I'd be curious to see the metadata pull request stacked on top of this one, I can try to integrate in that into uv and we can see what it looks like in-use?

mpizenberg:
Ah yes, I like the new V.
As for `#[source]` in errors, honestly I haven't coded in rust for so long that I don't know what happens there ^^.

Eh2406:
If we all like the basic direction this is going, let's merge. There will be more breaking changes before we cut `0.3`, if we find problems or change our minds we can fix it then.

> I'd be curious to see the metadata pull request stacked on top of this one, I can try to integrate in that into uv and we can see what it looks like in-use?

That does seem like the next step, and I'm happy to have you take the first stab at it. Something like add a `type Metadat: Eq + Clone + Debug + Display` to `DependencyProvider` and use it in `Kind` and `External` so that it can be passed in with a `Incompatibility` and read as part of the `DerivationTree`. If that seeming overwhelming, start with something that useful for you and we can generalize from there.

zanieb:
Sounds good to me!

Eh2406:
If someone could officially give a approving review, we can get this merged.

zanieb (approved):

PR #193, opened 2024-03-18
feat: ensure successful round-trip of RON

Eh2406:
cc https://github.com/ron-rs/ron/issues/530

When using the new format our benchmark files could not be read. Apparently this is going to be fixed in the next version of the RON dep. For now use the prerelease version.

### DIFF
--- a/src/range.rs
+++ b/src/range.rs
@@ -862,7 +862,8 @@ pub mod tests {
 
     proptest! {
 
-        // Testing negate ----------------------------------
+        // Testing serde ----------------------------------
+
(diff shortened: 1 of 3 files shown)


src/range.rs
```diff
@@ -863,6 +863,13 @@ pub mod tests {
     proptest! {
 
         // Testing negate ----------------------------------
```
mpizenberg: This "Testing negate" comment section should probably move after the serde round trip test.

Eh2406: Good call.

konstin (approved):

PR #194, opened 2024-03-22
Update priority queue

konstin:
I've been looking at duplicate dependencies in our tree and priority-queue 1 was pulling in old versions of indexmap and hashbrown, so i've updated all dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,18 +21,18 @@ include = ["Cargo.toml", "LICENSE", "README.md", "src/**", "tests/**", "examples
 
 [dependencies]
 indexmap = "2.0.2"
-priority-queue = "1.1.1"
+priority-queue = "2.0.2"
 thiserror = "1.0"
 rustc-hash = "1.1.0"
 serde = { version = "1.0", features = ["derive"], optional = true }
 log = "0.4.14" # for debug logs in tests
 
 [dev-dependencies]
-proptest = "0.10.1"
-ron = "0.6"
+proptest = "1.4.0"
+ron = "0.8.1"
 varisat = "0.2.2"
 criterion = "0.5"
-env_logger = "0.9.0"
+env_logger = "0.11.3"
 
 [[bench]]
 name = "large_case"


Cargo.toml
```diff
@@ -21,18 +21,18 @@ include = ["Cargo.toml", "LICENSE", "README.md", "src/**", "tests/**", "examples
 
 [dependencies]
 indexmap = "2.0.2"
-priority-queue = "1.1.1"
```
Eh2406, line 24: Change log at https://github.com/garro95/priority-queue?tab=readme-ov-file#changes which all seem reasonable.

Cargo.toml
```diff
@@ -21,18 +21,18 @@ include = ["Cargo.toml", "LICENSE", "README.md", "src/**", "tests/**", "examples
...
+priority-queue = "2.0.2"
 thiserror = "1.0"
 rustc-hash = "1.1.0"
 serde = { version = "1.0", features = ["derive"], optional = true }
 log = "0.4.14" # for debug logs in tests
 
 [dev-dependencies]
-proptest = "0.10.1"
```
Eh2406, line 31: Change log at https://github.com/proptest-rs/proptest/blob/master/proptest/CHANGELOG.md#100 which all seem reasonable.

Cargo.toml
```diff
@@ -21,18 +21,18 @@ include = ["Cargo.toml", "LICENSE", "README.md", "src/**", "tests/**", "examples
...
-ron = "0.6"
```
Eh2406, line 32: Change log at https://github.com/ron-rs/ron/blob/master/CHANGELOG.md#060---2020-05-21 which all seems reasonable. Although we also could use a bug fix from prerelease of `0.9`, but I have a separate PR for that.

Cargo.toml
```diff
@@ -21,18 +21,18 @@ include = ["Cargo.toml", "LICENSE", "README.md", "src/**", "tests/**", "examples
...
+proptest = "1.4.0"
+ron = "0.8.1"
 varisat = "0.2.2"
 criterion = "0.5"
-env_logger = "0.9.0"
```
Eh2406, line 35: Change log at https://github.com/rust-cli/env_logger/blob/main/CHANGELOG.md#090----2022-07-14 which I don't entirely understand but seemed reasonable.

Eh2406 (approved):

Eh2406:
The commit message needs to follow conventional commits in order to successfully merge.

konstin:
Sorry, fixed

konstin:
Is there a reason there is no "chore" label for commits?

Eh2406:
Not a good one that I understand. If you would like to add it to the CI go ahead and do so.

PR #196, opened 2024-03-22
Add dependabot

konstin:
Weekly updates for dependabot and cargo dependencies as PRs. This is the same configuration we are using in uv minus the automatic label: Feel to edit the interval if you prefer something else.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      artifacts:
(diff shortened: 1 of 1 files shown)


Eh2406 (approved):
We can give this a try and see how it goes. We may need to adjust grouping or customize the commit messages, but we can figure that out as the PR's come in.

zanieb:
There's currently no grouping which is more work but gives you discrete commits if there's a regression. I prefer more work merging but am not opposed to grouping here :)

Eh2406:
Looks like our CI want `type must be one of [build, ci, docs, feat, fix, perf, refactor, release, revert, style, test] [type-enum]` not `chore`. :-( And same on the other PRs.

PR #203, opened 2024-03-25
 ci: remove caching steps 

### LINKED ISSUE #201: Switch to `swatinem/rust-cache` instead of bespoke `actions/cache` in CI
https://github.com/Swatinem/rust-cache has some robust logic for caching Rust projects and I've never had problems with it. Might be worth switching over.

konstin:
Pubgrub is so fast to compile that caching in CI isn't required, so i removed it. Closes #201.

I've added `workflow_dispatch` to dry run these workflows.

The timing below are from my machine, but even if CI is an order of magnitude slower i don't think we should cache:

 ```
$ hyperfine --prepare "cargo clean" "cargo publish --dry-run" "cargo +nightly doc --no-deps"
 Benchmark 1: cargo publish --dry-run
   Time (mean ± σ):      1.661 s ±  0.272 s    [User: 2.240 s, System: 0.324 s]
   Range (min … max):    1.467 s …  2.207 s    10 runs

 Benchmark 2: cargo +nightly doc --no-deps
   Time (mean ± σ):      1.230 s ±  0.020 s    [User: 2.013 s, System: 0.312 s]
   Range (min … max):    1.202 s …  1.263 s    10 runs
```

### DIFF
--- a/.github/workflows/cargo_publish_dry_run.yml
+++ b/.github/workflows/cargo_publish_dry_run.yml
@@ -4,6 +4,7 @@ name: Check crate publishing works
 on:
   pull_request:
     branches: [ release ]
+  workflow_dispatch:
 
 env:
   CARGO_TERM_COLOR: always
@@ -15,10 +16,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install stable Rust
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          profile: minimal
+        uses: dtolnay/rust-toolchain
 
       - name: Get Cargo version
         id: cargo_version
--- a/.github/workflows/cargo_publish_dry_run.yml
+++ b/.github/workflows/cargo_publish_dry_run.yml
@@ -25,16 +25,5 @@ jobs:
         run: echo "::set-output name=version::$(cargo -V | tr -d ' ')"
         shell: bash
 
-      - name: Download cache
-        uses: actions/cache@v4
-        with:
-          path: |
-            ~/.cargo/registry/index/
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: .github/workflows/cargo_publish_dry_run.yml [hunk 1]
```diff
@@ -4,6 +4,7 @@ name: Check crate publishing works
 on:
   pull_request:
     branches: [ release ]
+  workflow_dispatch:
```
zanieb, line 7: Why this change?

.github/workflows/cargo_publish_dry_run.yml
```diff
@@ -15,26 +16,12 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install stable Rust
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          profile: minimal
+        uses: dtolnay/rust-toolchain
```
zanieb, line 19: Why this change?

Eh2406: cc https://github.com/pubgrub-rs/pubgrub/issues/164

konstin: This action is used everywhere else, this was the only exception, and action-rs is unmaintained.

zanieb (approved):

mpizenberg (approved):
Looks good to me, but why the lock changes in this PR? It seems unrelated.

konstin:
> Looks good to me, but why the lock changes in this PR? It seems unrelated.

Not sure what was out of sync, but solved by rebasing on dev

PR #208, opened 2024-04-15
Generic reason for custom incompatibility

konstin:
Restructure the incompatibility into pubgrub internal metadata and a custom external type.

The internal types are:
* not root: The root incompatibility that drives resolution
* no versions: We tried all versions and there is none remaining
* from dependency of: {A, not B} from A -> B
* derived from: {A, not C} from {A, not B} and {B, not C}
* custom: Any incompatibility that originates outside pubgrub. The idea is that users can set `M` to a custom enum of theirs. Examples:
  * The version would require building the package, but builds are disabled.
  * The package is not available in the cache, but internet access has been disabled.
  * The package has invalid metadata

I'm currently working on adapting the formatting to adapt to the uv of this. This should reduce our diff significantly.

This refactoring is relevant to collapsing custom incompatibilities, but i haven't looked into this properly yet.

### DIFF
--- a/src/internal/incompatibility.rs
+++ b/src/internal/incompatibility.rs
@@ -3,7 +3,7 @@
 //! An incompatibility is a set of terms for different packages
 //! that should never be satisfied all together.
 
-use std::fmt;
+use std::fmt::{self, Debug, Display};
 use std::sync::Arc;
 
 use crate::internal::arena::{Arena, Id};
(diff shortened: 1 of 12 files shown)


src/internal/incompatibility.rs
```diff
@@ -32,26 +32,44 @@ use crate::version_set::VersionSet;
...
-    DerivedFrom(IncompId<P, VS>, IncompId<P, VS>),
+    ///
+    /// For example, if a -> b and b -> c, we can derive a -> c.
+    DerivedFrom(IncompId<P, VS, M>, IncompId<P, VS, M>),
+    /// The package is unavailable for reasons outside pubgrub.
+    ///
+    /// Examples:
+    /// * The version would require building the package, but builds are disabled.
+    /// * The package is not available in the cache, but internet access has been disabled.
+    Custom(P, VS, M),
```
zanieb, line 72: Should we just call this `Unavailable` or `Unusable`? We talked about this a little back in https://github.com/pubgrub-rs/pubgrub/pull/153

konstin: I'm trying to name this something that says it's outside pubgrub, unavailable/unusable sounds too much like it could come from the resolution itself

zanieb: I wonder if `External` is overloaded.

konstin: There's the `External` type in derivation tree. It's more non-derived than external, but it would be an invasive change

zanieb: Yeah I agree that's my concern about it being overloaded 🤷‍♀️ I find it confusing there honestly but curious to see what Jacob thinks of all this.

Eh2406:
Sorry for the delay. I just had a chance to look at the diff, and I absolutely love it. Thank you so much! I need to poke at this in an editor before I write up the few nits I have.

src/internal/incompatibility.rs
Eh2406: I came in expecting to want `Unusable`, but the discussion has made me less certain. @mpizenberg is the one with a knack for naming things.

src/internal/incompatibility.rs
```diff
@@ -155,6 +184,11 @@ impl<P: Package, VS: VersionSet> Incompatibility<P, VS> {
         if dep_term != other.get(p2) {
             return None;
         }
+        // The metadata must be identical to merge
+        let self_metadata = self.metadata();
+        if self_metadata != other.metadata() {
+            return None;
+        }
```
Eh2406: I don't think this code path is needed at the moment, because we can only get here if `as_dependency` is `Some` and dependencies can't hold metadata. But perhaps they should be able to hold metadata, like what file and line number this dependency edge came from.

Thoughts?

src/internal/incompatibility.rs
mpizenberg: I generally like this change. I will have a bit of time to review code/naming after May (I have a conf talk next week and changing job end of this month). Otherwise consider I'm ok with it and might (or not) suggest a renaming PR later.

src/internal/incompatibility.rs
```diff
@@ -96,14 +113,26 @@ impl<P: Package, VS: VersionSet> Incompatibility<P, VS> {
         }
     }
 
-    /// Create an incompatibility to remember
-    /// that a package version is not selectable
-    /// because its list of dependencies is unavailable.
-    pub fn unavailable_dependencies(package: P, version: VS::V) -> Self {
+    /// Create an incompatibility for a reason outside pubgrub.
+    #[allow(dead_code)] // Used by uv
```
konstin, line 117: We can also drop this here and only use it in our fork

src/internal/incompatibility.rs
konstin: Removed it unless @zanieb wants to keep this

konstin:
Should we create a trait for `M: Eq + Clone + Debug + Display`?

src/internal/incompatibility.rs
konstin: We could merge away external and use that name here instead: https://github.com/pubgrub-rs/pubgrub/compare/dev...astral-sh:pubgrub:konsti/dev/merge-external

src/internal/incompatibility.rs
zanieb: We actually were just talking about needing that... but it can wait :D

Eh2406:
Yes. I think we should have a trait. Although it could be in a follow-up PR. It should, one day, have a method `can_merge(&Self, &Self) -> Option<Self>` instead of the `Eq` bound. `None` would mark that the two incompatibilities should not be merged. And `Some(M)` would provide the new metadata for the merged incompatibility. That would allow a error like "unavailable because of HTTP 404, with request ID 123456789" to be merged with "unavailable because of HTTP 404, with request ID 987654321" even though they are not structurally equal.

Eh2406 (approved):
I poked at this in an editor, and `rustc` told me why everything was there. 
Adding a trait and renaming the type can be done in follow-up PR's.

PR #214, opened 2024-05-02
Move formatting of explanations into the report formatter 

konstin:
Move the explanation formatting from `DefaultStringReporter` to the `ReportFormatter` trait, so they can be overridden downstream (https://github.com/astral-sh/uv/blob/20e9589662396ca655e253c48afe7c60e05a6590/crates/uv-resolver/src/pubgrub/report.rs#L33). Upstreamed from https://github.com/astral-sh/pubgrub/pull/19.

This resolves a large chunk of our diff with upstream (dev to uv main is 1278 lines, this PR to uv main is 743, measured as `git diff <branch>..konsti/main-rebase-generic-reason | wc -l`.

### DIFF
--- a/examples/unsat_root_message_no_version.rs
+++ b/examples/unsat_root_message_no_version.rs
@@ -2,7 +2,7 @@
 
 use pubgrub::error::PubGrubError;
 use pubgrub::range::Range;
-use pubgrub::report::Reporter;
+use pubgrub::report::{Derived, Reporter};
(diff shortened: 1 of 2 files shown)


Eh2406 (approved):
Seems reasonable, but I will leave final merge to @zanieb who has the deepest understanding of this code.

zanieb (approved):
Yeah I think these makes a lot of sense as part of the formatter.

mpizenberg:
Thanks for backporting that work from uv!

PR #216, opened 2024-05-02
Rename dependencies type to available and unavailable

konstin:
In uv, dependencies are either available or unavailable. They are not unknown, but rather missing due to some brokenness: We're offline but the dep is not cached, the version list failed to deserialize, etc. (https://github.com/astral-sh/uv/blob/0b84eb01408eb0e90b5720b027359aac10708665/crates/uv-resolver/src/resolver/mod.rs#L945-L996). This change is a rename of the variants of `Dependencies` to reflect that, upstreaming the change from uv.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -121,7 +121,6 @@
 //!
 //! The third method [get_dependencies](crate::solver::DependencyProvider::get_dependencies)
 //! aims at retrieving the dependencies of a given package at a given version.
-//! Returns [None] if dependencies are unknown.
 //!
 //! In a real scenario, these two methods may involve reading the file system
 //! or doing network request, so you may want to hold a cache in your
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -121,7 +121,6 @@
 //!
 //! The third method [get_dependencies](crate::solver::DependencyProvider::get_dependencies)
 //! aims at retrieving the dependencies of a given package at a given version.
-//! Returns [None] if dependencies are unknown.
```
konstin, line 124: It returns `Result<Dependencies<DP::P, DP::VS, DP::M>, DP::Err>`, not an option.

Eh2406:
Seams good to me but @mpizenberg is the naming wizard.

mpizenberg (approved):
Thanks for the propose change @konstin
I also like it

PR #219, opened 2024-05-07
Add `DerivationTree.packages() -> HashSet<&P>`

konstin:
In uv, we check the available versions to determine if there a prerelease versions and show an appropriate hint if required, mainly collecting package versions [here](https://github.com/astral-sh/uv/blob/94cf60457466d1234069c52b19e1bab545129c99/crates/uv-resolver/src/error.rs#L238) to display a hint [here](https://github.com/astral-sh/uv/blob/20e9589662396ca655e253c48afe7c60e05a6590/crates/uv-resolver/src/pubgrub/report.rs#L424-L448). 

This is the most efficient way of writing this function, but the derivation tree is small in the number of packages involved.

### DIFF
--- a/src/report.rs
+++ b/src/report.rs
@@ -7,6 +7,8 @@ use std::fmt::{self, Debug, Display};
 use std::ops::Deref;
 use std::sync::Arc;
 
+use rustc_hash::FxHashSet;
+
 use crate::package::Package;
 use crate::term::Term;
 use crate::type_aliases::Map;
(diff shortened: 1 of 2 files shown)


src/report.rs
```diff
@@ -71,6 +73,30 @@ pub struct Derived<P: Package, VS: VersionSet, M: Eq + Clone + Debug + Display>
 }
 
 impl<P: Package, VS: VersionSet, M: Eq + Clone + Debug + Display> DerivationTree<P, VS, M> {
+    /// Get all packages referred to in the derivation tree.
+    pub fn packages(&self) -> FxHashSet<&P> {
```
Eh2406: Can we use the `crate::type_aliases::Set;`?

src/report.rs
```diff
@@ -71,6 +73,30 @@ pub struct Derived<P: Package, VS: VersionSet, M: Eq + Clone + Debug + Display>
...
+                }
+                External::NoVersions(p, _)
+                | External::NotRoot(p, _)
+                | External::Custom(p, _, _) => {
+                    packages.insert(p);
+                }
+            },
+            Self::Derived(derived) => {
+                packages.extend(derived.terms.keys());
+                packages.extend(derived.cause1.packages().iter());
```
Eh2406, line 93: Can we leave ourselves a breadcrumb comment here about how the recursive implementation is simple but what problems it might cause and how we could fix them?

Eh2406 (changes requested):
This seems like a useful thing to have. It should probably have some tests. Separately it seems similar to code in `errors_the_same_with_only_report_dependencies`, perhaps this method could be used there.

konstin:
I added `.packages()` to the existing error test case to avoid duplicating the setup code, let me know if there's a better place for it.

Eh2406 (approved):

PR #227, opened 2024-06-03
Fix a clippy lint: `impl Default for Arena`

konstin:
This showed up as a clippy warning (parameterless new but no default) when removing `#![allow(clippy::all)]` on our `main`, though surprisingly this doesn't happen on upstream `dev`. I think it's a good change either way.


Eh2406 (approved):
Looks good.

PR #238, opened 2024-07-15
Return `NoSolutionError` from unit propagation

konstin:
In uv, the only pubgrub error that can occur is a `NoSolutionError`, and the only place it can occur is `unit_propagation`. By returning `NoSolutionError` instead of `PubGrubError`, we can remove `unreachable!()` calls on uv's side. `NoSolutionError` is a type alias for `DerivationTree`.

Since `unit_propagation` is unreachable publicly, this is not a breaking change.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -7,15 +7,19 @@ use thiserror::Error;
 use crate::report::DerivationTree;
 use crate::solver::DependencyProvider;
 
+/// There is no solution for this set of dependencies.
+pub type NoSolutionError<DP> = DerivationTree<
+    <DP as DependencyProvider>::P,
+    <DP as DependencyProvider>::VS,
+    <DP as DependencyProvider>::M,
+>;
+
 /// Errors that may occur while solving dependencies.
 #[derive(Error)]
-pub enum PubGrubError<DP>
-where
-    DP: DependencyProvider,
-{
+pub enum PubGrubError<DP: DependencyProvider> {
     /// There is no solution for this set of dependencies.
     #[error("No solution")]
-    NoSolution(DerivationTree<DP::P, DP::VS, DP::M>),
+    NoSolution(NoSolutionError<DP>),
 
     /// Error arising when the implementer of
     /// [DependencyProvider]
@@ -62,13 +66,19 @@ where
     Failure(String),
 }
 
+impl<DP: DependencyProvider> From<NoSolutionError<DP>> for PubGrubError<DP> {
+    fn from(err: NoSolutionError<DP>) -> Self {
+        Self::NoSolution(err)
+    }
+}
+
 impl<DP> std::fmt::Debug for PubGrubError<DP>
 where
     DP: DependencyProvider,
 {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            Self::NoSolution(arg0) => f.debug_tuple("NoSolution").field(arg0).finish(),
+            Self::NoSolution(err) => f.debug_tuple("NoSolution").field(&err).finish(),
             Self::ErrorRetrievingDependencies {
                 package,
                 version,
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -6,7 +6,7 @@
 use std::collections::HashSet as Set;
 use std::sync::Arc;
 
-use crate::error::PubGrubError;
+use crate::error::NoSolutionError;
 use crate::internal::arena::Arena;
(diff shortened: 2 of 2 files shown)


src/error.rs
```diff
@@ -7,15 +7,19 @@ use thiserror::Error;
 use crate::report::DerivationTree;
 use crate::solver::DependencyProvider;
 
+/// There is no solution for this set of dependencies.
+pub type NoSolutionError<DP> = DerivationTree<
+    <DP as DependencyProvider>::P,
+    <DP as DependencyProvider>::VS,
+    <DP as DependencyProvider>::M,
+>;
```
konstin, line 15: I'm not really sure about this; should we rename `DerivationTree` to `NoSolutionError`, or should this be a newtype?

src/error.rs
```diff
@@ -62,13 +66,19 @@ where
...
+    fn from(err: NoSolutionError<DP>) -> Self {
+        Self::NoSolution(err)
+    }
+}
+
 impl<DP> std::fmt::Debug for PubGrubError<DP>
 where
     DP: DependencyProvider,
 {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
konstin, line 79: What's the reason for the custom debug impl here?

src/internal/core.rs
```diff
@@ -184,16 +184,14 @@ impl<DP: DependencyProvider> State<DP> {
...
         incompatibility: IncompDpId<DP>,
-    ) -> Result<(DP::P, IncompDpId<DP>), PubGrubError<DP>> {
+    ) -> Result<(DP::P, IncompDpId<DP>), NoSolutionError<DP>> {
         let mut current_incompat_id = incompatibility;
         let mut current_incompat_changed = false;
         loop {
             if self.incompatibility_store[current_incompat_id]
                 .is_terminal(&self.root_package, &self.root_version)
             {
-                return Err(PubGrubError::NoSolution(
```
Eh2406, line 194: Should we go one step further and return only `Err(current_incompat_id)`? This would allow a user to skip the overhead of `build_derivation_tree` if the presence of an error is the only thing they care about.

konstin: I've changed this method, but left `unit_propagation`, so we don't leak the incompatibility id type.

mpizenberg:
Sounds like a good idea to me to have a more granular error type if that let’s you use the unit propagation loop directly without caring about the other types of errors (that I suppose you have your own versions of).

src/internal/core.rs
mpizenberg: > This would allow a user to skip the overhead of build_derivation_tree if the presence of an error is the only thing they care about.

Sounds good too. This would need to make public the `build_derivation_tree` for those in need. That function is in a core private module. Can it be relocated in a public module? I don’t remember if there is any type or function call in there that requires private access.

mpizenberg: Ha, it depends on the core `State` because it requires access to the incompat store, which isn’t public right?

src/error.rs
mpizenberg: Derivation Tree is the technical term used in the pubgrub algorithm description so it’s good that it exists in the code base. An alias like this is fine to me.

src/internal/core.rs
Eh2406: Our public API would not need to change, because all of this is technically private. `uv` is using it anyway, but they can mark `pub` anything they need to. Making this optimization available to users of our public API, would require some significant thought or we could decide that this is only available to teams using their own resolution loop.

mpizenberg: Ah I see, you meant just in the context of uv, which already has access to this.

konstin: For our level of pubgrub usage, we want to call `unit_propagation` ourselves and have a custom report formatter, but we don't want to expose details such as incompatibility ids, so i really like `pub fn unit_propagation(&mut self, package: DP::P) -> Result<(), NoSolutionError<DP>>` as level of abstraction.

Eh2406 (approved):
Looks good. Unless we have a good reason we may as well have API's that are convenient for `uv`. In the abstract, this project needs the right to break internal APIs, even if that makes things a little more awkward for `uv`. So if there is a good reason for 
`.map_err(|id| state.build_derivation_tree(id))` to move up to the resolution loop we will do that then. Similarly if we need a more complicated error type in the future, for example if `unit_propagation` is slow enough that `should_cancel` needs to be moved in we will revert this then. But there's no reason to be pointlessly antagonistic, let's merge what's easiest for `uv`.

PR #240, opened 2024-07-16
Move all public symbols to the top level

konstin:
This is a prototype for moving all symbols to the top level namespace, removing the modules from the public api.

image

### DIFF
--- a/examples/branching_error_reporting.rs
+++ b/examples/branching_error_reporting.rs
@@ -1,10 +1,9 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use pubgrub::error::PubGrubError;
-use pubgrub::range::Range;
-use pubgrub::report::{DefaultStringReporter, Reporter};
-use pubgrub::solver::{resolve, OfflineDependencyProvider};
-use pubgrub::version::SemanticVersion;
+use pubgrub::{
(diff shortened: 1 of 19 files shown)


mpizenberg:
I usually like modularized code, based on data structures. I also like the exercise of modularizing it in ways that make partial reuse possible without over exposing things to the public or without making cyclic dependencies. I also like the fact that usually, the module hierarchy can be used effectively in the naming scheme.

All that being said, in pubgrub case, most data structure names are self descriptive and do not necessarily need the module hierarchy, which is sometimes redundant. Same goes for the exposed functions. There is only resolve, which is also self descriptive. And by using `pub use` in the root module, we also avoid the possible cyclic dependencies that could arise if all was actually in a single file.

So I think all my usual reasons for not having all at the top are addressed, AND it makes it much easier to work with for users of the crate. So I’m very in favor of this :)

Waiting for feedback on potential downsides.

Eh2406 (approved):
I see no downsides. Let's do it!

konstin:
I have a followup change (https://github.com/astral-sh/pubgrub/compare/konsti/pubgrub-top-level-namespace...astral-sh:pubgrub:konsti/pubgrub-cleanup-internal-imports?expand=1), but i made the PR from my fork so i can't stack them in this repo.

Eh2406:
@konstin, Did you want to open a PR for this follow-up work?

PR #244, opened 2024-07-30
feat: allow depending on self

Eh2406:
In general PubGrub allows cyclical imports. It is totally okay for `A` to depend on `B` which depends on `A`. We have a special exception for `A` depending on `A`. This is extra code and kind of inconsistent.

The argument for the error message was to force the user to choose one of the two reasonable semantics:
- Self dependencies are okay, but meaningless. In which case the dependency provider should filter them out.
- Self dependencies are not okay. In which case the dependency provider should return `Unavailable` with a reason.

I believe that the UV fork of this project removed this code to implement the "self dependencies are okay" semantics.

In Cargo we (currently) take a somewhat inconsistent position. We treat the self dependency as okay for dependency resolution, but then error later due to the cyclical import. I would like to maintain bug for bug compatibility as long as possible to avoid tying upgrading to PubGrub to other breaking changes. The easiest way to implement this "it's not a problem for resolution but the edge does exist for cycle checking" is for PubGrub to simply not error. (Although if we decide not to merge this I have other options.)

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -33,19 +33,6 @@ pub enum PubGrubError<DP: DependencyProvider> {
         source: DP::Err,
     },
 
-    /// Error arising when the implementer of
-    /// [DependencyProvider]
-    /// returned a dependency on the requested package.
-    /// This technically means that the package directly depends on itself,
-    /// and is clearly some kind of mistake.
-    #[error("{package} {version} depends on itself")]
-    SelfDependency {
-        /// Package whose dependencies we want.
-        package: DP::P,
-        /// Version of the package for which we want the dependencies.
-        version: DP::V,
-    },
-
     /// Error arising when the implementer of [DependencyProvider] returned an error in the method
     /// [choose_version](DependencyProvider::choose_version).
     #[error("Decision making failed")]
@@ -83,11 +70,6 @@ where
                 .field("package", package)
                 .field("version", version)
                 .field("source", source)
-                .finish(),
-            Self::SelfDependency { package, version } => f
-                .debug_struct("SelfDependency")
-                .field("package", package)
-                .field("version", version)
                 .finish(),
             Self::ErrorChoosingPackageVersion(arg0) => f
                 .debug_tuple("ErrorChoosingPackageVersion")
--- a/src/solver.rs
+++ b/src/solver.rs
@@ -153,12 +153,6 @@ pub fn resolve<DP: DependencyProvider>(
                     ));
                     continue;
                 }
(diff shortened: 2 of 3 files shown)


mpizenberg:
It seems to have been added for version 0.2, back in 2020 (https://github.com/pubgrub-rs/pubgrub/blob/ff09b460c36de7584a05e91b29e869d26ae4b2ff/CHANGELOG.md?plain=1#L54).

I think I added it because cyclical dependencies are forbidden in Elm, and I don’t think its trivial to figure out without the solver signaling them. But your comment about cargo behavior seems to imply it is not too complex

konstin (approved):

konstin:
I believe cyclical dependencies are something that the caller should decide on when processing the metadata (so they can allow it in python and ban it in elm), so not something that pubgrub should check.

konstin (approved):

mpizenberg:
The thing is non-trivial cyclical dependencies are not easy to spot without the dependency solver. And I can see 3 different situations:
1. All cyclical dependencies are allowed
2. Cyclical dependencies are tolerated while solving but just discarded as invalid solutions, so the solver continues
3. Cyclical are strictly forbidden and any discovered cyclic dependency that may occur while building your app and not having published yet your new version but using a dependency that depends on your very package at an earlier (compatible range) version should raise an error to prevent you from going through making a release.

How do you differentiate these 3 behaviors without the help of the solver and only metadata?

mpizenberg:
My bad, the current solver doesn’t check the above 3 anyway. It’s just checked in direct dependencies. Well I guess then it’s an easy check on the provider part to decide.

mpizenberg (approved):

Eh2406:
> In general PubGrub allows cyclical imports. It is totally okay for `A` to depend on `B` which depends on `A`. We have a special exception for `A` depending on `A`. This is extra code and kind of inconsistent.

The only check that PubGrub currently supports if the trivial one.

If you have a structure like
```rust
    fn get_dependencies(
        &self,
        package: &P,
        version: &VS::V,
    ) -> Result<Dependencies<P, VS, Self::M>, NewErr> {
        let mut out = Dependencies::default();
        ...
        Ok(Dependencies::Available(out))
    }
```
then to maintain the current behavior you just have to add
```rust
        if out.contains_key(package) {
           return Err(NewErr::New());
        }
```
or to backtrack on self dependencies you add
```rust
        if out.contains_key(package) {
           return Ok(Dependencies::Unavailable("self dep"));
        }
```

PR #250, opened 2024-08-20
Implement `[Partial]Ord` for `Range`

konstin:
In uv, we want to have a stable ordering of certain types that `Range<Version>`, both for determinism internally and for deterministically ordered output files.

This PR adds `impl<V: PartialOrd> PartialOrd for Range<V>` and `impl<V: Ord> Ord for Range<V>` implementations to `Range`. We use a simple ordering scheme where we zip the segments and compare all bounds in order. If all bounds are equal, the longer range is considered greater. (And if all zipped bounds are equal and we have the same number of segments, the ranges are equal). Not that this is distinct from contains operations, `r1 ≤ r2` (implemented by `Ord`) and `r1 ⊆ r2` (`subset_of`) have no relationship (even though `⊆` is a partial ordering on ranges).

Upstreams https://github.com/astral-sh/pubgrub/pull/31

### DIFF
--- a/src/range.rs
+++ b/src/range.rs
@@ -322,25 +322,51 @@ impl<V: Ord> Range<V> {
 /// left is smaller, since it starts earlier.
 /// ```
 fn cmp_bounds_start<V: PartialOrd>(left: Bound<&V>, right: Bound<&V>) -> Option<Ordering> {
-    match (left, right) {
-        (Unbounded, Unbounded) => Some(Ordering::Equal),
-        (Included(_left), Unbounded) => Some(Ordering::Greater),
-        (Excluded(_left), Unbounded) => Some(Ordering::Greater),
-        (Unbounded, Included(_right)) => Some(Ordering::Less),
-        (Included(left), Included(right)) => left.partial_cmp(right),
+    Some(match (left, right) {
+        // left:   ∞-----
+        // right:  ∞-----
+        (Unbounded, Unbounded) => Ordering::Equal,
+        // left:     [---
+        // right:  ∞-----
+        (Included(_left), Unbounded) => Ordering::Greater,
+        // left:     ]---
+        // right:  ∞-----
+        (Excluded(_left), Unbounded) => Ordering::Greater,
+        // left:   ∞-----
+        // right:    [---
+        (Unbounded, Included(_right)) => Ordering::Less,
+        // left:   [----- OR [----- OR   [-----
+        // right:    [--- OR [----- OR [---
+        (Included(left), Included(right)) => left.partial_cmp(right)?,
         (Excluded(left), Included(right)) => match left.partial_cmp(right)? {
-            Ordering::Less => Some(Ordering::Less),
-            Ordering::Equal => Some(Ordering::Greater),
-            Ordering::Greater => Some(Ordering::Greater),
+            // left:   ]-----
+            // right:    [---
+            Ordering::Less => Ordering::Less,
+            // left:   ]-----
+            // right:  [---
+            Ordering::Equal => Ordering::Greater,
+            // left:     ]---
+            // right:  [-----
+            Ordering::Greater => Ordering::Greater,
         },
-        (Unbounded, Excluded(_right)) => Some(Ordering::Less),
+        // left:   ∞-----
+        // right:    ]---
+        (Unbounded, Excluded(_right)) => Ordering::Less,
         (Included(left), Excluded(right)) => match left.partial_cmp(right)? {
-            Ordering::Less => Some(Ordering::Less),
-            Ordering::Equal => Some(Ordering::Less),
-            Ordering::Greater => Some(Ordering::Greater),
+            // left:   [-----
+            // right:    ]---
+            Ordering::Less => Ordering::Less,
+            // left:   [-----
+            // right:  ]---
+            Ordering::Equal => Ordering::Less,
+            // left:     [---
+            // right:  ]-----
+            Ordering::Greater => Ordering::Greater,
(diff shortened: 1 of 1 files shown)


src/range.rs
```diff
@@ -308,6 +308,104 @@ impl<V: Ord> Range<V> {
...
+/// left:   [--------∞
+/// right:  [-----]
+/// left is greater, since it starts earlier.
+///
+/// left:   [-----[
+/// right:  [-----]
+/// left is smaller, since it ends earlier.
+/// ```
+fn cmp_bounds_end<V: PartialOrd>(left: Bound<&V>, right: Bound<&V>) -> Option<Ordering> {
+    match (left, right) {
```
Eh2406: All branches return `Some(` does it end up being clear as `Some(match (left, right) {`?

src/range.rs
```diff
@@ -308,6 +308,104 @@ impl<V: Ord> Range<V> {
...
+        (Unbounded, Unbounded) => Some(Ordering::Equal),
+        (Included(_left), Unbounded) => Some(Ordering::Less),
+        (Excluded(_left), Unbounded) => Some(Ordering::Less),
```
Eh2406: Some of these branches are repetitive. Is it easier to read with `(_, Unbounded)` and similar groupings?

src/range.rs
```diff
@@ -308,6 +308,104 @@ impl<V: Ord> Range<V> {
...
+/// ```text
+/// left:   [--------∞
+/// right:  [-----]
+/// left is greater, since it starts earlier.
+///
+/// left:   [-----[
+/// right:  [-----]
+/// left is smaller, since it ends earlier.
+/// ```
+fn cmp_bounds_end<V: PartialOrd>(left: Bound<&V>, right: Bound<&V>) -> Option<Ordering> {
```
Eh2406, line 387: These seem conceptually similar to `left_end_is_smaller` and friends below, can they be used to simplify these methods? If not can those helpers be simplified by calling these functions?

Eh2406 (approved):
This is exactly the definition I would've chosen for an arbitrary ordering of ranges. I just have a couple of questions about code readability. We should go ahead and merge if the answers are "no".

src/range.rs
konstin: Done

src/range.rs
konstin: I added better docs to indicate why i left them as separate branches.

src/range.rs
konstin: The methods are different iff we are in an `Unbounded` branch, so unfortunately we can'T merge them.

Eh2406: I was not suggesting merging `cmp_bounds_end` with `cmp_bounds_start`. I was suggesting using them in `left_start_is_smaller` and `left_end_is_smaller`. Something like

```rust
fn left_start_is_smaller<V: PartialOrd>(left: Bound<V>, right: Bound<V>) -> bool {
    cmp_bounds_start(left.as_ref(), right.as_ref())
        .unwrap_or(Ordering::Greater)
        .is_le()
}

fn left_end_is_smaller<V: PartialOrd>(left: Bound<V>, right: Bound<V>) -> bool {
    cmp_bounds_end(left.as_ref(), right.as_ref())
        .unwrap_or(Ordering::Greater)
        .is_le()
}
```

src/range.rs
```diff
@@ -308,6 +308,154 @@ impl<V: Ord> Range<V> {
...
+        // left:   ---[
+        // right:  -----∞
+        (Excluded(_left), Unbounded) => Ordering::Less,
+        // left:  -----∞
+        // right: ---]
+        (Unbounded, Included(_right)) => Ordering::Greater,
+        // left:   -----] OR -----] OR ---]
+        // right:    ---] OR -----] OR -----]
+        (Included(left), Included(right)) => left.partial_cmp(right)?,
+        (Excluded(left), Included(right)) => match left.partial_cmp(right)? {
```
Eh2406, line 404: Would this be easier to read as `left.partial_cmp(right)?.then(Ordering::Less)`? It would be more concise, but not necessarily more readable.

mpizenberg: Personally, I like the explicitness instead of wondering what is the semantic of `then` for partial orders.

src/range.rs
```diff
@@ -1069,4 +1217,26 @@ pub mod tests {
...
+            Range::singleton(2u32).union(&Range::singleton(3u32)),
+            Range::singleton(2u32)
+                .union(&Range::singleton(3u32))
+                .union(&Range::singleton(4u32)),
+            Range::singleton(2u32).union(&Range::singleton(4u32)),
+            Range::singleton(3u32),
+        ];
+        let mut versions_sorted = versions.to_vec();
+        versions_sorted.sort();
+        assert_eq!(versions_sorted, versions);
```
mpizenberg, line 1241: I’d also compare it with the sorted of the reverse of `versions`, to make sure the implementation isn’t "just stable" because of mistakes returning `Ordering::Equal` instead of something else.

konstin: Good idea!

src/range.rs
konstin: I see, i misread the comment.

konstin: These functions are hard to reason about and easy to flip a branch. I prefer the explicitness for both of them.

src/range.rs
konstin: I prefer as much explicitness as possible here, so that you can reason the correctness of each branch in isolation by looking at as few examples as possible.

PR #251, opened 2024-08-22
perf: remove more redundant scanning

Eh2406:
Doing the stronger version of #175 because it turns out to be important to cargo crates.

While investigating https://github.com/Eh2406/pubgrub-crates-benchmark/issues/6  I took a shot in the dark and thought it might have to do with this optimization. Indeed this one change brings the time to analyze the Solana crates down from 127.25min to 90.85min a 28% improvement.

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -128,10 +128,9 @@ impl<DP: DependencyProvider> State<DP> {
                     Relation::AlmostSatisfied(package_almost) => {
                         // Add `package_almost` to the `unit_propagation_buffer` set.
                         // Putting items in `unit_propagation_buffer` more than once waste cycles,
-                        // but so does checking for duplicates.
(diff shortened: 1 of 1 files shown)


mpizenberg:
Nice one! Is it not too penalizing for the other cases?

mpizenberg (approved):

Eh2406:
2% regression for elm 3% improvement from "large case", so basically all within the noise.

konstin:
I've rebased uv onto the latest commits and the are no changes in our benchmarks.

image

Eh2406:
That makes perfect sense. This only really makes a difference in certain "pathological" cases. They just happened to be so slow that they are a large percentage of the benchmark suite I've chosen. But I'm glad it didn't regress your use cases.

PR #264, opened 2024-10-28
Improve Term documentation

x-hgg-x:
This PR adds better documentation for `Term`, clarifies the difference between `Positive(r)` and `Negative(r.complement())`, and adds justifications for `Term` mathematical operations.

### DIFF
--- a/src/term.rs
+++ b/src/term.rs
@@ -9,20 +9,22 @@ use crate::VersionSet;
 
 /// A positive or negative expression regarding a set of versions.
 ///
-/// If a version is selected then `Positive(r)` and `Negative(r.complement())` are equivalent, but
-/// they have different semantics when no version is selected. A `Positive` term in the partial
-/// solution requires a version to be selected. But a `Negative` term allows for a solution that
-/// does not have that package selected. Specifically, `Positive(VS::empty())` means that there was
-/// a conflict, we need to select a version for the package but can't pick any, while
-/// `Negative(VS::full())` would mean it is fine as long as we don't select the package.
+/// `Positive(r)` and `Negative(r.complement())` are not equivalent:
+/// * the term `Positive(r)` is satisfied if the package is selected AND the selected version is in `r`.
+/// * the term `Negative(r.complement())` is satisfied if the package is not selected OR the selected version is in `r`.
+///
+/// A `Positive` term in the partial solution requires a version to be selected, but a `Negative` term
+/// allows for a solution that does not have that package selected.
+/// Specifically, `Positive(VS::empty())` means that there was a conflict (we need to select a version for the package
+/// but can't pick any), while `Negative(VS::full())` would mean it is fine as long as we don't select the package.
 #[derive(Debug, Clone, Eq, PartialEq)]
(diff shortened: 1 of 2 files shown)


src/term.rs
```diff
@@ -110,6 +113,7 @@ impl<VS: VersionSet> Term<VS> {
     pub(crate) fn is_disjoint(&self, other: &Self) -> bool {
         match (self, other) {
             (Self::Positive(r1), Self::Positive(r2)) => r1.is_disjoint(r2),
+            // Unselected package is allowed in both terms.
             (Self::Negative(r1), Self::Negative(r2)) => r1 == &VS::empty() && r2 == &VS::empty(),
```
x-hgg-x: Should the result always be `false` here, since both terms contains the unselected package ?

mpizenberg: I don’t remember exactly when this was written, but my gut feeling is that this was created mainly to be used in multiple other places where the contents (semantic) of this function was already pre-existing. Some kind of refactoring, where an appropriate name was needed, and `is_disjoint` seemed like the best match.

Edit: The exact refactor happened here: https://github.com/pubgrub-rs/pubgrub/commit/0d80e9d40d13ce1df96ad8da84a263d28ae6e6ba
PR for context: https://github.com/pubgrub-rs/pubgrub/pull/177

mpizenberg:
I personally like the suggested changes. Thank you @x-hgg-x

src/term.rs
mpizenberg: Otherwise, I agree with you that two negative terms are never technically disjoint, so maybe there is an issue? Would be surprising considering the amount of tests around the project, but not impossible.

Eh2406: We often handle the empty case specially. For example it gets filtered out of incompatibilities. So I would not be surprised if in the end-to-end  tests every time we call `is_disjoint` one of the arguments has been already determined to be non-empty.

On the other hand, I'm surprised that were passing `is_disjoint_through_intersection`.   Perhaps our strategy is not generating the empty set often enough? Yup, looks like we generate ranges with `1..10`  segments which does not gen the empty set.

Eh2406:
@konstin you are the one most recently working to understand this so I leave the final decision to you.

src/term.rs
x-hgg-x: The `is_disjoint_through_intersection` test fails if we restrict the generated range to `0..1` segments in the proptest strategy:
```
Test failed: assertion `left == right` failed
  left: true
 right: false.
minimal failing input: r1 = Negative(
    Range {
        segments: [],
    },
), r2 = Negative(
    Range {
        segments: [],
    },
)
```

konstin (approved):
I'll wait till you converge on https://github.com/pubgrub-rs/pubgrub/pull/264#discussion_r1821200936, otherwise r+ from me

src/term.rs
Eh2406: 🎉 nice find!

Let's modify this line to always return false with a comment about why, and modify the generation to `0..10` so that we don't regress in the future.

mpizenberg: Agreed. Thanks for the finding!

PR #271, opened 2024-11-01
Add codespeed for continuous profiling (and sudoku)

konstin:
While pubgrub's performance is critical for both uv and cargo, there currently no benchmarking happening in the repository. [Codspeed](https://codspeed.io) runs our benchmarks with instruction counting, reporting more numbers with less variance than wall time.

We get feedback on every PR, can see trends over time and there are flamegraphs and flamegraph diffs in the web view. We've made good experiences with it in both ruff and uv.

With codspeed installed, we can start adding real-world benchmarks for uv and cargo to the pubgrub repo, and then optimize those. Next to the large case example, i've added the synthetic but fun sudoku example.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ thiserror = "1.0"
 version-ranges = { version = "0.1.0", path = "version-ranges" }
 
 [dev-dependencies]
-criterion = "0.5"
+criterion = { version = "2.7.2", package = "codspeed-criterion-compat" }
 env_logger = "0.11.5"
 proptest = "1.5.0"
 ron = "=0.9.0-alpha.0"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -32,7 +32,7 @@ thiserror = "1.0"
 version-ranges = { version = "0.1.0", path = "version-ranges" }
 
 [dev-dependencies]
-criterion = "0.5"
+criterion = { version = "2.7.2", package = "codspeed-criterion-compat" }
```
Eh2406, line 35: Does this change how I would run benchmarks locally?

Eh2406 (approved):
I will need to do some research about how to use it, but this is clearly an improvement!

PR #272, opened 2024-11-01
Regress performance: Use `Arc<usize>` for more sudoku realism

konstin:
and have codspeed catch the regression.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -3,7 +3,7 @@ name: Benchmarks (CodSpeed)
 on:
   push:
     branches:
-      - main
+      - dev
   pull_request:
   workflow_dispatch:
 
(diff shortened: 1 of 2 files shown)


konstin:
This is a good example to why the sudoku benchmark is very unrealistic (we can silence it from reporting) and why we need to port realistic examples:

image

Eh2406:
This is really cool! Thanks for the demonstration.

.github/workflows/benchmarks.yml
```diff
@@ -3,7 +3,7 @@ name: Benchmarks (CodSpeed)
 on:
   push:
     branches:
-      - main
+      - dev
```
konstin, line 6: @Eh2406 i used the wrong branch name, that's why we currently don't get data from ~~main~~ dev.

Eh2406:
I didn't realize this was something you actually wanted merged, I thought it was just a proof of what would happen if we regress performance. Merging now.

Eh2406 (approved):

konstin:
I had changed my mind :)

The `Arc`ing made the sudoku more realistic, so I discarded the smallvec change.

Issue #249, opened 2024-08-18
Support for `Range::union_many` or similar


zanieb:
I find myself constructing a range in a loop with `union`, which we worry is not performant. It'd be nice to have a `union_many` that takes an iterable of ranges and unions them in a single efficient operation.

Some example usage

https://github.com/astral-sh/uv/blob/0335efd0a901ee2351e663d0488b1ce87fd4cc80/crates/pep508-rs/src/marker/algebra.rs#L638-L645

https://github.com/astral-sh/uv/blob/e1a8beb64b9f6eb573b3af616ffff680c282a316/crates/uv-resolver/src/pubgrub/report.rs#L1098-L1102

I started to take a look at this myself, but realized I'm out of my depth :)

Eh2406:
Yes, `union` is linear in the length of the ranges being merged so doing it in a loop is quadratic.
The fundamental `union` algorithm is:
1. Create a combined list of all of the segments from all inputs.
2. Sort this list based on segments start.
3. Merge consecutive overlapping segments.
4. Declare what's left the union range.

Some optimizations can be done with the sort because we know the inputs are themselves sorted. Similarly we can duplicate the list as we constructe it for the classic 2 argument case.

A `union_many` could use a binary heap to find the next segment to process.

mpizenberg:
Well summarized!

It’s been a long time since I looked at this. I suppose a rather simple adaptation of the current code would just need to adjust the `smaller_interval` selection among a list of segments iterators (one per range). Instead of just checking the "left" and "right", it would check every iterator head.

https://github.com/pubgrub-rs/pubgrub/blob/68fd200cf6e1acfb4431fc9c04698d76b00173d3/src/range.rs#L437

That would be better than repeatedly computing union, but still not optimal. An optimal one I guess would also remember how the iterator heads are sorted among themselves. Is that what you meant with the binary heap @Eh2406 ?

Eh2406:
Yes. `smaller_interval` is "among all non-exhausted iterators pick the first segment from the one who's got the smallest start point". If we had a `IterWithOrdBySmallestStart` then we could calculate `smaller_interval` by popping a `BinaryHeap<IterWithOrdBySmallestStart>`. If the `IterWithOrdBySmallestStart` is not exhausted then re-add what's left to the `BinaryHeap`. If the binary keep is empty then you're done.

PR #273, opened 2024-11-04
`Ranges::from_iter`, `Ranges:from_unsorted` and document ranges invariants

konstin:
## Motivation

https://github.com/astral-sh/uv/pull/8797/files#diff-17170cd58cc6474d6c6ee0413af4b3b28610aca267bbe7863411627219c26549R232 needs to modify an existing `Ranges`. We don't want to allow direct access to ranges since this allows breaking the version-ranges invariants. This led me to two questions: How do we best construct ranges from a number of versions (#249), does the user need to hold up the invariants or do we sort and merge the given ranges? And: What are our invariants?

Given my previous profiling, construction ranges isn't nearly noticeable in benchmarks (but i'm gladly proven otherwise), the input ranges are too small and we do orders of magnitude more ranges constructions in the pubgrub algorithm itself, so the primary problem of https://github.com/astral-sh/uv/pull/8797/files#diff-17170cd58cc6474d6c6ee0413af4b3b28610aca267bbe7863411627219c26549R232 and #249 is ergonomics and zero-to-low overhead construction is a secondary problem, one that i'm tackling only because it would be inconsistent to have a slow API in pubgrub. Currently, we don't have a method for constructing ranges from multiple segments and this makes for an unergonomic API.

## Invariants

For the invariants, we need 1), but i'm not clear if we also need 2) and 3) as strong as they are. They are currently invariants in the version ranges, but we could weaken them. What i like about them is that they imply that any instance of ranges is "fully reduced", it can't be simplified further (without knowning the actual versions available).

1. The segments are sorted, from lowest to highest (through `Ord`).
2. Each segment contains at least one version (start < end).
3. There is at least one version between two segments.

## API

I added two functions: A `from_iter` where the user has to pass valid segments. This wants to be a `try_collect`, but that's still unstable. It is targeted at https://github.com/astral-sh/uv/pull/8797/files#diff-17170cd58cc6474d6c6ee0413af4b3b28610aca267bbe7863411627219c26549R232, which currently does this in a less secure fashion. There, you'd replace `iter_mut().for_each(...)` with an `.into_iter().map(...).collect()` (With the `IntoIter` i've also added this shouldn't even be an allocation; i'm optimizing this too for consistency's sake but i'd be surprised if it mattered to users). The other is `from_unsorted` which is ergonomics and could be used for https://github.com/astral-sh/uv/blob/0335efd0a901ee2351e663d0488b1ce87fd4cc80/crates/pep508-rs/src/marker/algebra.rs#L638-L645 and https://github.com/astral-sh/uv/blob/e1a8beb64b9f6eb573b3af616ffff680c282a316/crates/uv-resolver/src/pubgrub/report.rs#L1098-L1102 from #249. The user passes arbitrary segments and we're merging them into valid ranges.

Please discuss :)

### DIFF
--- a/version-ranges/src/lib.rs
+++ b/version-ranges/src/lib.rs
@@ -50,6 +50,13 @@ use smallvec::{smallvec, SmallVec};
 #[cfg_attr(feature = "serde", derive(serde::Serialize))]
 #[cfg_attr(feature = "serde", serde(transparent))]
 pub struct Ranges<V> {
+    /// Internally, [`Ranges`] are an ordered list of segments, where segment is a bounds pair.
+    ///
+    /// Invariants:
+    /// 1. The segments are sorted, from lowest to highest (through `Ord`).
+    /// 2. Each segment contains at least one version (start < end).
+    /// 3. There is at least one version between two segments.
+    ///
     /// Profiling in <https://github.com/pubgrub-rs/pubgrub/pull/262#discussion_r1804276278> showed
     /// that a single stack entry is the most efficient. This is most likely due to `Interval<V>`
     /// being large.
@@ -283,6 +290,37 @@ impl<V: Ord> Ranges<V> {
         }
     }
 
+    /// We want to use `iterator_try_collect`, but since it's unstable at the time of writing,
+    /// we expose a public `FromIterator<(Bound<V>, Bound<V>)>` method and use this for internal
+    /// testing.
+    fn try_from(
+        into_iter: impl IntoIterator<Item = (Bound<V>, Bound<V>)>,
+    ) -> Result<Self, FromIterError> {
+        let mut iter = into_iter.into_iter();
+        let Some(mut previous) = iter.next() else {
+            return Ok(Self {
+                segments: SmallVec::new(),
+            });
+        };
+        let mut segments = SmallVec::with_capacity(iter.size_hint().0);
+        for current in iter {
+            if !valid_segment(&previous.start_bound(), &previous.end_bound()) {
+                return Err(FromIterError::InvalidSegment);
+            }
+            if !end_before_start_with_gap(&previous.end_bound(), &current.start_bound()) {
+                return Err(FromIterError::OverlappingSegments);
+            }
+            segments.push(previous);
+            previous = current;
+        }
+        if !valid_segment(&previous.start_bound(), &previous.end_bound()) {
+            return Err(FromIterError::InvalidSegment);
+        }
+        segments.push(previous);
+        Ok(Self { segments })
+    }
+
+    /// See `Range.segments` docstring.
     fn check_invariants(self) -> Self {
         if cfg!(debug_assertions) {
             for p in self.segments.as_slice().windows(2) {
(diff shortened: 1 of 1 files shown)


version-ranges/src/lib.rs
```diff
@@ -283,6 +290,37 @@ impl<V: Ord> Ranges<V> {
         }
     }
 
+    /// We want to use `iterator_try_collect`, but since it's unstable at the time of writing,
+    /// we expose a public `FromIterator<(Bound<V>, Bound<V>)>` method and use this for internal
+    /// testing.
+    fn try_from(
```
Eh2406, line 296: when I see `try_from` I think of the `TryFrom` trait. Perhaps `try_from_iterator`?

version-ranges/src/lib.rs
```diff
@@ -283,6 +290,37 @@ impl<V: Ord> Ranges<V> {
...
+                return Err(FromIterError::OverlappingSegments);
+            }
+            segments.push(previous);
+            previous = current;
+        }
+        if !valid_segment(&previous.start_bound(), &previous.end_bound()) {
+            return Err(FromIterError::InvalidSegment);
+        }
+        segments.push(previous);
+        Ok(Self { segments })
```
Eh2406, line 320: let's put a call to `.check_invariants()` to make it easier to spot if we break things.

version-ranges/src/lib.rs
```diff
@@ -283,6 +290,37 @@ impl<V: Ord> Ranges<V> {
         }
     }
 
+    /// We want to use `iterator_try_collect`, but since it's unstable at the time of writing,
```
Eh2406, line 293: This user facing documentation also needs to describe what this function does, and what the requirements are on the arguments, and when it returns an error.

version-ranges/src/lib.rs
```diff
@@ -820,12 +858,84 @@ impl<V: Ord + Clone> Ranges<V> {
...
+                Ordering::Equal
+            } else if left_start_is_smaller(a.start_bound(), b.start_bound()) {
+                Ordering::Less
+            } else {
+                Ordering::Greater
+            }
+        });
+        // TODO(konsti): Handroll this. We're relying on the union implementation treating each
+        // segment as potentially coming from the other ranges and merging overlapping bounds.
+        Self { segments }.union(&Self {
```
Eh2406, line 884: Agree with your comment here, it's cute that `union` with `empty` happens normalize our invariants, but it seems grungy to rely on it. I rather unroll it here. If there's too much duplicated code then we can have a private `reestablish_invariants_by_unionizing` that both methods can call.

version-ranges/src/lib.rs
```diff
@@ -820,12 +858,84 @@ impl<V: Ord + Clone> Ranges<V> {
...
+        // We have three constraints we need to fulfil:
+        // 1. The segments are sorted, from lowest to highest (through `Ord`): By sorting.
+        // 2. Each segment contains at least one version (start < end): By `union`.
+        // 3. There is at least one version between two segments: By `union`.
+        let mut segments: SmallVec<_> = segments
+            .into_iter()
+            .filter(|segment| valid_segment(&segment.start_bound(), &segment.end_bound()))
+            .collect();
+        segments.sort_by(|a: &Interval<V>, b: &Interval<V>| {
+            if a.start_bound() == b.start_bound() {
```
Eh2406, line 873: I think we can reuse the existing `cmp_bounds_start` for this sort.

version-ranges/src/lib.rs
```diff
@@ -820,12 +858,84 @@ impl<V: Ord + Clone> Ranges<V> {
...
     /// Iterate over the parts of the range.
     pub fn iter(&self) -> impl Iterator<Item = (&Bound<V>, &Bound<V>)> {
         self.segments.iter().map(|(start, end)| (start, end))
     }
 }
 
+impl<V> IntoIterator for Ranges<V> {
+    type Item = (Bound<V>, Bound<V>);
+    // TODO(konsti): We must not leak the type here!
+    type IntoIter = smallvec::IntoIter<[Interval<V>; 1]>;
```
Eh2406, line 898: Specifically, this has to be opaque because we want to change the the representation to no longer have `Interval<V>`

Eh2406 (approved):
These all look like useful changes!

version-ranges/src/lib.rs
zanieb: Yeah this comment seems more like an internal note?

konstin: This isn't user facing, the function only exists internally for testing; if it weren't for the proptests this function wouldn't exist at all, `from_iter` is the real thing. I'll update the docs to redirect to `from_iter`.

konstin:
@Eh2406 what do you think about the invariants and about the functions in general? I've also been thinking that if we handroll the from unsorted variant, we could fast-path already matching inputs so we may not need the panicking function?

Eh2406:
> what do you think about the invariants and about the functions in general?

I think those are the invariants we currently maintain. The reason we need a "fully reduced" form or at least a canonicalised form is so that `Eq` works. But as long as these invariants are internal details, we can always change them later. Better internal documentation is always welcome.

The new functionality seems extremely useful.

konstin:
Thank you for all the discussion! I've split this into four PRs now:

* #276 
* #277 
* #278 
* #279

PR #276, opened 2024-11-08
Add `IntoIter` on `Ranges`

konstin:
To modify a version ranges and consume them with a construction method from https://github.com/pubgrub-rs/pubgrub/pull/273, we want to allow not cloning the versions. This is supported through `.into_iter()`.

### DIFF
--- a/version-ranges/src/lib.rs
+++ b/version-ranges/src/lib.rs
@@ -846,6 +846,27 @@ impl<V: Ord + Clone> Ranges<V> {
     }
 }
 
+// Newtype to avoid leaking our internal representation.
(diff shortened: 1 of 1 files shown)


Eh2406:
Do we also want to implement `ExactSizeIterator` and `DoubleEndedIterator` and `FusedIterator` on `RangesIter`?

konstin:
Good idea, added `ExactSizeIterator` and `DoubleEndedIterator`, `FusedIterator` seems to already come from `.fuse()` though?

Eh2406 (approved):
Thanks!

PR #278, opened 2024-11-08
Add `FromIter` for `Ranges`

### LINKED ISSUE #249: Support for `Range::union_many` or similar

konstin:
Add a method to construct ranges from an iterator of arbitrary segments. This allows to `.collect()` an iterator of tuples of bounds. This is more ergonomic than folding the previous ranges with the next segment each time, and also faster.

Split out from https://github.com/pubgrub-rs/pubgrub/pull/273
Closes https://github.com/astral-sh/pubgrub/pull/33
Fixes https://github.com/pubgrub-rs/pubgrub/issues/249

### DIFF
--- a/version-ranges/src/lib.rs
+++ b/version-ranges/src/lib.rs
@@ -879,7 +879,7 @@ impl<V> IntoIterator for Ranges<V> {
     }
 }
 
-impl<V: Ord + Debug> FromIterator<(Bound<V>, Bound<V>)> for Ranges<V> {
+impl<V: Ord> FromIterator<(Bound<V>, Bound<V>)> for Ranges<V> {
     /// Constructor from arbitrary, unsorted and potentially overlapping ranges.
     ///
     /// This is equivalent, but faster, to computing the [`Ranges::union`] of the
@@ -889,6 +889,10 @@ impl<V: Ord + Debug> FromIterator<(Bound<V>, Bound<V>)> for Ranges<V> {
         // 1. The segments are sorted, from lowest to highest (through `Ord`): By sorting.
         // 2. Each segment contains at least one version (start < end): By `union`.
         // 3. There is at least one version between two segments: By `union`.
+
+        // For this implementation, we choose to only build a single smallvec and insert or remove
+        // in it, instead of e.g. collecting the segments into one smallvec, sorting that and then
+        // construction the second smallvec without shifting.
         let mut segments: SmallVec<[Interval<V>; 1]> = SmallVec::new();
 
         for segment in iter {
@@ -908,7 +912,8 @@ impl<V: Ord + Debug> FromIterator<(Bound<V>, Bound<V>)> for Ranges<V> {
                     &segment.start_bound(),
                 );
 
-            // Is it overlapping with the following segment?
+            // Is it overlapping with the following segment? We'll check if there's more than one
+            // overlap later.
             let next_overlapping = insertion_point < segments.len()
                 && !end_before_start_with_gap(
                     &segment.end_bound(),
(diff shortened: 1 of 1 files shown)


version-ranges/src/lib.rs
```diff
@@ -1130,6 +1211,12 @@ pub mod tests {
             }
             assert!(simp.segments.len() <= range.segments.len())
         }
+
+        #[test]
+        fn from_iter_valid(segments in proptest::collection::vec(any::<(Bound<u32>, Bound<u32>)>(), ..30)) {
+            // We check invariants in the method.
+            Ranges::from_iter(segments.clone());
```
Eh2406: We could strengthen this test by checking that the result is equivalent to iteratively calling union.

version-ranges/src/lib.rs
```diff
@@ -826,6 +826,87 @@ impl<V: Ord + Clone> Ranges<V> {
...
+                    // We can reuse the existing element by extending it.
+                    segments[insertion_point].0 = segment.0;
+                }
+                (false, false) => {
+                    // previous:  |------|
+                    // segment:             |------|
+                    // following:                      |------|
+                    //
+                    // final:    |------|   |------|   |------|
+                    segments.insert(insertion_point, segment);
```
Eh2406, line 1053: This is doing insertion sort. Witch is `O(n^2)`, because this `insert` (or the `remove` above) is `O(n)`. Shoving all the segments into a heap should get us `O(n*log(n))` time, but take `O(n)` space. I don't know if it actually matters in practice.

If the point is to be practical I'm happy to merge this with a comment about alternative implementations. If the point of this is to implement the algorithmically best solution, we can do better than this.

version-ranges/src/lib.rs
konstin: This caught a significant oversight!

version-ranges/src/lib.rs
konstin: One solution would be collecting the segments into a sorted datastructure, then building a second ranges without any shifting around: We pay the price of two datastructure for no moving around. This solution works with only a single ranges, inserting or removing elements as we detect overlaps: If the elements are already in order, it's only pushes to a single smallvec (I chose not to reserve, since we could vastly over-reserve). As usual, I'd go with this implementation and change it if benchmarking with users shows that another algorithm would be better.

version-ranges/src/lib.rs
```diff
@@ -826,6 +826,175 @@ impl<V: Ord + Clone> Ranges<V> {
...
+impl<V: Ord> FromIterator<(Bound<V>, Bound<V>)> for Ranges<V> {
+    /// Constructor from arbitrary, unsorted and potentially overlapping ranges.
+    ///
+    /// This is equivalent, but faster, to computing the [`Ranges::union`] of the
+    /// [`Ranges::from_range_bounds`] of each segment.
+    fn from_iter<T: IntoIterator<Item = (Bound<V>, Bound<V>)>>(iter: T) -> Self {
+        // We have three constraints we need to fulfil:
+        // 1. The segments are sorted, from lowest to highest (through `Ord`): By sorting.
+        // 2. Each segment contains at least one version (start < end): By `union`.
+        // 3. There is at least one version between two segments: By `union`.
```
Eh2406: "By `union`" does not match the implementation. While I'm making nits about comments, let's  make this clear.

version-ranges/src/lib.rs
Eh2406: Can we add a comment  on these two lines explaining that the call is `O(n)`  which makes the algorithm `O(n^2)`  but it should be good enough for now. That way  somebody looking at a flame graph that's hot here  has some clue that there is  a fix we've already thought of.

Eh2406 (approved):
Two nits about comments.

version-ranges/src/lib.rs
konstin: That's a good catch, i forgot to update the comment after changing the implementation.

version-ranges/src/lib.rs
konstin: Added a comment here and a more general explanation of the tradeoff above.

konstin:
For future reference, there's a `from_normalized` fast path follow-up in https://github.com/pubgrub-rs/pubgrub/compare/dev...konsti/dev/from-noramlized.

PR #279, opened 2024-11-08
Add `Ranges::from_normalized`

konstin:
Add a fast path alternative to `Ranges::from_iter`. This makes the ranges invariants part of the API, which I think is fine because i don't expect them to change. I'm not sure if we should yet, since this is not a performance critical part really.

### DIFF
--- a/version-ranges/src/lib.rs
+++ b/version-ranges/src/lib.rs
@@ -283,6 +283,39 @@ impl<V: Ord> Ranges<V> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


Eh2406:
Seems reasonable on a quick read.

konstin:
I'll leave it as a draft for now, we can re-evaluate this when we have a use case not covered by #278

konstin:
It doesn't look like we're merging this soon.

PR #280, opened 2024-11-08
Add .vscode in .gitignore

x-hgg-x:
Add `.vscode` in `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
 /target
-
+.vscode


zanieb:
My personal stance is that editor-specific gitignore entries belong in a user-level gitignore (`git config --global core.excludesfile $HOME/.gitignore`) since they're not specific to the project.

Eh2406:
I concur with @zanieb position here. Git also has a per project the equivalent of `excludesfile` in ... `.git/info/exclude` I think.

Does one of those approaches work for you @x-hgg-x ?

x-hgg-x:
Yes it works, I will close this PR.

PR #281, opened 2024-11-08
Add bench for backtracking

x-hgg-x:
Part of #274.

Add a new bench to follow backtracking performance.


benches/backtracking.rs
```diff
@@ -0,0 +1,52 @@
...
+use std::collections::BTreeMap;
+
+use criterion::*;
+use pubgrub::OfflineDependencyProvider;
+use version_ranges::Ranges;
+
+type Deps = BTreeMap<u32, BTreeMap<u8, BTreeMap<u32, Vec<(u8, u8)>>>>;
+
+fn bench(b: &mut Bencher, data: &str) {
+    let dependencies = ron::de::from_str::<Deps>(data).unwrap();
```
Eh2406: Why does this use a different file format than the existing benchmark files?

benches/backtracking.rs
```diff
@@ -0,0 +1,52 @@
...
+                }
+                (d, r)
+            });
+            dependency_provider.add_dependencies(p, v, deps_iter);
+        }
+    }
+
+    if let Some((p, v)) = dependencies
+        .first_key_value()
+        .and_then(|(&p, vmap)| vmap.first_key_value().map(|(&v, _)| (p, v)))
```
Eh2406: It would be really nice if this identified "the problematic crate from this file" more directly than "the one with the smallest index". I get that if this is generated from a arena based system package `0` version `0` has to be the root of the resolution. But, I definitely isolated real-world test cases where a non-root package was actually slower than the original problem.

Time resolve each package/version, and benchmark the slowest one would work. But timing is so unreliable without running things repeatedly. Count the number of `should cancel` calles might work, as a proxy for time. Depending on where were spending our time. If we exposed a way to count the number of backtracks, we could use that as a proxy. Even if it's not "always the correct answer" this it at least does what the name of this benchmark claims.

test-examples/backtracking.ron
Eh2406, line 1: Can the name of this file gives some hints about where it came from?

Also, it's a large file both in file size and runtime. It's not always possible, but if we could figure out how to make it 1/4 the size while only making it 4x faster it would make a huge difference. It would be amazing to do super linear reductions, if we could get it 1/10 the size but still more then 1/2 the time it would be amazing. But there is no reliable way to find that. The longer a benchmark runs the more likely it is to be interrupted by the OS or by thermal throttling. On the other hand, the reason we care about this case is because it's slow.

Eh2406 (changes requested):
I look forward to seeing these numbers improve!

test-examples/backtracking.ron
mpizenberg: Thanks a lot for putting this together! I’d say I’ve the same concerns as Jacob here. We have been extra careful keeping pubgrub slim on all fronts, including git.

benches/backtracking.rs
x-hgg-x: > It would be really nice if this identified "the problematic crate from this file"

This bench comes from the resolution of `solana-archiver-lib v1.1.12` for cargo, and I already know this is the slowest crate for pubgrub (from your benchmark at https://github.com/Eh2406/pubgrub-crates-benchmark/issues/6#issuecomment-2402992814). So for me there is no need to try resolving for all its dependencies, which will only slows the bench.

benches/backtracking.rs
x-hgg-x: The new format is more compact to reduce the file size.

test-examples/backtracking.ron
x-hgg-x: Bench files are now smaller.

benches/backtracking.rs
x-hgg-x: I have updated the bench files which are now synthetic. I didn't include the generation code in this PR because it needs code from #274. Here is the code for reference: [gen.rs.tar.gz](https://github.com/user-attachments/files/17689720/gen.rs.tar.gz).

konstin:
Can we improve prioritization to avoid going into the pathological case? The following bad hack already gives a 1.5x speedup, could we avoid reassigning all the other packages after (bad) solana versions by improving `prioritize`?

```rust
    fn prioritize(
        &mut self,
        package: Package,
        range: VersionSet,
        arena: &PackageArena<Self::P>,
    ) -> Self::Priority {
        let mut x = ((range.count() as u64) << 32) + package.get() as u64;
        if !arena.pkg(package).unwrap().to_string().contains("solana") {
            x = x + 1 << 62
        }
        Reverse(x)
    }
```

```
hyperfine --runs 3 "./pubgrub-bench-base" "./pubgrub-bench-prio"
Benchmark 1: ./pubgrub-bench-base
  Time (mean ± σ):      4.013 s ±  0.003 s    [User: 3.907 s, System: 0.104 s]
  Range (min … max):    4.011 s …  4.017 s    3 runs
 
Benchmark 2: ./pubgrub-bench-prio
  Time (mean ± σ):      2.643 s ±  0.004 s    [User: 2.500 s, System: 0.142 s]
  Range (min … max):    2.639 s …  2.647 s    3 runs
 
Summary
  ./pubgrub-bench-prio ran
    1.52 ± 0.00 times faster than ./pubgrub-bench-base
```

Re https://github.com/Eh2406/pubgrub-crates-benchmark/issues/6#issuecomment-2397911042: In the backtracking itself, why don't we learn the incompatibilities, shouldn't we be able to memoize that only a certain range works and skip over it in the next attempt? For https://github.com/Eh2406/pubgrub-crates-benchmark/issues/6#issuecomment-2400296431, what's the representation for features and virtual packages, can we improve that? In uv, we found a backtracking bug that was fixed by improving the ordering and bounds around feature packages.

For adding large, generated files to the repository, we should add information on how they were generated and how they should be interpreted: As a non-cargo dev, I currently see only a large tree of numbers, and can't tell how they should be resolved or what it means if we're taking a certain path through the packages and versions.

x-hgg-x:
> Can we improve prioritization to avoid going into the pathological case? 

Yes it should be possible, but we need to find a generic method without knowing the shape of the dependency tree.

Maybe we can compute the current tree depth and prioritize packages with smaller depths ?

> Re [Eh2406/pubgrub-crates-benchmark#6 (comment)](https://github.com/Eh2406/pubgrub-crates-benchmark/issues/6#issuecomment-2397911042): In the backtracking itself, why don't we learn the incompatibilities, shouldn't we be able to memoize that only a certain range works and skip over it in the next attempt? For [Eh2406/pubgrub-crates-benchmark#6 (comment)](https://github.com/Eh2406/pubgrub-crates-benchmark/issues/6#issuecomment-2400296431), what's the representation for features and virtual packages, can we improve that? 

These examples are using the current resolver of cargo, which doesn't use pubgrub yet.
 
> For adding large, generated files to the repository, we should add information on how they were generated and how they should be interpreted.

I updated the benchmark which is now synthetic and not based on cargo anymore (see https://github.com/pubgrub-rs/pubgrub/pull/281#discussion_r1835548180).

konstin:
> Yes it should be possible, but we need to find a generic method without knowing the shape of the dependency tree.
>
> Maybe we can compute the current tree depth and prioritize packages with smaller depths ?

There's a big space of possibilities here! For inspiration, here's uv's heuristic: https://github.com/astral-sh/uv/blob/14507a1793f12bf884786be49d716ca8f8322340/crates/uv-resolver/src/pubgrub/priority.rs. We pick more tightly constrained packages first. Another option is to prioritize package that caused conflicts earlier (by counting conflicts). I'm sure @Eh2406 has good ideas specifically for cargo, too.

konstin:
Given that the benchmark is synthetic, can we generate it on-the-fly in the bench code? This would also make documenting it easier, together with how this was extracted from the real solana crate case. (Large generated files in the repo, especially when we have different types of syntaxes for benchmarks, tend to become a maintenance burden).

Generally, with more context on the change and more documentation on the feature, i could give you a more helpful review here.

benches/backtracking.rs
Eh2406: If it's synthetic, I think I'd rather have the generator than the generated file. It also needs good documentation about:
- Why it is pathological? Why does the structure force the resolver to spend it's time on what.
- What does it have to do with real cases? What real  situation induced this kind of pathological encoding.

The existing "large case"  file does not do well on either of these standards. It is synthetic.  It was pathological due to problems that have long since been fixed.  The only thing real world about it is that it is the kind of graph our prop-test  tends to generate.

The new sudoku benchmarks  are semisynthetic.  It is a real ways to encode a real problem, which makes them non-synthetic. On the other hand using a  pubgrub is a sudoku solver is pretty odd behavior by itself. It runs so smoothly that it doesn't have a pathology to document.

x-hgg-x:
I updated the PR: the bench data is now generated in the code.

benches/backtracking.rs
```diff
@@ -0,0 +1,38 @@
...
+use version_ranges::Ranges;
+
+fn bench(c: &mut Criterion, case: &str, package_count: u32, version_count: u32) {
+    let mut dependency_provider = OfflineDependencyProvider::<u32, Ranges<u32>>::new();
+    let r = Ranges::strictly_lower_than(version_count);
+    dependency_provider.add_dependencies(0u32, 0u32, [(1u32, r)]);
+
+    for n in 1..package_count {
+        for v in 0..version_count {
+            let r = Ranges::strictly_lower_than(v);
```
Eh2406: This  pattern of layers of crates that each depend on the next can easily cause exponential  resolution time in naïve backtracking resolvers. Some variation on this is definitely a benchmark we should keep an eye on.

Why is this `strictly_lower_than` instead of `higher_than`?

I would imagine that  if were using `strictly_lower_than` and  we are processing version 9,  then we have to check  the dependency `(n + 1, <9)`,  which we discover doesn't work.  then when we go on to processing version 8 we now have to process `(n + 1, <8)`. This only takes a  constant amount of time  because `<8`  is a subset of `<9`.  On the other hand  if we used `higher_than`  and we process version 9,  then we have to check  the dependency `(n + 1, >=9)`, which we discover doesn't work.  then when we go on to processing version 8 we now have to process `(n + 1, >=8)`. Which involves actually having to go do some work to figure out there's a problem with the 8 we have not looked at befor.

If `strictly_lower_than`  is slower than `higher_than`  I would very much want to know why.

That same question could also be asked as  how is this like the realistic  Solana example?  For most of those crates we have the pattern `n, v, [(n + 1, ^v)` Where `^`  basically de-sugars to `>=`. Could you explain why  this synthetic benchmark will have similar performance characteristics to Solana?

benches/backtracking.rs
```diff
@@ -0,0 +1,38 @@
...
+        b.iter(|| {
+            let _ = pubgrub::resolve(&dependency_provider, 0u32, 0u32);
+        })
+    });
+}
+
+fn bench_group(c: &mut Criterion) {
+    bench(c, "backtracking_small", 5, 60);
+    bench(c, "backtracking_medium", 5, 200);
+    bench(c, "backtracking_large", 5, 500);
```
Eh2406: Why 60, 200, 500? Why only 5 layers?

benches/backtracking.rs
x-hgg-x: I replaced `strictly_lower_than` with `higher_than` which has the same performance.

benches/backtracking.rs
x-hgg-x: These values are not special, there were adjusted so that the benchmark takes around 10 times more time between `backtracking_medium` and `backtracking_small` and between `backtracking_large` and `backtracking_medium` (here the resolve time is cubic over the number of versions).

The only constraint is that I want one benchmark with less than 63 versions and another with more than 63 versions, so that I can compare performance of u64 version set in future commits.

benches/backtracking.rs
x-hgg-x: This benchmark reproduces the backtracking behavior of Solana and is close to the test that I recently added in cargo: https://github.com/Eh2406/pubgrub-crates-benchmark/issues/6#issuecomment-2400296431.

benches/backtracking.rs
Eh2406: I have had a few minutes to play with it locally. Running the `500` takes 16 min for me witch seams like a long time to be waiting for results.

> (here the resolve time is cubic over the number of versions).

I used [`benchmark_group`](https://docs.rs/criterion/latest/criterion/struct.BenchmarkGroup.html) and `bench_with_input` to make a nice graph.

lines

Which is definitely not linear.  Throwing it in Excel a quadratic fit the data quite well `R=0.995` (with the intercept pin to zero). But 5 points  may not provide enough data to distinguish  between a quadratic in a cubic.

benches/backtracking.rs
Eh2406: Interestingly the  flame graph provided by codspeed https://codspeed.io/pubgrub-rs/pubgrub/branches/x-hgg-x:new-bench shows this benchmark spending most of its time in `satisfier_search`, which  if I remember correctly is not where `Solana` spends its time.

benches/backtracking.rs
Eh2406: By the way, this micro benchmark gets a lot faster when using #257. 

```
backtracking/60         time:   [2.4683 ms 2.4746 ms 2.4825 ms]
                        change: [-91.340% -91.314% -91.284%] (p = 0.00 < 0.05)
                        Performance has improved.
backtracking/100        time:   [6.5867 ms 6.5943 ms 6.6021 ms]
                        change: [-94.393% -94.384% -94.376%] (p = 0.00 < 0.05)
                        Performance has improved.
backtracking/150        time:   [14.602 ms 14.617 ms 14.632 ms]
                        change: [-95.852% -95.847% -95.843%] (p = 0.00 < 0.05)
                        Performance has improved.
backtracking/200        time:   [25.453 ms 25.489 ms 25.525 ms]
                        change: [-96.715% -96.710% -96.705%] (p = 0.00 < 0.05)
                        Performance has improved.
```

benches/backtracking.rs
x-hgg-x: I committed new benchmarks representing various other patterns found in solana.

benches/backtracking.rs
x-hgg-x: I committed two new benchmarks which will not be impacted if #257 is merged.

konstin:
> I updated the PR: the bench data is now generated in the code.

Thanks, that's much easier to follow!

Can you add some docstrings explaining the benchmarks functions? Both what package structure they build, and how this relates to the real world case we are modelling here, like telling they story around how we get from trying to fix the slowest cargo-pubgrub cases to this small function. As a maintainer, it's important to have context on the changes, so we can reason about future change to this code. Imagine someone with no cargo background sees changes in these benchmarks from their pubgrub changes, and tries to make sense of what they mean. Or the other way round, someone finds a way to speed up these benchmarks, can they be confident that this does indeed improve solana or have they accidentally just hacked the benchmark?

x-hgg-x:
These new benchmarks are not specific to the cargo implementation nor a specific crate resolution: they describe general patterns which can cause backtracking, so they should be useful even for `uv`.

konstin:
I think it's good to have such benchmarks (they are definitely way more realistic than sudoku dummies), but in general we should aim at benchmarks that represent real world use cases. While pubgrub is a SAT solver technically, packaging often follows very specific patterns and often there are only a few hard patterns per ecosystem, which we should try to capture in benchmarks. See e.g. this talk about the perf oriented js package manager [oregene](https://github.com/orogene/orogene): https://youtu.be/eh3VME3opnE?si=vuUuRkJJVsoSOqWZ&t=323.

In uv, this was notably https://github.com/pubgrub-rs/pubgrub/issues/135, other representative slow example these were boto3/botocore/urllib (backtracking over two packages, largely a prefetching problem plus also https://github.com/pubgrub-rs/pubgrub/issues/135), apache-airflow (here package prioritization mattered a lot) and transformers with all extras (a huge dep tree). We optimized on bio-embeddings, too, which was a somewhat artificial case, but exercised a lot of the problems in resolving scientic/ML projects, including transformers. It allowed profiling for https://github.com/pubgrub-rs/pubgrub/pull/177, which in turn unblocked https://github.com/astral-sh/uv/pull/2452.

In cargo, solana is the exceptionally difficult case, which seems to follow a pattern we don't observe in uv: Even the pathological bio-embeddings benchmark now takes only 3s on my machine, and i don't know any real package that would even take 1s with cached metadata on my machine.

Again, I'm not trying to say that these benchmarks are bad (we should have diverse benchmark coverage!), but I want to explain why sharing context and adding documentation around benchmarks is important.

x-hgg-x:
I added some docstrings to explain how these benchmarks were derived from a real cargo resolution.

Eh2406 (approved):
These look like reasonable things for our project to keep an eye on.

PR #282, opened 2024-11-14
Release version-ranges v0.1.1

konstin:
Release version-ranges v0.1.1 and start a changelog. With this release out, i want to do an announcement post.

### DIFF
--- a/version-ranges/Changelog.md
+++ b/version-ranges/Changelog.md
@@ -1,0 +1,12 @@
+# Changelog
+
+Changelog for the version-ranges crate.
+
+## v0.1.1
+
+* Added `Ranges::from_iter`
+* Implement `IntoIter` on `Ranges`
+
+## v0.1.0
+
+Initial release!


Cargo.toml
```diff
@@ -37,7 +37,7 @@ env_logger = "0.11.5"
 proptest = "1.5.0"
 ron = "=0.9.0-alpha.0"
 varisat = "0.2.2"
-version-ranges = { version = "0.1.0", path = "version-ranges", features = ["proptest"] }
```
Eh2406: Why does this version number need to be bumped?  PubGrub  does not rely on any of the new functionality.

Eh2406 (approved):
One nit. But it is all good.

Cargo.toml
konstin: Normally i just bump patch versions out of convenience (we don't need to test something older if the bound is on latest).

We should consider setting up CI that checks that pubgrub works with the lowest crates.io version-ranges, if we aren't on a version that isn't released to crates.io yet.

PR #283, opened 2024-11-19
Move offline provider into a separate module

x-hgg-x:
Part of https://github.com/pubgrub-rs/pubgrub/pull/274.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -212,6 +212,7 @@
 
 mod error;
 mod package;
+mod provider;
 mod report;
 mod solver;
 mod term;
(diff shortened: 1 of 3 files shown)


konstin:
There are no code (logic) changes here, right? Cause that means that our benchmarks are noisy :/

x-hgg-x:
Iterators in `choose_version` and `prioritize` were no longer inlined after moving the code. Adding `#[inline]` fixed this.

Eh2406 (approved):
Looks good to me. Leaving open if @konstin has thoughts.

konstin (approved):

PR #285, opened 2024-11-21
Monitor only the resolve step in the sudoku bench

x-hgg-x:
This PR modifies the sudoku bench to only monitor the resolve step.


Eh2406 (approved):
This seems reasonable, @konstin up to you.

konstin:
Some more general thoughts: For real benchmarks, we should track both the setup time and the runtime, since the user has to pay both. But since the sudoku is more a gadget (we can take it out of codspeed now that we have better benchmarks), it's fine to only track the resolve step here.

konstin (approved):

PR #286, opened 2024-11-21
Add cold annotations

x-hgg-x:
This PR add `#[cold]` annotations to specific functions to hint the compiler that they should not be inlined, so that other more critical functions like iterators has better chance to be inlined.

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -79,6 +79,7 @@ impl<DP: DependencyProvider> State<DP> {
     }
 
     /// Add an incompatibility to the state.
+    #[cold]
     pub(crate) fn add_incompatibility_from_dependencies(
         &mut self,
         package: Id<DP::P>,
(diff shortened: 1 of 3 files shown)


Eh2406:
Why do we think these functions are cold? Is there perf data showing this helping?

x-hgg-x:
> Is there perf data showing this helping?

Yes, this will avoid future unrelated drops in performance when moving code (the addition of `#[inline]` in #283 was not enough). In #274 this improved performance by 5-10%, because iterators were not inlined before.

> Why do we think these functions are cold?

The problem this solves is that LLVM tries to inline a maximum of code if possible, but if a function is too big it will stop inlining in the middle of the function because the analysis cost is too big.
By giving LLVM specific cold functions which doesn't cause performance drops when not inlined, the inline cost is reduced and the function content is more likely to be inlined.
I chose these specific functions because they were already not inlined, so this PR prevents their inlining.

Eh2406 (approved):
Seams ok.

PR #291, opened 2024-12-05
 use intermediate satisfier causes in priority statistics 

Eh2406:
This provides more information to the `DP::prioritize` call. Inspired by @x-hgg-x  investigation https://github.com/pubgrub-rs/pubgrub/pull/274#issuecomment-2504970942, it provides the number of times the package has been involved in any backtracking conflict.  Inspired by UV's  prioritization, it provides the order the package was discovered. When using `Priority = (conflict_count, Reverse<matched versions>, Reverse<discovery_order>)` Resolving all crates (including Solana) on 10 threads went from `162.45min` to `88.96min`! The slowest individual crate went from `solana-archiver-lib@1.1.0 95s` to `solana-archiver-lib@1.1.7 21s`! (Note that the runtime for any individual crate may vary run to run due to system load and scheduling.)

The new fields are hidden behind a `PackageResolutionStatistics` with getters,  to make it easier to add more in the future without additional breaking changes. Getters are used in case some value we want to return is expensive to  calculate and so we only want to calculate it if someone can read it.  I attempted to use that pattern  to avoid the hash map look up for `conflict_count`, but then `PackageResolutionStatistics` ends up generic over `P`. Which seemed unpleasant. If people have other suggestions for how to future proof this API I'm all ears!

Theoretically the package name and range  could also become getters on `PackageResolutionStatistics`. But I consider those two exponentially more important and more useful than the ones where adding now. Deserving of always being computed and getting dedicated API surface. But opinions may vary.

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -209,37 +209,36 @@ impl<DP: DependencyProvider> State<DP> {
         Ok(satisfier_causes)
     }
 
-    /// Return the root cause or the terminal incompatibility.
-    /// CF <https://github.com/dart-lang/pub/blob/master/doc/solver.md#unit-propagation>
+    /// Return the root cause or the terminal incompatibility. CF
+    /// <https://github.com/dart-lang/pub/blob/master/doc/solver.md#unit-propagation>
     ///
-    /// Usually by the time we have a conflict `unit_propagation` has done a lot of work.
-    /// So the actual conflict we find is important, but not particularly actionable.
-    /// It says something like "the dependency on package X and the dependency on package Y are incompatible".
-    /// To make it actionable we want to track it back to decisions that made the dependency required.
-    /// "The decision on B is incompatible with the decision on C,
-    ///  because unit propagation from just those decisions will lead to the conflict about X and Y"
-    /// is much more actionable, backtrack until one of those decisions can be revisited.
-    /// To make a practical, we really only need one of the terms to be a decision.
-    /// We may as well leave the other terms general. Something like
-    /// "the dependency on the package X is incompatible with the decision on C" tends to work out pretty well.
-    /// Then if A turns out to also have a dependency on X the resulting root cause is still useful.
-    /// Of course, this is more heuristics than science. If the output is too general, then `unit_propagation` will
-    /// handle the confusion by calling us again with the next most specific conflict it comes across.
-    /// If the output is to specific, then the outer `solver` loop will eventually end up calling us again
-    /// until all possibilities are enumerated.
+    /// When we found a conflict, we want to learn as much as possible from it, to avoid making (or
+    /// keeping) decisions that will be rejected. Say we found that the dependency requirements on X and the
+    /// dependency requirements on Y are incompatible. We may find that the decisions on earlier packages B and C
+    /// require us to make incompatible requirements on X and Y, so we backtrack until either B or C
+    /// can be revisited. To make it practical, we really only need one of the terms to be a
+    /// decision. We may as well leave the other terms general. Something like "the dependency on
+    /// the package X is incompatible with the decision on C" tends to work out pretty well. Then if
+    /// A turns out to also have a dependency on X the resulting root cause is still useful.
+    /// (`unit_propagation` will ensure we don't try that version of C.)
+    /// Of course, this is more heuristics than science. If the output is too general, then
+    /// `unit_propagation` will handle the confusion by calling us again with the next most specific
+    /// conflict it comes across. If the output is too specific, then the outer `solver` loop will
+    /// eventually end up calling us again until all possibilities are enumerated.
     ///
-    /// This function combines incompatibilities with things that make the problem inevitable to end up with a
-    /// more useful incompatibility. For the correctness of the PubGrub algorithm only the final output is required.
-    /// By banning the final output, unit propagation will prevent the intermediate steps from occurring again,
-    /// at least prevent the exact same way. However, the statistics collected for `prioritize`may want
-    /// to analyze those intermediate steps. For example we might start with "there is no version 1 of Z",
-    /// and `conflict_resolution` may be able to determine that "that was inevitable when we picked version 1 of X"
-    /// which was inevitable when picked W and ... and version 1 of B, which was depended on by version 1 of A.
-    /// Therefore the root cause may simplify all the way down to "we cannot pick version 1 of A".
-    /// This will prevent us going down this path again. However when we start looking at version 2 of A,
-    /// and discover that it depends on version 2 of B, we will want to prioritize the chain of intermediate steps
-    /// to confirm if it has a problem with the same shape.
-    /// The `satisfier_causes` argument keeps track of these intermediate steps so that the caller can use.
+    /// To end up with a more useful incompatibility, this function combines incompatibilities into
+    /// derivations. Fulfilling this derivation implies the later conflict. By banning it, we
+    /// prevent the intermediate steps from occurring again, at least in the exact same way.
+    /// However, the statistics collected for `prioritize` may want to analyze those intermediate
+    /// steps. For example we might start with "there is no version 1 of Z", and
+    /// `conflict_resolution` may be able to determine that "that was inevitable when we picked
+    /// version 1 of X" which was inevitable when we picked W and so on, until version 1 of B, which
+    /// was depended on by version 1 of A. Therefore the root cause may simplify all the way down to
+    /// "we cannot pick version 1 of A". This will prevent us going down this path again. However
+    /// when we start looking at version 2 of A, and discover that it depends on version 2 of B, we
+    /// will want to prioritize the chain of intermediate steps to check if it has a problem with
+    /// the same shape. The `satisfier_causes` argument keeps track of these intermediate steps so
+    /// that the caller can use them for prioritization.
     #[allow(clippy::type_complexity)]
     #[cold]
     fn conflict_resolution(
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -133,7 +133,7 @@ impl<DP: DependencyProvider> State<DP> {
         &mut self,
         package: Id<DP::P>,
     ) -> Result<SmallVec<(Id<DP::P>, IncompDpId<DP>)>, NoSolutionError<DP>> {
-        let mut root_causes = SmallVec::default();
+        let mut satisfier_causes = SmallVec::default();
         self.unit_propagation_buffer.clear();
         self.unit_propagation_buffer.push(package);
         while let Some(current_package) = self.unit_propagation_buffer.pop() {
@@ -186,12 +186,11 @@ impl<DP: DependencyProvider> State<DP> {
                 }
             }
             if let Some(incompat_id) = conflict_id {
-                let (package_almost, root_cause) =
-                    self.conflict_resolution(incompat_id)
-                        .map_err(|terminal_incompat_id| {
-                            self.build_derivation_tree(terminal_incompat_id)
-                        })?;
-                root_causes.push((package, root_cause));
+                let (package_almost, root_cause) = self
+                    .conflict_resolution(incompat_id, &mut satisfier_causes)
+                    .map_err(|terminal_incompat_id| {
+                        self.build_derivation_tree(terminal_incompat_id)
+                    })?;
                 self.unit_propagation_buffer.clear();
                 self.unit_propagation_buffer.push(package_almost);
                 // Add to the partial solution with incompat as cause.
@@ -207,16 +206,46 @@ impl<DP: DependencyProvider> State<DP> {
             }
         }
         // If there are no more changed packages, unit propagation is done.
-        Ok(root_causes)
+        Ok(satisfier_causes)
     }
 
     /// Return the root cause or the terminal incompatibility.
     /// CF <https://github.com/dart-lang/pub/blob/master/doc/solver.md#unit-propagation>
+    ///
+    /// Usually by the time we have a conflict `unit_propagation` has done a lot of work.
+    /// So the actual conflict we find is important, but not particularly actionable.
+    /// It says something like "the dependency on package X and the dependency on package Y are incompatible".
+    /// To make it actionable we want to track it back to decisions that made the dependency required.
+    /// "The decision on B is incompatible with the decision on C,
+    ///  because unit propagation from just those decisions will lead to the conflict about X and Y"
+    /// is much more actionable, backtrack until one of those decisions can be revisited.
+    /// To make a practical, we really only need one of the terms to be a decision.
+    /// We may as well leave the other terms general. Something like
+    /// "the dependency on the package X is incompatible with the decision on C" tends to work out pretty well.
+    /// Then if A turns out to also have a dependency on X the resulting root cause is still useful.
+    /// Of course, this is more heuristics than science. If the output is too general, then `unit_propagation` will
+    /// handle the confusion by calling us again with the next most specific conflict it comes across.
+    /// If the output is to specific, then the outer `solver` loop will eventually end up calling us again
+    /// until all possibilities are enumerated.
+    ///
+    /// This function combines incompatibilities with things that make the problem inevitable to end up with a
+    /// more useful incompatibility. For the correctness of the PubGrub algorithm only the final output is required.
+    /// By banning the final output, unit propagation will prevent the intermediate steps from occurring again,
+    /// at least prevent the exact same way. However, the statistics collected for `prioritize`may want
+    /// to analyze those intermediate steps. For example we might start with "there is no version 1 of Z",
+    /// and `conflict_resolution` may be able to determine that "that was inevitable when we picked version 1 of X"
+    /// which was inevitable when picked W and ... and version 1 of B, which was depended on by version 1 of A.
+    /// Therefore the root cause may simplify all the way down to "we cannot pick version 1 of A".
+    /// This will prevent us going down this path again. However when we start looking at version 2 of A,
+    /// and discover that it depends on version 2 of B, we will want to prioritize the chain of intermediate steps
+    /// to confirm if it has a problem with the same shape.
+    /// The `satisfier_causes` argument keeps track of these intermediate steps so that the caller can use.
     #[allow(clippy::type_complexity)]
     #[cold]
     fn conflict_resolution(
         &mut self,
         incompatibility: IncompDpId<DP>,
+        satisfier_causes: &mut SmallVec<(Id<DP::P>, IncompDpId<DP>)>,
     ) -> Result<(Id<DP::P>, IncompDpId<DP>), IncompDpId<DP>> {
         let mut current_incompat_id = incompatibility;
         let mut current_incompat_changed = false;
@@ -240,6 +269,7 @@ impl<DP: DependencyProvider> State<DP> {
                             previous_satisfier_level,
                         );
                         log::info!("backtrack to {:?}", previous_satisfier_level);
+                        satisfier_causes.push((package, current_incompat_id));
                         return Ok((package, current_incompat_id));
                     }
                     SatisfierSearch::SameDecisionLevels { satisfier_cause } => {
@@ -251,6 +281,7 @@ impl<DP: DependencyProvider> State<DP> {
                         );
                         log::info!("prior cause: {}", prior_cause.display(&self.package_store));
                         current_incompat_id = self.incompatibility_store.alloc(prior_cause);
+                        satisfier_causes.push((package, current_incompat_id));
                         current_incompat_changed = true;
                     }
                 }
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -2,8 +2,10 @@ use std::cmp::Reverse;
 use std::collections::BTreeMap;
 use std::convert::Infallible;
 
-use crate::solver::PackageResolutionStatistics;
-use crate::{Dependencies, DependencyConstraints, DependencyProvider, Map, Package, VersionSet};
+use crate::{
+    Dependencies, DependencyConstraints, DependencyProvider, Map, Package,
+    PackageResolutionStatistics, VersionSet,
+};
 
 /// A basic implementation of [DependencyProvider].
 #[derive(Debug, Clone, Default)]
@@ -102,15 +104,15 @@ impl<P: Package, VS: VersionSet> DependencyProvider for OfflineDependencyProvide
         range: &Self::VS,
         package_statistics: &PackageResolutionStatistics,
     ) -> Self::Priority {
-        (
-            package_statistics.conflict_count(),
-            Reverse(
-                self.dependencies
-                    .get(package)
-                    .map(|versions| versions.keys().filter(|v| range.contains(v)).count())
-                    .unwrap_or(0),
-            ),
-        )
+        let version_count = self
+            .dependencies
+            .get(package)
+            .map(|versions| versions.keys().filter(|v| range.contains(v)).count())
+            .unwrap_or(0);
+        if version_count == 0 {
+            return (u32::MAX, Reverse(0));
+        }
+        (package_statistics.conflict_count(), Reverse(version_count))
     }
 
     #[inline]
--- a/src/solver.rs
+++ b/src/solver.rs
@@ -126,8 +126,8 @@ pub fn resolve<DP: DependencyProvider>(
             "unit_propagation: {:?} = '{}'",
             &next, state.package_store[next]
         );
-        let root_causes = state.unit_propagation(next)?;
-        for (affected, incompat) in root_causes {
+        let satisfier_causes = state.unit_propagation(next)?;
+        for (affected, incompat) in satisfier_causes {
             conflict_tracker
                 .entry(affected)
                 .or_default()
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -71,7 +71,8 @@
 //! and [SemanticVersion] for versions.
 //! This may be done quite easily by implementing the three following functions.
 //! ```
-//! # use pubgrub::{DependencyProvider, Dependencies, SemanticVersion, Ranges, DependencyConstraints, Map, PackageResolutionStatistics};
+//! # use pubgrub::{DependencyProvider, Dependencies, SemanticVersion, Ranges,
+//! #               DependencyConstraints, Map, PackageResolutionStatistics};
 //! # use std::error::Error;
 //! # use std::borrow::Borrow;
 //! # use std::convert::Infallible;


src/provider.rs
```diff
@@ -92,16 +95,24 @@ impl<P: Package, VS: VersionSet> DependencyProvider for OfflineDependencyProvide
             .and_then(|versions| versions.keys().rev().find(|v| range.contains(v)).cloned()))
     }
 
-    type Priority = Reverse<usize>;
+    type Priority = (u32, Reverse<u32>);
```
x-hgg-x: Merging to a `u64` should improve performance.

examples/caching_dependency_provider.rs
```diff
@@ -57,8 +60,13 @@ impl<DP: DependencyProvider<M = String>> DependencyProvider for CachingDependenc
 
     type Priority = DP::Priority;
 
-    fn prioritize(&self, package: &DP::P, ranges: &DP::VS) -> Self::Priority {
-        self.remote_dependencies.prioritize(package, ranges)
+    fn prioritize(
+        &self,
+        package: &DP::P,
+        ranges: &DP::VS,
+        statis: &PackageResolutionStatistics,
```
x-hgg-x:
```suggestion
        stats: &PackageResolutionStatistics,
```

src/solver.rs
```diff
@@ -199,6 +200,46 @@ pub enum Dependencies<P: Package, VS: VersionSet, M: Eq + Clone + Debug + Displa
...
+
+impl PackageResolutionStatistics {
+    fn new<P: Package>(pid: Id<P>, conflict_count: &Map<Id<P>, u32>) -> Self {
+        Self {
+            discovery_order: pid.into_raw() as u32,
+            conflict_count: conflict_count.get(&pid).cloned().unwrap_or_default(),
+        }
+    }
+
+    /// The number of packages resolution new about the first time this package was mentioned.
```
x-hgg-x:
```suggestion
    /// The number of packages known by Pubgrub when this package was mentioned for the first time.
```

src/provider.rs
Eh2406: I would've expected LLVM to be able to optimize these to the same thing. Even if not,  I would not expect that  the actual `Ord`  calls are a significant portion of the runtime.  Are you able to actually measure an impact?

x-hgg-x: There was an measurable impact on solana before the improved priorisation because the `PriorityQueue` contained more elements so the `Ord` impls were called more, but now when using better priorisation the impact is negligible.

x-hgg-x: Comment can be resolved.

Eh2406:
@konstin what do you think? Do you want me to reduce a benchmark where this makes a big difference?

src/solver.rs
```diff
@@ -199,6 +200,46 @@ pub enum Dependencies<P: Package, VS: VersionSet, M: Eq + Clone + Debug + Displa
...
+    /// The number of times this package was involved in a conflict that caused a back jump.
+    ///
+    /// When resolution is proceeding normally, this value will stay at `0` for all packages.
+    /// Therefore, using this for prioritization will not affect the properties of simple cases
+    /// like checking a lock file.
+    /// Prioritizing based on this value directly allows the resolver to focus on the packages
+    /// it is having the most problems with.
+    ///
+    /// Note: The exact values depend on implementation details of PubGrub. So should not be relied on and may change.
+    pub fn conflict_count(&self) -> u32 {
```
konstin: This may be very useful for something i'm writing right now :eyes:

I'm tracking two different properties atm: Which package previously chosen package A is conflicting with the currently chosen package B, because i want to deprioritize A (it's always making the latter choices fail) while prioritizing B (if i pick B first, I have a good chance at getting the right version of A, without walking through all versions of B first), which requires some mechanism of knowing which is which.

src/solver.rs
```diff
@@ -199,6 +200,46 @@ pub enum Dependencies<P: Package, VS: VersionSet, M: Eq + Clone + Debug + Displa
...
+
+    /// The number of times this package was involved in a conflict that caused a back jump.
+    ///
+    /// When resolution is proceeding normally, this value will stay at `0` for all packages.
+    /// Therefore, using this for prioritization will not affect the properties of simple cases
+    /// like checking a lock file.
+    /// Prioritizing based on this value directly allows the resolver to focus on the packages
+    /// it is having the most problems with.
+    ///
+    /// Note: The exact values depend on implementation details of PubGrub. So should not be relied on and may change.
```
konstin: We should be clearer about what this value means. Something like:

> Every time an incompatibility is fulfilled, the package involved are marked as conflicting. PubGrub builds incompatibilities lazily, and there construction may change over time, so the value may also change across PubGrub releases.

src/solver.rs
```diff
@@ -199,6 +200,46 @@ pub enum Dependencies<P: Package, VS: VersionSet, M: Eq + Clone + Debug + Displa
...
+/// Some statistics about how much trouble the resolver has had with a package.
+pub struct PackageResolutionStatistics {
+    discovery_order: u32,
+    conflict_count: u32,
+}
+
+impl PackageResolutionStatistics {
+    fn new<P: Package>(pid: Id<P>, conflict_count: &Map<Id<P>, u32>) -> Self {
+        Self {
+            discovery_order: pid.into_raw() as u32,
```
konstin: If you manually call alloc in some order, e.g. for dependencies before your actual package for a pre-fetching related reason, would that break this order?

konstin (approved):
No idea how they'll benchmark on uv (as in, will we see a perf hit or not?), but i like giving users the information and the better default prioritize.

src/solver.rs
Eh2406: Technically it's the order alloc is called. So the actual values would certainly be affected by prefetching. However, prioritizes is only be called on  packages that are available for decision-making, so  the impact on prioritization should not be dramatically affected.

src/solver.rs
Eh2406: I don't know that a user would know what a `incompatibility` is. I'm not finding a good balance between too much detail and explaining why things could change.

zanieb: I think a consumer of the library probably needs to know what an incompatibility is, right?

mpizenberg: Not every user. But one that start playing with conflict counts should definitely IMO.

konstin: We can use a two-tiered explanantion: For basic users, this number is a measure for conflictingn-ess, the higher it is, the more you want to prioritize the package. For advanced users, it's helpful to know that this is the number of incompatibilities that were satisfied that this package was involved in, so a user can implement heuristics on top of it (as we want in uv), e.g. weighing of bounds-strictness vs. conflicting-ness.

x-hgg-x:
Counting the root cause instead of the first incompatibility is around 5% faster in CPU time for my implementation (https://github.com/x-hgg-x/pubgrub-bench) to resolve all crates (2789s -> 2663s), and 12% faster to resolve `solana` crates (385s -> 341s).

konstin:
I found something similar for uv, where I'm also using the incompatibility (https://github.com/astral-sh/pubgrub/pull/36/commits/04005a0afc33a5372bf5c567fe1fdaf717657fc4). I also had to skip incompatibilities that were only (package A, proxy package for A) to avoid our marker package type. It also looks like this is only counting conflicts in unit propagation, but not when adding dependencies, where we had the most conflicts in uv.

They aren't clashing, but we could also try building the conflict counts on top of [astral-sh/pubgrub@`04005a0` (#36)](https://github.com/astral-sh/pubgrub/pull/36/commits/04005a0afc33a5372bf5c567fe1fdaf717657fc4), which works successfully as conflict counter in uv. I can upstream that commit if it helps.

arielb1:
The latest solver change doesn't 100% fix the problem:
```
       Bucket:solana-sdk@1.x.y/dep:ed25519-dalek SemverPubgrub { norml: 1.18.23 | 1.18.25 | 1.18.26, pre: <E2><88><85> }  depends on Bucket:ed25519-dalek@1.x.y/default=true SemverPubgrub { norml: >=1.0.1, <1.0.2, pre: <E2><88><85> }
[2024-12-15T19:42:11Z INFO  pubgrub::internal::core] backtrack to DecisionLevel(2)
[2024-12-15T19:42:11Z INFO  pubgrub::solver] DP chose: Id::<Names>(46) = 'Bucket:solana-sdk@1.x.y/default=true' @ Some(Version { major: 1, minor: 18, patch: 22 })
[2024-12-15T19:42:11Z INFO  pubgrub::internal::partial_solution] add_decision: Id::<Names>(46) @ 1.18.22
[2024-12-15T19:42:11Z INFO  pubgrub::solver] unit_propagation: Id::<Names>(46) = 'Bucket:solana-sdk@1.x.y/default=true'
[2024-12-15T19:42:11Z INFO  pubgrub::solver] DP chose: Id::<Names>(279) = 'Bucket:ed25519-dalek@1.x.y/default' @ Some(Version { major: 1, minor: 0, patch: 0, pre: Prerelease("pre.3") })
[2024-12-15T19:42:11Z INFO  pubgrub::solver] add_decision (not first time): Id::<Names>(279) = 'Bucket:ed25519-dalek@1.x.y/default' @ 1.0.0-pre.3
[2024-12-15T19:42:11Z INFO  pubgrub::solver] unit_propagation: Id::<Names>(279) = 'Bucket:ed25519-dalek@1.x.y/default'
[2024-12-15T19:42:11Z INFO  pubgrub::solver] DP chose: Id::<Names>(280) = 'Bucket:ed25519-dalek@1.x.y/u64_backend' @ Some(Version { major: 1, minor: 0, patch: 0, pre: Prerelease("pre.3") })[2024-12-15T19:42:11Z INFO  pubgrub::solver] add_decision (not first time): Id::<Names>(280) = 'Bucket:ed25519-dalek@1.x.y/u64_backend' @ 1.0.0-pre.3
[2024-12-15T19:42:11Z INFO  pubgrub::solver] unit_propagation: Id::<Names>(280) = 'Bucket:ed25519-dalek@1.x.y/u64_backend'
[2024-12-15T19:42:11Z INFO  pubgrub::solver] DP chose: Id::<Names>(281) = 'Bucket:ed25519-dalek@1.x.y/std' @ Some(Version { major: 1, minor: 0, patch: 0, pre: Prerelease("pre.3") })
[2024-12-15T19:42:11Z INFO  pubgrub::solver] add_decision (not first time): Id::<Names>(281) = 'Bucket:ed25519-dalek@1.x.y/std' @ 1.0.0-pre.3
[2024-12-15T19:42:11Z INFO  pubgrub::solver] unit_propagation: Id::<Names>(281) = 'Bucket:ed25519-dalek@1.x.y/std'
[2024-12-15T19:42:11Z INFO  pubgrub::solver] DP chose: Id::<Names>(284) = 'Bucket:ed25519-dalek@1.x.y/rand' @ Some(Version { major: 1, minor: 0, patch: 0, pre: Prerelease("pre.3") })
[2024-12-15T19:42:11Z INFO  pubgrub::solver] add_decision (not first time): Id::<Names>(284) = 'Bucket:ed25519-dalek@1.x.y/rand' @ 1.0.0-pre.3
[2024-12-15T19:42:11Z INFO  pubgrub::solver] unit_propagation: Id::<Names>(284) = 'Bucket:ed25519-dalek@1.x.y/rand'
[2024-12-15T19:42:11Z INFO  pubgrub::solver] DP chose: Id::<Names>(286) = 'Bucket:ed25519-dalek@1.x.y/dep:rand' @ Some(Version { major: 1, minor: 0, patch: 0, pre: Prerelease("pre.3") })
[2024-12-15T19:42:11Z INFO  pubgrub::solver] add_decision (not first time): Id::<Names>(286) = 'Bucket:ed25519-dalek@1.x.y/dep:rand' @ 1.0.0-pre.3
[2024-12-15T19:42:11Z INFO  pubgrub::solver] unit_propagation: Id::<Names>(286) = 'Bucket:ed25519-dalek@1.x.y/dep:rand'
[2024-12-15T19:42:11Z INFO  pubgrub::solver] DP chose: Id::<Names>(355) = 'Bucket:solana-sdk@1.x.y/default' @ Some(Version { major: 1, minor: 18, patch: 22 })
[2024-12-15T19:42:11Z INFO  pubgrub::internal::partial_solution] add_decision: Id::<Names>(355) @ 1.18.22
[2024-12-15T19:42:11Z INFO  pubgrub::solver] unit_propagation: Id::<Names>(355) = 'Bucket:solana-sdk@1.x.y/default'
[2024-12-15T19:42:11Z INFO  pubgrub::solver] DP chose: Id::<Names>(356) = 'Bucket:solana-sdk@1.x.y/full' @ Some(Version { major: 1, minor: 18, patch: 22 })
[2024-12-15T19:42:11Z INFO  pubgrub::internal::partial_solution] add_decision: Id::<Names>(356) @ 1.18.22
[2024-12-15T19:42:11Z INFO  pubgrub::solver] unit_propagation: Id::<Names>(356) = 'Bucket:solana-sdk@1.x.y/full'
[2024-12-15T19:42:11Z INFO  pubgrub::solver] DP chose: Id::<Names>(357) = 'Bucket:solana-sdk@1.x.y/byteorder' @ Some(Version { major: 1, minor: 18, patch: 22 })
[2024-12-15T19:42:11Z INFO  pubgrub::internal::partial_solution] add_decision: Id::<Names>(357) @ 1.18.22
[2024-12-15T19:42:11Z INFO  pubgrub::solver] unit_propagation: Id::<Names>(357) = 'Bucket:solana-sdk@1.x.y/byteorder'
[2024-12-15T19:42:11Z INFO  pubgrub::solver] DP chose: Id::<Names>(358) = 'Bucket:solana-sdk@1.x.y/ed25519-dalek-bip32' @ Some(Version { major: 1, minor: 18, patch: 22 })
[2024-12-15T19:42:11Z INFO  pubgrub::internal::partial_solution] add_decision: Id::<Names>(358) @ 1.18.22
[2024-12-15T19:42:11Z INFO  pubgrub::solver] unit_propagation: Id::<Names>(358) = 'Bucket:solana-sdk@1.x.y/ed25519-dalek-bip32'
[2024-12-15T19:42:11Z INFO  pubgrub::solver] DP chose: Id::<Names>(359) = 'Bucket:solana-sdk@1.x.y/ed25519-dalek' @ Some(Version { major: 1, minor: 18, patch: 22 })
[2024-12-15T19:42:11Z INFO  pubgrub::internal::partial_solution] add_decision: Id::<Names>(359) @ 1.18.22
[2024-12-15T19:42:11Z INFO  pubgrub::solver] unit_propagation: Id::<Names>(359) = 'Bucket:solana-sdk@1.x.y/ed25519-dalek'
[2024-12-15T19:42:11Z INFO  pubgrub::solver] DP chose: Id::<Names>(373) = 'Bucket:solana-sdk@1.x.y/dep:ed25519-dalek' @ Some(Version { major: 1, minor: 18, patch: 22 })
[2024-12-15T19:42:11Z INFO  pubgrub::internal::partial_solution] not adding Id::<Names>(373) @ 1.18.22 because of its dependencies
[2024-12-15T19:42:11Z INFO  pubgrub::solver] unit_propagation: Id::<Names>(373) = 'Bucket:solana-sdk@1.x.y/dep:ed25519-dalek'
[2024-12-15T19:42:11Z INFO  pubgrub::internal::core] Start conflict resolution because incompat satisfied:
       Bucket:solana-sdk@1.x.y/dep:ed25519-dalek SemverPubgrub { norml: 1.18.22 | 1.18.23 | 1.18.25 | 1.18.26, pre: <E2><88><85> }  depends on Bucket:ed25519-dalek@1.x.y/default=true SemverPubgrub { norml: >=1.0.1, <1.0.2, pre: <E2><88><85> }
```

I would expect it to be able to prioritize this chain without "wandering":
```
solana-sdk@1.x.y/default=true -> Bucket:solana-sdk@1.x.y/default -> solana-sdk@1.x.y/full -> solana-sdk@1.x.y/ed25519-dalek -> solana-sdk@1.x.y/dep:ed25519-dalek
```

The "root cause" it finds is
```
[2024-12-15T19:44:45Z INFO  pubgrub::solver] DP chose: Id::<Names>(39) = 'Bucket:ed25519-dalek@1.x.y/default=true'
[2024-12-15T19:44:45Z INFO  pubgrub::solver] DP chose: Id::<Names>(46) = 'Bucket:solana-sdk@1.x.y/default=true'
```

Which is insufficient since it could find the rest of the packages in the chain to load them.

Eh2406:
Having spent the morning staring at the logs @arielb1 found. It's (now) clear that one of the problems here is a difference in perspective from  before and after things are "loaded". Once a conflict is found, we are in the "after loading" perspective where `Bucket:solana-sdk@1.x.y/default=true is incompatible with the prior decision for Bucket:ed25519-dalek@1.x.y/default=true`  is an obvious implication of `Bucket:solana-sdk@1.x.y/default=true  depends on ... depends on Bucket:ed25519-dalek@1.x.y/default=true`. Because the entire chain is built out of singleton  requirements (that can only match one thing), `unit_propagation` would be able to see that they are the same. Therefore, `root_cause` is generated as the shorter and clearer version.  Similarly, because wearing the "after loaded" perspective we can  see that we need to do one back jump all the way to `Bucket:solana-sdk@1.x.y/default=true`, instead of back jumping one (inevitable) decision at a time. But after that background, we are now looking at a new version of `Bucket:solana-sdk@1.x.y/default=true` and so are in the "before loading" situation. So the simplified version does not have enough information for us to tell which of these not yet evaluated packages actually reduce down to the problem.

That would suggest that counting the packages that were  involved in each step of the loop in `conflict_resolution` would be more effective.  It didn't work yesterday, but none of my performance numbers are looking the same as yesterday. So I should try again.

src/internal/core.rs
```diff
@@ -202,34 +208,38 @@ impl<DP: DependencyProvider> State<DP> {
...
+                    return Ok((package, current_incompat_id));
+                }
+                SatisfierSearch::SameDecisionLevels { satisfier_cause } => {
+                    let prior_cause = Incompatibility::prior_cause(
+                        current_incompat_id,
+                        satisfier_cause,
+                        package,
+                        &self.incompatibility_store,
+                    );
+
```
Eh2406: This change needs a comment about what it is and why it is here.

Eh2406:
Rebased, which removed most of the code.  Added in extensively wordy comment.  I'm not so sure about the quality of this comment. It may have too many words and say too little.  Edits are welcome.

Overall performance,  measured with `cargo r -r -- --with-solana -m pub -t 36 --filter solana` went from ` Pub CPU time: 29.64min` to `21.98min`. `--filter solana-archiver-lib` went from `2.78min` to `2.46min`.

konstin:
I'm seeing a massive speedup with solana-archiver-lib. I'm testing with 8 threads since 16 overflow with dev.

**Before** (dev, 3bef331564078fa0ece1c850518d41efe2c22b18)
```
$ cargo run -r -- -m pub --with-solana --filter solana-archiver-lib -t 8
Running in mode Pub on 8 rayon threads.
        index commit hash: 82086e46740d7a9303216bfac093e7268a95121f
        index commit time: 2024-11-30T18:18:14Z
               index size: 32
          solana in index: 32
             Pub CPU time:   257.76s ==   4.30min
           Cargo CPU time: skipped
Cargo check lock CPU time: skipped
  Pub check lock CPU time: skipped
                Wall time:    37.05s ==   0.62min
```

**After** (PR, e08db340950e3da69ba1b0e2eaa0449300b59625)
```
$ cargo run -r -- -m pub --with-solana --filter solana-archiver-lib -t 8
Running in mode Pub on 8 rayon threads.
        index commit hash: 82086e46740d7a9303216bfac093e7268a95121f
        index commit time: 2024-11-30T18:18:14Z
               index size: 32
          solana in index: 32
             Pub CPU time:   143.82s ==   2.40min
           Cargo CPU time: skipped
Cargo check lock CPU time: skipped
  Pub check lock CPU time: skipped
                Wall time:    20.17s ==   0.34min
```

It's an improvement in both measurements, thought I'm getting a very different speedup.

Counterintuitively, the should cancel count is way worse while the runtime performance improves. Is there a good non-walltime metric?

image


Patch:

```rust
    type Priority = (u32, Reverse<usize>);

    fn prioritize(
        &self,
        package: &Names<'c>,
        range: &RcSemverPubgrub,
        package_conflicts_counts: &PackageResolutionStatistics,
    ) -> Self::Priority {
        let matches_count = match package {
            Names::Links(_name) => {
                // PubGrub automatically handles when any requirement has no overlap. So this is only deciding a importance of picking the version:
                //
                // - If it only matches one thing, then adding the decision with no additional dependencies makes no difference.
                // - If it can match more than one thing, and it is entirely equivalent to picking the packages directly which would make more sense to the users.
                //
                // So only rubberstamp links attributes when all other decisions are made, by setting the priority as low as it will go.
                usize::MAX
            }

            Names::Wide(_, req, _, _) => self.count_wide_matches(range, &package.crate_(), req),
            Names::WideFeatures(_, req, _, _, _) | Names::WideDefaultFeatures(_, req, _, _) => self
                .count_wide_matches(range, &package.crate_(), req)
                .saturating_add(1),

            Names::Bucket(_, _, _) => self.count_matches(range, &package.crate_()),
            Names::BucketFeatures(_, _, _) | Names::BucketDefaultFeatures(_, _) => self
                .count_matches(range, &package.crate_())
                .saturating_add(1),
        };
        (
            package_conflicts_counts.conflict_count(),
            Reverse(matches_count),
        )
    }
```

(I'll still have to give the updated PR a proper read)

examples/caching_dependency_provider.rs
```diff
@@ -62,12 +62,11 @@ impl<DP: DependencyProvider<M = String>> DependencyProvider for CachingDependenc
...
         &self,
-        package: &Self::P,
-        range: &Self::VS,
-        package_statistics: &PackageResolutionStatistics,
+        package: &DP::P,
+        ranges: &DP::VS,
+        stats: &PackageResolutionStatistics,
     ) -> Self::Priority {
-        self.remote_dependencies
-            .prioritize(package, range, package_statistics)
```
Eh2406: Don't know how this snuck in. I will have to back it out.

Eh2406:
> I'm seeing a massive speedup with solana-archiver-lib. I'm testing with 8 threads since 16 overflow with dev.

Has memory consumption gotten worse? 16 was working for you before, right? I will need to come back to the oddly high memory usage.

> It's an improvement in both measurements, thought I'm getting a very different speedup.

I mostly watch the `CPU time`. But it is odd how different % improved is between my computer and yours.

> Counterintuitively, the should cancel count is way worse while the runtime performance improves. Is there a good non-walltime metric?

I have not found one yet. Sorry.

> Patch

And I forgot to push the code I was running before I logged off the work computer. Logged back on to push this for you. I was using https://github.com/Eh2406/pubgrub-crates-benchmark/commit/80048df733a601d1dd3aad683497fce995a89719

Eh2406:
`cargo run -r -- -m pub --with-solana --filter solana-archiver-lib -t 8`

with 3bef331564078fa0ece1c850518d41efe2c22b18 with https://github.com/Eh2406/pubgrub-crates-benchmark/commit/80048df733a601d1dd3aad683497fce995a89719

```
             Pub CPU time:   143.91s ==   2.40min
                Wall time:    18.97s ==   0.32min
```

with e08db340950e3da69ba1b0e2eaa0449300b59625 with https://github.com/Eh2406/pubgrub-crates-benchmark/commit/80048df733a601d1dd3aad683497fce995a89719

```
             Pub CPU time:   126.50s ==   2.11min
                Wall time:    16.67s ==   0.28min
```

with 3bef331564078fa0ece1c850518d41efe2c22b18 and your patch

```
             Pub CPU time:   437.86s ==   7.30min
                Wall time:    63.07s ==   1.05min
```

with e08db340950e3da69ba1b0e2eaa0449300b59625 with your patch

```
             Pub CPU time:   251.98s ==   4.20min
                Wall time:    35.26s ==   0.59min
```

konstin:
With https://github.com/Eh2406/pubgrub-crates-benchmark/commit/80048df733a601d1dd3aad683497fce995a89719 i'm getting the same numbers too now :)

src/internal/core.rs
```diff
@@ -190,16 +189,46 @@ impl<DP: DependencyProvider> State<DP> {
...
+    /// By banning the final output, unit propagation will prevent the intermediate steps from occurring again,
+    /// at least prevent the exact same way. However, the statistics collected for `prioritize`may want
+    /// to analyze those intermediate steps. For example we might start with "there is no version 1 of Z",
+    /// and `conflict_resolution` may be able to determine that "that was inevitable when we picked version 1 of X"
+    /// which was inevitable when picked W and ... and version 1 of B, which was depended on by version 1 of A.
+    /// Therefore the root cause may simplify all the way down to "we cannot pick version 1 of A".
+    /// This will prevent us going down this path again. However when we start looking at version 2 of A,
+    /// and discover that it depends on version 2 of B, we will want to prioritize the chain of intermediate steps
+    /// to confirm if it has a problem with the same shape.
+    /// The `satisfier_causes` argument keeps track of these intermediate steps so that the caller can use.
```
konstin: Tried some editing:

```markdown
Return the root cause or the terminal incompatibility. CF
<https://github.com/dart-lang/pub/blob/master/doc/solver.md#unit-propagation>

When we found a conflict, we want to learn as much as possible from it, to avoid making (or
keeping) decisions that will be rejected. Say we found that the decision for X and the
decision for Y are incompatible. We may find that the decisions on earlier packages B and C
require us to make incompatible decisions on X and Y, so we backtrack until either B or C
can be revisited. To make it practical, we really only need one of the terms to be a
decision. We may as well leave the other terms general. Something like "the dependency on
the package X is incompatible with the decision on C" tends to work out pretty well. Then if
A turns out to also have a dependency on X the resulting root cause is still useful. Of
course, this is more heuristics than science. If the output is too general, then
`unit_propagation` will handle the confusion by calling us again with the next most specific
conflict it comes across. If the output is too specific, then the outer `solver` loop will
eventually end up calling us again until all possibilities are enumerated.

To end up with a more useful incompatibility, this function combines incompatibilities into
derivations. Fulfilling this derivation implies the later conflict. By banning it, we
prevent the intermediate steps from occurring again, at least in the exact same way.
However, the statistics collected for `prioritize` may want to analyze those intermediate
steps. For example we might start with "there is no version 1 of Z", and
`conflict_resolution` may be able to determine that "that was inevitable when we picked
version 1 of X" which was inevitable when we picked W and so on, until version 1 of B, which
was depended on by version 1 of A. Therefore the root cause may simplify all the way down to
"we cannot pick version 1 of A". This will prevent us going down this path again. However
when we start looking at version 2 of A, and discover that it depends on version 2 of B, we
will want to prioritize the chain of intermediate steps to check if it has a problem with
the same shape. The `satisfier_causes` argument keeps track of these intermediate steps so
that the caller can use them for prioritization.
```

> Then if A turns out to also have a dependency on X the resulting root cause is still
> useful.

How will we apply the root cause?

> If the output is too general, then `unit_propagation` will handle the confusion by calling
> us again with the next most specific conflict it comes across. If the output is too
> specific, then the outer `solver` loop will eventually end up calling us again until all
> possibilities are enumerated.

In which direction are "too general" and "too specific" to be understood here?

src/internal/core.rs
```diff
@@ -190,16 +189,46 @@ impl<DP: DependencyProvider> State<DP> {
...
     #[allow(clippy::type_complexity)]
     #[cold]
     fn conflict_resolution(
         &mut self,
         incompatibility: IncompDpId<DP>,
+        satisfier_causes: &mut SmallVec<(Id<DP::P>, IncompDpId<DP>)>,
```
konstin, line 247: If this turns out to show up in anyone's profile while they don't use the output, we can replace it by a callback, but for now :shipit:

Eh2406:
Feel free to merge  when you had a chance to review the code.

konstin (approved):

konstin:
Code looks good, we can iterate over the comment now or later, it's not blocking merging

src/internal/core.rs
Eh2406: Overall I like the edits. I have a nit about

> Say we found that the decision for X and the decision for Y are incompatible.

I'm going to quibble about the term `decision` in this sentence. In my brain `decision`  implies something that became constrained because of a call to `choose_version`/`add_decision`. Because we check for conflicts in `add_version` simple cases end up handled there. We generally do not end up here with a conflict that is already in terms of decisions. Instead  We start with something abstract like `X (in some range) depends on Y (in some range)`. It's the job of this function to peel back the layers, figure out why the partial solution has an assignment (which is not necessarily decision) that is disjoint with the mentioned range for Y and why the partial solution has an assignment that is a subset of the mentioned range for X.

> How will we apply the root cause?

When we discover that A  has a dependency on X `unit_propagation` will rescan the root cause 
`"the dependency on the package X is incompatible with the decision on C"`, and remove that version of C from the available  range stored in partial solution.

> In which direction are "too general" and "too specific" to be understood here?

I intended for to specific to mean referring to exact decided versions. "B == 123 is incompatible with C == 456" would be very specific. Extremely actionable, `unit_propagation`  needs to do very little work to figure out if it's relevant. But also not likely to  be reusable. We are likely to end up with `O(# versions of B * # versions of C)`  of these incompatibilities laying around. "any version of X is incompatible with any version of Y"  would be very general. There will only ever need to be one, and it will be relevant  and useful any time X or Y  come up again. But, it takes a lot of work to figure out how it relates to decisions or where we need to backtrack to.

Obviously the wording of the paragraphs needs  work, thank you for helping me with it!

Eh2406:
I accidentally did some benchmarking without this commit because I forgot we were still discussing comment.  I took your language, and changed a few little things. I'm totally open to a follow-up PR to improve the comment,  But for now I'm going to merge so I don't mess up my numbers again.

src/provider.rs
```diff
@@ -102,15 +104,15 @@ impl<P: Package, VS: VersionSet> DependencyProvider for OfflineDependencyProvide
...
-                    .map(|versions| versions.keys().filter(|v| range.contains(v)).count())
-                    .unwrap_or(0),
-            ),
-        )
+        let version_count = self
+            .dependencies
+            .get(package)
+            .map(|versions| versions.keys().filter(|v| range.contains(v)).count())
+            .unwrap_or(0);
+        if version_count == 0 {
```
konstin, line 112: Does 0 mean that there is a conflict?

Eh2406: Yes, if we have 0 `version_count` then `get_dependencies` will be `Unavailable`.

PR #294, opened 2024-12-11
Improve documentation of partial solution

konstin:
Improve the docstrings for the partial solution and rename some fields to be easier to understand.

Thanks to @Eh2406 for providing the context for these fields!

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -21,7 +21,7 @@ pub(crate) struct State<DP: DependencyProvider> {
     #[allow(clippy::type_complexity)]
     incompatibilities: Map<Id<DP::P>, Vec<IncompDpId<DP>>>,
 
-    /// As optimization, store the ids of incompatibilities that are already contradicted.
+    /// As an optimization, store the ids of incompatibilities that are already contradicted.
     ///
     /// For each one keep track of the decision level when it was found to be contradicted.
     /// These will stay contradicted until we have backtracked beyond its associated decision level.
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -21,7 +21,8 @@ pub(crate) struct State<DP: DependencyProvider> {
     #[allow(clippy::type_complexity)]
     incompatibilities: Map<Id<DP::P>, Vec<IncompDpId<DP>>>,
 
-    /// Store the ids of incompatibilities that are already contradicted.
+    /// As optimization, store the ids of incompatibilities that are already contradicted.
+    ///
     /// For each one keep track of the decision level when it was found to be contradicted.
     /// These will stay contradicted until we have backtracked beyond its associated decision level.
     contradicted_incompatibilities: Map<IncompDpId<DP>, DecisionLevel>,
--- a/src/internal/partial_solution.rs
+++ b/src/internal/partial_solution.rs
@@ -66,7 +66,7 @@ pub(crate) struct PartialSolution<DP: DependencyProvider> {
     /// The max heap allows quickly `pop`ing the highest priority package.
     prioritized_potential_packages:
         PriorityQueue<Id<DP::P>, DP::Priority, BuildHasherDefault<FxHasher>>,
-    /// Track whether we have never backtracked to enable fast path optimizations.
+    /// Whether we have never backtracked, to enable fast path optimizations.
     has_ever_backtracked: bool,
 }
 
--- a/src/internal/partial_solution.rs
+++ b/src/internal/partial_solution.rs
@@ -35,7 +35,9 @@ pub(crate) struct PartialSolution<DP: DependencyProvider> {
     /// Store for all known package decisions and package derivations.
     ///
     /// "assignment" refers to both packages with decisions and package with only derivations and
-    /// no decision yet.
+    /// no decision yet. We combine this in a single index map, where different sections (of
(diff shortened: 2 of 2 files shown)


konstin:
That's a dubious looking perf regression

src/internal/core.rs
```diff
@@ -21,7 +21,8 @@ pub(crate) struct State<DP: DependencyProvider> {
     #[allow(clippy::type_complexity)]
     incompatibilities: Map<Id<DP::P>, Vec<IncompDpId<DP>>>,
 
-    /// Store the ids of incompatibilities that are already contradicted.
+    /// As optimization, store the ids of incompatibilities that are already contradicted.
```
zanieb:
```suggestion
    /// As an optimization, store the ids of incompatibilities that are already contradicted.
```

src/internal/partial_solution.rs
```diff
@@ -30,24 +30,43 @@ impl DecisionLevel {
 #[derive(Clone, Debug)]
 pub(crate) struct PartialSolution<DP: DependencyProvider> {
     next_global_index: u32,
+    /// The number of decisions that have been made, equal to the number of packages with decisions.
     current_decision_level: DecisionLevel,
```
zanieb, line 34: "packages with decisions" being "packages with a decided version"?

src/internal/partial_solution.rs
```diff
@@ -30,24 +30,43 @@ impl DecisionLevel {
...
-    ///    changed since the last time `prioritize` has been called. Within this range there is no sorting.
-    /// 3. `[changed_this_decision_level..]` Contains all packages that **have** had there assignments changed since
-    ///    the last time `prioritize` has been called. The inverse is not necessarily true, some packages in the range
-    ///    did not have a change. Within this range there is no sorting.
+    /// Store for all known package decisions and package derivations.
+    ///
+    /// "assignment" refers to both packages with decisions and package with only derivations and
+    /// no decision yet.
+    ///
+    /// `[..current_decision_level]`: Packages that have had a decision made, sorted by the
```
zanieb, line 42: Can you provide a little more context before jumping into these ranges? Like set the stage for the format of this commentary?

src/internal/partial_solution.rs
```diff
@@ -30,24 +30,43 @@ impl DecisionLevel {
...
-    /// to its `Priority`. But, it also maintains a max heap of packages by `Priority` order.
+    /// Index into `package_assignments` to decide which packages need to be re-prioritized.
+    prioritize_decision_level: usize,
+    /// The undecided packages order by their `Priority`.
+    ///
+    /// The max heap allows quickly `pop`ing the highest priority package.
     prioritized_potential_packages:
         PriorityQueue<Id<DP::P>, DP::Priority, BuildHasherDefault<FxHasher>>,
-    changed_this_decision_level: usize,
+    /// Track whether we have never backtracked to enable fast path optimizations.
```
zanieb:
```suggestion
    /// Whether we have never backtracked, to enable fast path optimizations.
```

src/internal/partial_solution.rs
```diff
@@ -30,24 +30,43 @@ impl DecisionLevel {
...
+    /// "assignment" refers to both packages with decisions and package with only derivations and
+    /// no decision yet.
+    ///
+    /// `[..current_decision_level]`: Packages that have had a decision made, sorted by the
+    /// `decision_level`. The section is can be seen as the partial solution, it contains a
+    /// mapping from package name to decided version. The sorting makes it very efficient to
+    /// extract the solution, and to backtrack to a particular decision level. The
+    /// `AssignmentsIntersection` is always a `Decision`.
+    ///
+    /// `[changed_this_decision_level..]`: Packages that are dependencies of some other package,
```
zanieb: `changed_this_decision_level` looks removed?

src/internal/partial_solution.rs
mpizenberg: yes

mpizenberg:
> That's a dubious looking perf regression

the only two structural changes are a tuple converted to a struct with field names, and another struct which has the order of its fields changed. Could it be byte alignement due to the new fields order?

konstin:
I can't reproduce this locally, and definitely not in the 6% range:

```
sudoku-hard             time:   [564.46 µs 564.59 µs 564.72 µs]
                        change: [-0.6899% -0.3674% -0.0552%] (p = 0.02 < 0.05)
                        Change within noise threshold.
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) low mild
  8 (8.00%) high severe
```

Eh2406:
> Could it be byte alignement due to the new fields order?

Unless a structure is marked `repr(c)` rust is allowed to reorder the fields to best pack alignment. So it would be very odd if changing the definition order led to a significant difference in size. It could affect the order fields are dropped, or the implementation of Hash or Eq. I have seen the order effect Eq in such a way that performance changed, but I don't think that applies here.

Overall I am happy with this change, but leave final decision on wordsmithing questions to @mpizenberg.

Eh2406 (approved):

mpizenberg (approved):
It’s a clear improvement in my point of view. Thank you!

konstin:
Had some trouble rebasing but i think everything is in.

PR #295, opened 2024-12-11
Return `Id<DP::P>` over `&Id<DP::P>`

konstin:
It's more efficient to return a u32 than to return a reference, and it avoids a borrow when returning it.

### DIFF
--- a/src/internal/incompatibility.rs
+++ b/src/internal/incompatibility.rs
@@ -246,8 +246,10 @@ impl<P: Package, VS: VersionSet, M: Eq + Clone + Debug + Display> Incompatibilit
     }
 
     /// Iterate over packages.
-    pub(crate) fn iter(&self) -> impl Iterator<Item = (&Id<P>, &Term<VS>)> {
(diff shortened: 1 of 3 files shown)


src/internal/incompatibility.rs
```diff
@@ -246,8 +246,10 @@ impl<P: Package, VS: VersionSet, M: Eq + Clone + Debug + Display> Incompatibilit
     }
 
     /// Iterate over packages.
-    pub(crate) fn iter(&self) -> impl Iterator<Item = (&Id<P>, &Term<VS>)> {
-        self.package_terms.iter()
+    pub(crate) fn iter(&self) -> impl Iterator<Item = (Id<P>, &Term<VS>)> {
```
konstin, line 249: The plan is to expose this iterator from unit propagation and adding dependencies to allow tracking conflicts in the caller.

zanieb (approved):
Makes sense to me. Probably worth waiting for @Eh2406's review too.

PR #301, opened 2024-12-20
Allow `Ranges::contains` to accept borrows, e.g. `&str` for `Ranges<String>`

konstin:
This PR upstreams https://github.com/astral-sh/pubgrub/pull/35.

---

This PR borrows a trick from
[HashMap](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.contains_key) to enable users to pass (e.g.) `&str` to `Ranges::contains`, given `Ranges<String>`.

### DIFF
--- a/version-ranges/src/lib.rs
+++ b/version-ranges/src/lib.rs
@@ -229,7 +229,11 @@ impl<V: Ord> Ranges<V> {
     }
 
     /// Returns true if self contains the specified value.
-    pub fn contains(&self, version: &V) -> bool {
+    pub fn contains<Q>(&self, version: &Q) -> bool
+    where
+        V: Borrow<Q>,
+        Q: ?Sized + PartialOrd,
+    {
         self.segments
             .binary_search_by(|segment| {
                 // We have to reverse because we need the segment wrt to the version, while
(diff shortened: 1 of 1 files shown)


Eh2406 (approved):

Eh2406:
What is our process for releasing a new version of this crate?

konstin:
It's updating `Changelog.md`, bumping the version number and running `cargo publish` locally.

Eh2406:
Two thoughts about this PR.

1.  We should probably add a test, even if it's just to compile test,  for some basic use cases unlocked by this PR.
2.  Do we want hash maps API here? Seems like what we really need is `Q: PartialOrd<V>` not `V: Borrow<Q>`?

Eh2406:
> Do we want hash maps API here? Seems like what we really need is `Q: PartialOrd<V>` not `V: Borrow<Q>`?

`Q: PartialOrd<V>` gets everything compile, and looks quite nice. Unfortunately it doesn't work for 
```
can't compare `str` with `std::string::String`
the trait `PartialOrd<std::string::String>` is not implemented for `str`
the trait `PartialOrd` is implemented for `str`
for that trait implementation, expected `str`, found `std::string::String`
```
So moving on with my life.

konstin:
This works, but i don't know if it's better:

```rust
use std::ops::Deref;

struct Db<V>(Vec<V>);

impl<V> Db<V> {
    pub fn contains<'a, Q, CommonCmp>(&self, version: &'a Q) -> bool
    where
        // We deref both internal and external type to this common type that we use for comparison.
        CommonCmp: PartialOrd<CommonCmp> + ?Sized,
        V: Deref<Target = CommonCmp>,
        &'a Q: Deref<Target = CommonCmp>,
        Q: PartialOrd<CommonCmp> + ?Sized,
    {
        self.0
            .binary_search_by(|segment| segment.deref().partial_cmp(version.deref()).unwrap())
            // An equal interval is one that contains the version
            .is_ok()
    }
}

fn main() {
    let haystack = Db(vec!["1".to_string(), "2".to_string(), "3".to_string()]);
    let one = haystack.contains("1");
    println!("1: {}", one);
    let four = haystack.contains(&"4".to_string());
    println!("4: {}", four);
}
```

PR #307, opened 2025-01-22
Try codspeed walltime runners

konstin:
The docs say only python let's see if it indeed fails.


adriencaccia:
Hey @konstin, it indeed failed on our end 😅 
For the wall time runs to work in Rust, we first have to support the new runner mode in https://github.com/CodSpeedHQ/codspeed-rust

konstin:
Sorry - I didn't check the docs before asking for access, we'll try again when rust support launches.

Issue #308, opened 2025-01-22
v0.3 stable tracking issue


konstin:
The last release was mid 2021. Since then there were [significant changes](https://github.com/pubgrub-rs/pubgrub/compare/v0.2.1...dev). We should make a new stable release:

Todos:

- [x] Finish writing Changelog
- [x] Update the book https://github.com/pubgrub-rs/guide/pull/14
- [x] Check that the rustdoc is all up-to-date https://github.com/pubgrub-rs/pubgrub/pull/315
- [x] meta: triage issues
- [x] fix or close #239
- [x] fix the name of `choose_package_version` https://github.com/pubgrub-rs/pubgrub/issues/239#issuecomment-2228488646
- [x] merge into the release branch
- [x] mark that merged commit as a git hub release

I'll update this issue as we come closer to the release.

Eh2406:
cc https://github.com/pubgrub-rs/pubgrub/milestone/1

Eh2406:
Yesterday I read over the docks, and with your updates they now look ready for 0.3!
Today I read over the guide and found a couple of things that need to be updated before 0.3.
- [x] `multiple_versions.md` utilizes choose_package_with_fewest_versions which no longer exists.
- [x] same for `public_private.md`.
- [x] the examples in the  guide are also at  https://github.com/pubgrub-rs/advanced_dependency_providers which will need to be updated.

There are lots of sections that could probably be updated or improved, but should not be blockers. Most dramatically the "How can I contribute?".

Eh2406:
Do you want to do anything else with the book/change log or is what we have good enough? If it's good enough, that I think were clear to publish.

konstin:
The changelog could use a bit rephrasing (like "we basically rewrote this crate"), then it's ready to ship from my end.

konstin:
pubgrub v0.3.0 is released.

PR #309, opened 2025-01-22
(Pre)Release 0.3.0-alpha.1

konstin:
The last release was mid 2021. Since then there were [significant changes](https://github.com/pubgrub-rs/pubgrub/compare/v0.2.1...dev). To prepare for a new stable release, we start with an alpha release, both to get back into the habit of making releases and to give users an opportunity to try it out. 

Part of #308

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,54 @@
 
 All notable changes to this project will be documented in this file.
 
-## Unreleased [(diff)][unreleased-diff]
+## [0.3.0] UNRELEASED [(diff with 0.2.1)][unreleased-diff]
+
+PubGrub 0.3 improves the interfaces and speeds resolution significantly.
+
+All public interfaces are now in the root of the crate. 
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,54 @@
 
 All notable changes to this project will be documented in this file.
 
-## Unreleased [(diff)][unreleased-diff]
+## [0.3.0] UNRELEASED [(diff with 0.2.1)][unreleased-diff]
```
Eh2406, line 5: Somewhere we should document how were going to change the pre-release version number for breaking and non-breaking changes.

Eh2406:
Thank you for pushing this forward! I'm sorry I dropped the ball on making releases.

CHANGELOG.md
Eh2406: we're going to treat all prereleases of breaking changes, otherwise we might as well just make normal releases.

Eh2406 (approved):

PR #312, opened 2025-01-24
Update the docs for prioritize

konstin:
I've written the docs with focus on a new user that implements a dependency provider for the first time rather than with a focus on exact definitions.

### DIFF
--- a/src/solver.rs
+++ b/src/solver.rs
@@ -293,43 +293,6 @@ pub trait DependencyProvider {
     /// assign [`String`] as placeholder.
     type M: Eq + Clone + Debug + Display;
 
-    /// [Decision making](https://github.com/dart-lang/pub/blob/master/doc/solver.md#decision-making)
-    /// is the process of choosing the next package
-    /// and version that will be appended to the partial solution.
-    ///
-    /// Every time such a decision must be made, the resolver looks at all the potential valid
-    /// packages that have changed, and a asks the dependency provider how important each one is.
-    /// For each one it calls `prioritize` with the name of the package and the current set of
-    /// acceptable versions.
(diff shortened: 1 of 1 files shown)


src/solver.rs
```diff
@@ -293,54 +293,44 @@ pub trait DependencyProvider {
...
     ///
-    /// > Pub chooses the latest matching version of the package
-    /// > with the fewest versions that match the outstanding constraint.
-    /// > This tends to find conflicts earlier if any exist,
-    /// > since these packages will run out of versions to try more quickly.
-    /// > But there's likely room for improvement in these heuristics.
+    /// Decisions are always made for the highest priority package first. The order of decisions
+    /// determines which solution is chosen and can drastically change the performances of the
+    /// solver. If there is a conflict between two package versions, the package with the higher
+    /// priority is preserved and the lower priority gets discarded. Usually, you want to decide
```
Eh2406: > If there is a conflict between two package versions, the priority is preserved and the lower priority gets discarded.

I'm not following. What scenario are you trying to describe?

konstin: Added the word "decision".

Say:
choose A
decide A 2
choose B 
decide B 2

and then we discover A 2 -> C 1, B 2 -> C 2.

In this case, we will next try B 1, not A 1: The version of A 2 gets preserved, while B 2 gets discarded.

Eh2406: Perhaps: "If there is a conflict between two package versions, decisions will be backtracked until the lower priority package version is discarded preserving the higher priority package." or something.

Eh2406 (approved):
Looks good to me, but @mpizenberg is our master wordsmith.

mpizenberg (approved):
It’s fine by me

Issue #239, opened 2024-07-15
Remove `PubGrubError::Failure`


konstin:
`PubGrubError` currently has a `PubGrubError::Failure(String)` variant with two usages: "a package was chosen but we don't have a term." and "choose_package_version picked an incompatible version". Both of these only occur when breaking the dependency provider contract. We should either replace them by panics, or replace the `String` with two dedicated variants.

mpizenberg:
`choose_package_version` is also an old name based on the v0.2 API.

I don’t have strong opinions on this. I think having a failure type instead of panics can provide better error messages since it can be unwind at the most adequate level to provide useful information as to why and in which context that error occurred.

Eh2406:
Looking more closely at both examples, they should clearly be panics. For the first one it's not even clear how this could be a user's fault, something in the algorithm has gone very wrong if we end up in the situation. For the second one, the user has clearly messed up, but it's not hard to  provide all of the context namely what package was requested with what VS and what V was returned.

On the other hand,  this provided us a way to add new errors without making a breaking change. If that something we care about we should mark this enum as "nonexhaustive".

konstin:
As a user, i'd like to handle pubgrub errors exhaustively and if required match on or add context to `DependencyProvider::Err`, so i'd new variants to be breaking changes.

On a related note, in uv we removed `PubGrubError::SelfDependency` since in python sometimes packages depend on themselves since pip allows it, it's just a no-op.

Eh2406:
Seems reasonable. That all sounds good.

Eh2406:
Leaving open for the ErrorChoosingPackageVersion vs choose_version inconsistency.

PR #313, opened 2025-01-27
Track outdated priorities in a set

konstin:
pubgrub currently assumes that package priorities can only change when the derivations of a package change. In tracks the decision level range where derivations changed, and only updates decisions in this level.

This assumption is incorrect for uv, where the priority doesn't depend on the range, but on specifiers and discovery order per package name, not per `DP::P`, which contains virtual packages, too. This change enables more flexible priority updating for uv now and for pubgrub generally in the future.

Instead of updating all package priorities in a decision level range, we directly track the set of packages that need to be updated because we changed their derivations. This enables adding pushing priority updates from uv in our fork.

Currently, packages that are removed from prioritization, those that need to be changed and those that are added are all treated the same way, there might be some optimization by telling them apart.

This branch is based on https://github.com/pubgrub-rs/pubgrub/compare/dev...Eh2406:pubgrub:stop-prioritize.

Half of https://github.com/pubgrub-rs/pubgrub/issues/239, incidentally

### DIFF
--- a/src/internal/partial_solution.rs
+++ b/src/internal/partial_solution.rs
@@ -37,9 +37,9 @@ pub(crate) struct PartialSolution<DP: DependencyProvider> {
     /// Store for all known package decisions and package derivations.
     ///
     /// "assignment" refers to both packages with decisions and package with only derivations and
-    /// no decision yet. We combine this in a single index map, where different sections (of
-    /// indexes) contain package with different level of information, and make a decision moves a
-    /// package from the derivations sections to the decisions section.
+    /// no decision yet. We combine this in a single index map with a decisions section in front and
+    /// a derivations sections in the back, where making a decision moves a package from the
+    /// derivations sections to the decisions section.
     ///
     /// `[..current_decision_level]`: Packages that have had a decision made, sorted by the
     /// `decision_level`. The section is can be seen as the partial solution, it contains a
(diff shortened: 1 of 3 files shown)


konstin:
@Eh2406 can you run the cargo perf numbers on this? on uv it's perf neutral with https://github.com/astral-sh/uv/pull/10935

src/internal/partial_solution.rs
```diff
@@ -408,12 +406,15 @@ impl<DP: DependencyProvider> PartialSolution<DP> {
...
+                if pa.assignments_intersection.term().is_positive() {
+                    self.outdated_priorities.insert(*p);
+                }
                 true
             }
         });
-        // Throw away all stored priority levels, And mark that they all need to be recomputed.
-        self.prioritized_potential_packages.clear();
-        self.prioritize_decision_level = self.current_decision_level.0.saturating_sub(1) as usize;
+        // Throw away all stored priority levels and mark them for recomputing
```
Eh2406: the code this comment refers to has been removed.

Eh2406:
I just checked, no measurable difference on CPU time or memory usage for crates.io.

So the question is is the code going to be easier to maintain this way?

src/internal/partial_solution.rs
```diff
@@ -62,8 +63,6 @@ pub(crate) struct PartialSolution<DP: DependencyProvider> {
     /// range.
```
Eh2406: The comet ending on this line is now out of date.

konstin:
> So the question is is the code going to be easier to maintain this way?

I find the set-based approach significantly easier to understand.

Eh2406 (approved):

mpizenberg:
> I find the set-based approach significantly easier to understand.

Sounds logical to me too.

konstin:
It also looks faster in uv in a quick test (p = 1.18e-08):

```
$ hyperfine --warmup 2 --export-json a.json "./uv-main pip compile --no-progress scripts/requirements/airflow.in --universal" "./uv-branch pip compile --no-progress scripts/requirements/airflow.in --universal" --runs 100
Benchmark 1: ./uv-main pip compile --no-progress scripts/requirements/airflow.in --universal
  Time (mean ± σ):     111.5 ms ±   5.1 ms    [User: 128.9 ms, System: 108.6 ms]
  Range (min … max):   103.6 ms … 127.4 ms    100 runs
 
Benchmark 2: ./uv-branch pip compile --no-progress scripts/requirements/airflow.in --universal
  Time (mean ± σ):     106.7 ms ±   6.3 ms    [User: 120.3 ms, System: 110.0 ms]
  Range (min … max):    97.8 ms … 134.8 ms    100 runs
 
Summary
  ./uv-branch pip compile --no-progress scripts/requirements/airflow.in --universal ran
    1.05 ± 0.08 times faster than ./uv-main pip compile --no-progress scripts/requirements/airflow.in --universal
(.venv) home ~/projects/uv ⊤  uv run ../hyperfine/scripts/welch_ttest.py a.json 
```

PR #316, opened 2025-01-27
Replace failure branch with panic

### LINKED ISSUE #239: Remove `PubGrubError::Failure`

konstin:
When the user's implementation breaks the contract about choose version, we now panic. This contract is trivial to enforce (`vs.contains(v)`), so an error variant only creates an `unreachable!()` in user code.

Fixes #239

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -42,10 +42,6 @@ pub enum PubGrubError<DP: DependencyProvider> {
     /// returned an error in the method [should_cancel](DependencyProvider::should_cancel).
     #[error("We should cancel")]
     ErrorInShouldCancel(#[source] DP::Err),
-
-    /// Something unexpected happened.
-    #[error("{0}")]
-    Failure(String),
 }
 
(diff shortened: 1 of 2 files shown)


Eh2406 (approved):

mpizenberg:
Could we add a hint about potential reasons, and potential ways to diagnostic, fix the problem?

In the past, almost every time someone had an issue with pubgrub, it was because of not strictly abiding by the requirements to implement some trait.

konstin:
The method name, the expected range and actual version are all in the panic message, is there something more to add?

mpizenberg:
Not sure. Maybe something along the lines of "This issue most likely happened because you have a custom DependencyProvider implementation. Please make sure the implementation of the choose_version function is correct, and make sure that you follow all documented requirements for custom dependency providers." With a link to the docs somewhere.

PR #317, opened 2025-01-29
Align choose version error

### LINKED ISSUE #239: Remove `PubGrubError::Failure`

konstin:
The error was based on an older version of the dependency provider trait. I improved the docs and messages a bit along the way.

Fixes #239


Eh2406 (approved):

mpizenberg:
👍

PR #319, opened 2025-01-31
move solver mod comment to resolver

### LINKED ISSUE #185: Out of date documentation for Dev
In #108 an important piece of documentation was missed. The "Package and Version traits" section [rendered](https://pubgrub-rs.github.io/pubgrub/pubgrub/#package-and-version-traits) and [code](https://github.com/pubgrub-rs/pubgrub/blob/dev/src/lib.rs#L22) is wrong and out of date. This needs to be updated to describe the `VersionSet` trait.

Eh2406:
Fix #185, Part of #308

This useful piece of text was no longer part of our published docs, because it was module documentation on a private module.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,24 +8,6 @@
 //! we should try to provide a very human-readable and clear
 //! explanation as to why that failed.
 //!
-//! # Package and Version traits
-//!
-//! All the code in this crate is manipulating packages and versions, and for this to work
-//! we defined a [Package] trait
-//! that is used as bounds on most of the exposed types and functions.
-//!
-//! Package identifiers needs to implement our [Package] trait,
-//! which is automatic if the type already implements
-//! [Clone] + [Eq] + [Hash] + [Debug] + [Display](std::fmt::Display).
-//! So things like [String] will work out of the box.
-//!
-//! TODO! This is all wrong. Need to talk about VS, not Version.
-//! 
-//! Our Version trait requires
-//! [Clone] + [Ord] + [Debug] + [Display](std::fmt::Display).
-//! For convenience, this library provides [SemanticVersion]
-//! that implements semantic versioning rules.
-//!
 //! # Basic example
 //!
 //! Let's imagine that we are building a user interface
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -20,6 +20,7 @@
 //! So things like [String] will work out of the box.
 //!
 //! TODO! This is all wrong. Need to talk about VS, not Version.
+//! 
 //! Our Version trait requires
```
Eh2406: This paragraph is out of date and needs to be rewritten. I'm open to suggestions for how to explain `VS`.

src/solver.rs
```diff
@@ -108,7 +49,67 @@ impl PackageResolutionStatistics {
...
+///
+/// And because menu <1.1.0 depends on dropdown >=1.0.0 <2.0.0
+///   which depends on intl <4.0.0, every version of menu
+///   requires intl <4.0.0.
+///
+/// So, because root depends on both menu >=1.0.0 and intl >=5.0.0,
+///   version solving failed.
+/// ```
+///
+/// TODO: there is not Version traits
```
Eh2406: This paragraph is out of date and needs to be rewritten. I'm open to suggestions for how to explain `VS`.

konstin: I'd link people to `DependencyProvider` as main interface, otherwise we have to repeat all of that information.

src/lib.rs
```diff
@@ -59,6 +42,24 @@
...
 //! ```
 //!
+//! # Package and Version flexibility
+//!
+//! The [OfflineDependencyProvider] used in that example is generic over the way package names,
+//! version requirements, and version numbers are represented.
+//!
+//! The first bound is the type of package names. It can be anything that implements our [Package] trait.
+//! The [Package] trait is automatic if the type already implements
+//! [Clone] + [Eq] + [Hash] + [Debug] + [Display](std::fmt::Display).
```
konstin, line 52: Why's the `Display` link different than the others?

konstin (approved):

src/lib.rs
Eh2406: Because the others are in the prelude, and `Display` is not.

PR #338, opened 2025-05-09
Allow multiple self-dependencies

konstin:
In uv, we don't use the `DependencyConstraints` map, but pass in the dependencies through an iterator. This means there can be duplicate dependencies in the input. This would previously make `merge_dependents` panic if a package dependent on itself twice with the same range:

```toml
[package]
name = "foo"
version = "0.1.0"
dependencies = ["foo", "foo"]
```

The fix is to ignore self-dependencies when merging dependents, given that they are always trivially true or trivially false.


mpizenberg (approved):
It makes sense. And I don’t see any obvious downside considering merge_dependents is just an optimization.

PR #340, opened 2025-06-02
Dependabot: versioning-strategy increase-if-necessary

konstin:
PubGrub is a library, so we should widen the range where possible to not force users to upgrade versions.


mpizenberg (approved):
Yeah makes sense

PR #341, opened 2025-06-02
Revert "Dependabot: versioning-strategy increase-if-necessary (#340)"

konstin:
This reverts #340, as dependabot doesn't support it:

> The property '#/updates/1/versioning-strategy' value "increase-if-necessary" did not match one of the following values: lockfile-only, auto

This imho means that we can't use dependabot with pubgrub, as the main updates with want are semver-breaking updates to our deps; lockfile updates are a secondary priority, and we shouldn't bump `Cargo.toml` for compatible updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,5 @@ updates:
 
   - package-ecosystem: "cargo"
     directory: "/"
-    versioning-strategy: "increase-if-necessary"
     schedule:
       interval: "monthly"


mpizenberg:
The main upgrades we want are probably security upgrades, when there is a CVE. Otherwise I agree with your sentiment.

mpizenberg (approved):

konstin:
We can only handle CVEs to an extend as we are a library and not a binary; A binary using uv needs to set up security scanning to handle CVEs, including (or especially) for transitive dependencies. I see our responsibility as staying on supported major versions of dependencies, so users can make these transitive updates. (Though overall I don't think I've seen any CVEs that would have affected pubgrub yet, or not that many in rust overall, in uv the only ones we prominently noticed were around handling relative paths when unpacking archives.)

PR #348, opened 2025-07-01
Update to the 2024 edition

konstin:
There's a single change for the new lifetime rules, otherwise it's just Rustfmt.


mpizenberg (approved):
Ok, mostly formatting stuff indeed. With one reference/de-reference change.

konstin:
The other change is that this bumps our MSRV, but iirc there's no one depending on that (uv has an MSRV of 2 versions prior to the last release).

PR #352, opened 2025-08-27
Add method for incompatibilities between virtual and base packages

konstin:
Add a new method to add a single custom incompatibility that requires the base package and the proxy package share the same version range.

This intended for cases where proxy packages (also known as virtual packages) are used. Without this information, pubgrub does not know that these packages have to be at the same version. In cases where the base package is already to an incompatible version, this avoids going through all versions of the proxy package. In cases where there are two incompatible proxy packages, it avoids trying versions for both of them. Both improve performance (we don't need to check all versions when there is a conflict) and error messages (report a conflict of version ranges instead of enumerating the conflicting versions).

This method requires the user to manually register proxy packages, and is only available through `DependecyProvider`. Ideally, we had a more general concept of proxy packages in pubgrub, with built-in heuristics or at least misuse-safe methods.

There's several usage patterns for this method. The basic one is upon encountering a dependency on a proxy package with a range, using this method with its base package and that range.

Companion change in uv: https://github.com/astral-sh/uv/pull/15200

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -96,6 +96,40 @@ impl<DP: DependencyProvider> State<DP> {
         self.merge_incompatibility(id);
     }
 
+    /// Add a single custom incompatibility that requires that the base package and the proxy
+    /// package share the same version range.
+    ///
+    /// This intended for cases where proxy packages (also known as virtual packages) are used.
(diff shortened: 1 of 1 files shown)


konstin:
According to https://github.com/astral-sh/uv/pull/15200, this is a small performance improvement.

PR #355, opened 2025-09-02
Use Renovate instead of Dependabot

konstin:
Dependabot always updates the lower bounds `Cargo.toml`, instead of only `Cargo.lock` for compatible updates, while updating `Cargo.toml` for breaking updates , as we want it for a library.

From https://docs.renovatebot.com/configuration-options/#rangestrategy:

> replace: Replace the range with a newer one if the new version falls outside it, and update nothing otherwise
>
> update-lockfile: Update the lock file when in-range updates are available, otherwise replace for updates out of range.

Additionally with this migration, I've pinned GitHub Actions to hashes, which unlike tags can't be changed.

The basic Renovate config is smaller than the dependabot config, it uses autodiscovery and finds Cargo and GitHub Actions. There's a dashboard with the option to create and rebase PRs and with logs.

You can see a demo of this change and the PRs renovate creates in https://github.com/konstin/pubgrub-renovate.

To actually enable this change, we need to give the Renovate Mend GitHub app access to the repository.


konstin:
This is how the dashboard looks like in my test repo

image

Eh2406 (approved):

PR #377, opened 2025-10-09
Use stable order for dependencies

### LINKED ISSUE #373: Sort order inconsistent on 32-bit archs
I'm trying to package https://github.com/tweag/nickel which depends on pubgrub. One of the [tests](https://github.com/tweag/nickel/blob/master/package/tests/integration/inputs/path/conflicting-index-deps/Nickel-pkg.ncl) uses pubgrub to verify that certain output indicating a package resolution error will occur in the case of bad input. This test [consistently fails](https://gitlab.alpinelinux.org/strophy/aports/-/pipelines/364446) under Alpine Linux 3.22 on all 32 bit architectures I'm trying to package for: x86, armv7 and armhf
```
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot file: package/tests/snapshots/integration__package__tests__integration__inputs__path__conflicting-index-deps__Nickel-pkg.ncl.snap
Snapshot: package/tests/integration/inputs/path/conflicting-index-deps/Nickel-pkg.ncl
Source: package/tests/integration.rs:266
────────────────────────────────────────────────────────────────────────────────
test package/tests/integration/inputs/path/path-to-index-dep/Nickel-pkg.ncl               ... ok
Expression: e.to_string()
────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────────
    1     1 │ package version resolution failed:
    2       │-Because 'Path <filtered> depends on github:example/leaf 0.1.0 and top-level package 0.1.0 depends on 'Path <filtered> top-level package 0.1.0 depends on github:example/leaf 0.1.0.
    3       │-And because top-level package 0.1.0 depends on github:example/leaf 0.1.1, top-level package 0.1.0 is forbidden.
          2 │+Because 'Path <filtered> depends on github:example/leaf 0.1.0 and top-level package 0.1.0 depends on github:example/leaf 0.1.1, top-level package 0.1.0, 'Path <filtered> * are incompatible.
          3 │+And because top-level package 0.1.0 depends on 'Path <filtered> top-level package 0.1.0 is forbidden.
────────────┴───────────────────────────────────────────────────────────────────
Stopped on the first failure. Run `cargo insta test` to run all snapshots.
thread '<unnamed>' panicked at /home/buildozer/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/insta-1.43.2/src/runtime.rs:694:13:
snapshot assertion for 'package/tests/integration/inputs/path/conflicting-index-deps/Nickel-pkg.ncl' failed in line 266
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test package/tests/integration/inputs/path/conflicting-index-deps/Nickel-pkg.ncl          ... FAILED
Already have a cache entry at "/tmp/.tmpxu7Ted/cache_dir/git-packages/44b1c226afdc6220892c242f6b05c9c654aa6065/"
test package/tests/integration/inputs/path/duplicate-path/Nickel-pkg.ncl                  ... ok
test package/tests/integration/inputs/path/git-path-dep/Nickel-pkg.ncl                    ... ok
failures:
---- package/tests/integration/inputs/path/conflicting-index-deps/Nickel-pkg.ncl ----
test panicked: snapshot assertion for 'package/tests/integration/inputs/path/conflicting-index-deps/Nickel-pkg.ncl' failed in line 266
failures:
    package/tests/integration/inputs/path/conflicting-index-deps/Nickel-pkg.ncl
test result: FAILED. 14 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.32s
error: test failed, to rerun pass `-p nickel-lang-package --test integration`
```

While the content of the message effectively carries the same meaning, the different sort order causes the test to fail. Is this simply a case of this not being a well-written test, or not a reasonable thing to test? Or is there something fundamentally different in pubgrub regarding how 32-bit builds sort the dependencies when trying to resolve them?

Ref: https://github.com/tweag/nickel/issues/2372

konstin:
Fixes https://github.com/pubgrub-rs/pubgrub/issues/373.

By using a `Vec` instead of an `[Fx]HashMap` for `DependencyConstraints`, the insertion order of dependencies becomes deterministic, and thereby our internal prioritization. This fixes the problem in https://github.com/pubgrub-rs/pubgrub/issues/373 where priorities, and thereby resolutions, can change depending on the platform, such as 64-bit vs 32-bit. Now, the first package in a dependency list is always prioritized.

As `DependencyConstraints` had exposed the `FxHashMap`, this is breaking change that requires pubgrub v0.4.

I made `SelectedDependencies` opaque too for good measure.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,16 @@ jobs:
 
       - run: cargo test --all-features --workspace
 
+  test-i686:
+    name: Tests (32-bit)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
+      - uses: dtolnay/rust-toolchain@stable
+      - run: rustup target add i686-unknown-linux-gnu
+      - run: cargo build --workspace --target i686-unknown-linux-gnu
+      - run: cargo test --all-features --workspace --target i686-unknown-linux-gnu
+
   clippy:
     name: Clippy
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,9 @@ jobs:
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
       - uses: dtolnay/rust-toolchain@stable
+      - run: | 
+          sudo apt-get update
+          sudo apt-get install -y gcc-multilib
       - run: rustup target add i686-unknown-linux-gnu
       - run: cargo build --workspace --target i686-unknown-linux-gnu
       - run: cargo test --all-features --workspace --target i686-unknown-linux-gnu
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## 0.4.0
+
+`DependencyConstraints` is now an opaque type that retains the ordering of the dependencies across platforms. This fixes
+unstable resolutions across platform (https://github.com/pubgrub-rs/pubgrub/issues/373). `DependencyConstraints` can be
+constructed from an iterator. It is currently backed by a `Vec` internally.
+
 ## [0.3.0] - 2025-02-12 - [(diff with 0.2.1)][0.2.1-diff]
 
 PubGrub 0.3 has a more flexible interface and speeds resolution significantly. The public API is very different now, we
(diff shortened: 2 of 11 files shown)


mpizenberg:
I’d really love to have @Eh2406 point of view on this, especially performance-wise. Because all data structure changes in the lib have potential to significantly degrades performances.

Eh2406:
Overall having control over our types is a huge improvement.

For `DependencyConstraints` Forcing an allocation at all seems a little wasteful. I wish we could take `-> impl Iter`. The type shenanigans required to pull that off are quite extraordinary, and probably not worth it. But if we have to do an allocation, `Vec` is not worse that `Map`. (Although somewhat academically I do wonder why the benchmarks disagree with me.)

konstin:
I haven't checked, but I assume the sudoku perf regression is due to the changed traversal order, we call get dependencies a max of `9*9*9` and realistically much less, I don't see how this could have that much of an impact.

PR #389, opened 2025-11-10
ci: address security findings

woodruffw:
Hi there! This addresses a handful of low-impact findings from [zizmor](https://docs.zizmor.sh).

Key changes:

- Dropped the default permissions on all workflows to `{}` (i.e. no permissions), and have only added explicit permissions where actually needed (at the moment, just the docs workflow).
- Disabled `actions/checkout`'s implicit credential persistence everywhere.
- In a separate commit, I've removed/replaced the various usages of (unpinnable) `rust-toolchain` action with either the runner's own stable rust toolchain (which is already installed) or an explicit `rustup` invocation for a nightly toolchain where necessary.

Of these changes, I suspect the `rust-toolchain` one will merit the most attention 🙂 -- per https://github.com/dtolnay/rust-toolchain/issues/160 there's currently no great way to pin that action in a way that ensures reproducibility/hermetic CI runs, so I've opted to remove it instead. However, I've put that in a separate commit and I'm happy to remove it from the proposed changes here if you consider the unpinned action here an acceptable tradeoff!

Signed-off-by: William Woodruff <william@astral.sh>

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -7,12 +7,17 @@ on:
   pull_request:
   workflow_dispatch:
 
+permissions: {}
+
 jobs:
   benchmarks:
     name: Run benchmarks
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
+        with:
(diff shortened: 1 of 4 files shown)


konstin (approved):
Thanks!

PR #404, opened 2026-02-09
Readme refresher

konstin:
Update the readme to be more approachable and also include a reference to uv as production reference.

The readme is more low-level, as it targets people who may want to start using pubgrub.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,47 @@
-# PubGrub version solving algorithm
+# PubGrub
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,17 +1,31 @@
-# PubGrub version solving algorithm
+# PubGrub
 
-![license](https://img.shields.io/crates/l/pubgrub.svg)
```
konstin, line 3: GitHub shows the license in its UI now.

Eh2406 (approved):

PR #423, opened 2026-04-22
Allow creating DependencyConstraints from Vec

andrius-puksta-sensmetry:
Avoids reallocation if caller already has a Vec.

### DIFF
--- a/src/solver.rs
+++ b/src/solver.rs
@@ -324,6 +324,12 @@ impl<P, VS> Default for DependencyConstraints<P, VS> {
     }
 }
 
+impl<P, VS> From<Vec<(P, VS)>> for DependencyConstraints<P, VS> {
+    fn from(value: Vec<(P, VS)>) -> Self {
+        Self(value)
+    }
+}
+
 impl<P, VS> FromIterator<(P, VS)> for DependencyConstraints<P, VS> {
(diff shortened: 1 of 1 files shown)


konstin:
Can you say some more about the motivation for this change?

andrius-puksta-sensmetry:
I just upgraded a codebase to `pubgrub` 0.4. In 0.3, `DependencyConstraints` was a type alias for `pubgrup::Map`, which could be populated in any manner the user desires. The type in 0.4 requires supplying all the info in one go, and if for whatever reason the info has to be populated incrementally, the best way currently is:
1. `Vec::new`
2. (repeat) `vec.push`
3. `DependencyConstraints::from_iter(vec)`

Step 3 internally creates a new `Vec`, but we already have a vec with all the needed elements, so I see no reason to not allow constructing from it directly. Alternatively, `DependencyConstraints` could expose methods to populate the internal Vec incrementally.

konstin:
rustc explicitly optimizes this case, so using `Vec::from_iter(vec)` is optimized to a noop: https://github.com/rust-lang/rust/blob/main/library/alloc/src/vec/in_place_collect.rs. We don't need to add a special case for this, the compiler already avoids the reallocation for us.

PR #424, opened 2026-04-28
Publish from CI

konstin:
We've been successfully publishing from CI in the astral-sh fork, so I'm porting this change to the pubgrub repo. The main change is that the tags need to be project prefixed, `pubgrub-v1.2.3` instead of `v1.2.3`.

This means that we can remove `release.md` and use the standard process instead.

It's also more secure by using trusted publishing instead of API tokens.


mpizenberg:
Thanks, yeah that release.md was really an antique thing, and a CI-driven process with OIDC is much better.

One remark, a few weeks ago, there was a supply chain attack, I think tanstack? I don’t remember the specifics, but I remember they had trusted publishing but it didn’t save them. Have you read about that? Do we know if we are protected against a similar attack?

mpizenberg:
Found the postmortem, it was a cache poisoning thing https://tanstack.com/blog/npm-supply-chain-compromise-postmortem

konstin:
The problem for tanstack was using `pull_request_target`, which we do not use; `pull_request_target` is a big security hazard in github actions.

We could add zizmor to be stricter about CI, but that sounds like a bit of an overkill.

The other big hazard is updating dependencies, when a dependency was just taken over. I've filed https://github.com/pubgrub-rs/pubgrub/pull/427 to catch this, as most malicious takeovers are noticed in the first 24h after publish. We're generally protected by the locked file using locked versions. The only exception is the minimal-versions check, but that doesn't run in the release environment so it can't publish a new release, and the release action doesn't use any cache, i.e. that check is sufficiently isolated even when taken over. (Plus the trusted publishing prevents us from stealing any token even if the release environment got compromised, you'd only get that very specific publish)

PR #425, opened 2026-06-07
Use an arena for the derivation tree

### LINKED ISSUE #293: Stack overflow when displaying the derivation tree
When importing in pubgrub the cargo test from https://github.com/rust-lang/cargo/blob/bf79c8b154fa6b3e4d781c26935429948cbd9b2b/crates/resolver-tests/tests/resolve.rs#L529, I get a stack overflow in the test when displaying the derivation tree:

```rust
#[test]
fn resolving_with_constrained_cousins_backtrack() {
    const DEPTH: u32 = 100;
    const BRANCHING_FACTOR: u32 = 50;

    let mut dependency_provider =
        OfflineDependencyProvider::<String, Ranges<SemanticVersion>>::new();

    let root_deps = [
        ("backtrack_trap_v0_0".into(), Ranges::full()),
        ("cloaking".into(), Ranges::full()),
    ];
    dependency_provider.add_dependencies("root".into(), (0, 0, 0), root_deps);

    dependency_provider.add_dependencies(
        "cloaking".into(),
        (0, 0, 0),
        [("constrained".into(), Ranges::between((1, 0, 0), (1, 1, 0)))],
    );

    for v in 1..BRANCHING_FACTOR + 10 {
        dependency_provider.add_dependencies("constrained".into(), (1, 0, v), []);
    }
    dependency_provider.add_dependencies("constrained".into(), (1, 1, 0), []);

    for n in 0..DEPTH {
        for v in 1..BRANCHING_FACTOR {
            let v = (1, 0, v);
            dependency_provider.add_dependencies(
                format!("backtrack_trap_v0_{n}"),
                v,
                [(format!("backtrack_trap_v1_{n}"), Ranges::singleton(v))],
            );
            dependency_provider.add_dependencies(
                format!("backtrack_trap_v1_{n}"),
                v,
                [(format!("backtrack_trap_v2_{n}"), Ranges::singleton(v))],
            );
            dependency_provider.add_dependencies(
                format!("backtrack_trap_v2_{n}"),
                v,
                [(format!("backtrack_trap_v0_{}", n + 1), Ranges::full())],
            );
        }
    }
    for v in 1..BRANCHING_FACTOR {
        let v = (1, 0, v);
        dependency_provider.add_dependencies(
            format!("backtrack_trap_v0_{DEPTH}"),
            v,
            [("constrained".into(), Ranges::between((1, 1, 0), (2, 0, 0)))],
        );
    }

    let res = pubgrub::resolve(&dependency_provider, "root".into(), (0, 0, 0));

    if let Err(PubGrubError::NoSolution(error)) = res {
        // Stack overflow here
        let _out = format!("{error:?}");
        let _out = DefaultStringReporter::report(&error).to_string();
    };
}
```

konstin:
We hit a real world case of https://github.com/pubgrub-rs/pubgrub/issues/293 in uv, where uv would stack overflow trying to display the error for a popular package with a lot of versions: https://github.com/astral-sh/uv/issues/19672.

The idea is to treat a derivation tree similar to the general dependency tree by storing nodes in an arena, instead of as `Arc`s, and by using a stack instead of recursion for displaying errors.

This is a breaking change, as it changes the public derivation tree API.

Fixes https://github.com/pubgrub-rs/pubgrub/issues/293

### DIFF
--- a/src/internal/core.rs
+++ b/src/internal/core.rs
@@ -4,7 +4,6 @@
 //! to write a functional PubGrub algorithm.
 
 use std::collections::HashSet as Set;
-use std::sync::Arc;
 
 use crate::internal::{
     Arena, DecisionLevel, HashArena, Id, IncompDpId, IncompId, Incompatibility, PartialSolution,
@@ -409,17 +408,19 @@ impl<DP: DependencyProvider> State<DP> {
         let mut sorted_ids = all_ids.into_iter().collect::<Vec<_>>();
         sorted_ids.sort_unstable_by_key(|id| id.into_raw());
         let mut precomputed = Map::default();
+        let mut arena = Vec::with_capacity(sorted_ids.len());
         for id in sorted_ids {
-            let tree = Incompatibility::build_derivation_tree(
+            let node = Incompatibility::build_derivation_tree_node(
                 id,
                 &shared_ids,
                 &self.incompatibility_store,
                 &self.package_store,
                 &precomputed,
             );
-            precomputed.insert(id, Arc::new(tree));
+            let node_id = DerivationTree::alloc_node(&mut arena, node);
+            precomputed.insert(id, node_id);
         }
         // Now the user can refer to the entire tree from its root.
-        Arc::into_inner(precomputed.remove(&incompat).unwrap()).unwrap()
+        DerivationTree::from_arena(arena, precomputed.remove(&incompat).unwrap())
     }
 }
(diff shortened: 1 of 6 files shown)


mpizenberg:
Improving the error-reporting side of pubgrub was always something we wanted to do, but was never prioritized. And I agree it most likely needs some breaking changes. Personally I don’t have the bandwidth to review, but for those that may have the time, would you mind highlighting the api changes?

konstin:
The API here isn't fully done yet, this is rather half-baked atm.

Eh2406:
😋 🍪 . I would like to do a review of this before it gets merged.

PR #426, opened 2026-06-07
Do not use nightly in minimal-versions check

konstin:
Old versions sometimes stop working on recent nightly, which broke the check on main. Using stable rust avoids this problem in the future.

Also bump codspeed for the associated cc bump, to work around https://github.com/CodSpeedHQ/codspeed-rust/pull/182.


mpizenberg:
Being able to run without nightly is much better, agreed.

PR #428, opened 2026-06-23
ci: raise is-terminal and tempfile dev floors so minimal-versions builds on nightly

RyanJamesStewart:
The `Minimal versions` CI job (`cargo +nightly update -Zminimal-versions` then build and test) is latently broken. Under minimal versions, two dev-dep paths resolve `rustix 0.37`: `codspeed-criterion-compat -> is-terminal 0.4.6` and `proptest -> tempfile`. rustix below 0.38.6 relies on the `rustc_layout_scalar_valid_range_start` attribute, which the compiler removed in rust-lang/rust#155433, so it no longer compiles on current nightly (tracked in bytecodealliance/rustix#1620). The job only still passes because it has not re-run on a post-removal nightly.

This raises two dev-dep floors so minimal versions resolves a buildable rustix. `is-terminal = "0.4.8"` moves the codspeed path off the rustix 0.37 line, and `tempfile = "3.9"` raises the proptest path to rustix 0.38.26 (the lowest tempfile floor whose rustix builds; `tempfile 3.7` and `3.8` still resolve the unbuildable rustix 0.38.0). No direct rustix floor is added, so the normal latest-versions resolution is unchanged and keeps a single rustix.

Verified locally on nightly 2026-05-12:

- with `-Zminimal-versions`, rustix resolves to a single 0.38.26 and `cargo build --workspace` plus `cargo test --all-features --workspace` pass;
- the latest-versions build is unaffected (single rustix 1.1.4, tests pass).

The astral-sh/pubgrub fork hit the same breakage; the fix there (astral-sh/pubgrub#55) was redirected here so the fork can rebase on top.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,15 @@ criterion = { version = "4.7.0", package = "codspeed-criterion-compat" }
 env_logger = "0.11.6"
 # Lower-bound guard for `codspeed-criterion-compat` in the minimal-versions CI job.
 cc = "1.2.40"
+# Lower-bound guards to keep the transitive `rustix` buildable in the
+# minimal-versions CI job. rustix below 0.38.6 uses the
+# `rustc_layout_scalar_valid_range_start` attribute removed from the compiler in
+# rust-lang/rust#155433, so it no longer builds on nightly (see
+# bytecodealliance/rustix#1620). `codspeed-criterion-compat -> is-terminal` pulls
+# rustix 0.37 until is-terminal 0.4.8, and `proptest -> tempfile` resolves rustix
+# 0.38.0 until tempfile 3.9 raises its floor to rustix 0.38.26.
+is-terminal = "0.4.8"
+tempfile = "3.9"
 proptest = "1.6.0"
 ron = "0.12.0"
 varisat = "0.2.2"


konstin:
Can you please share the exact commands that reproduce the breakage? CI on main is clean, and it doesn't reproduce for me locally.

RyanJamesStewart:
You're right that it doesn't reproduce on `dev`, and I tracked down why. The minimal-versions job runs the `update` step with `+nightly` but builds and tests on stable (`cargo build` / `cargo test`), and that split is what keeps it green.

Under `-Zminimal-versions`, `dev` resolves a single `rustix 0.37.0` (pulled by `is-terminal 0.4.6`; `tempfile` minimizes to `3.0.2`, which predates rustix, so it adds no rustix path here). rustix 0.37 uses the `rustc_layout_scalar_valid_range_start` attribute that rust-lang/rust#155433 removed, but only when it is compiled by a nightly rustc. On stable it compiles fine, so the job passes.

Exact commands on a clean `dev` checkout (nightly 2026-05-12 or later):

```
cargo +nightly update -Zminimal-versions

# the CI steps, on stable: both pass
cargo build --workspace
cargo test --all-features --workspace

# compile that same dev-dep on nightly: fails
cargo +nightly test --all-features --workspace
```

The last command gives:

```
error: cannot find attribute `rustc_layout_scalar_valid_range_start` in this scope
error: could not compile `rustix` (lib) due to 4 previous errors
```

So the breakage is nightly-only, and the pubgrub-rs job never compiles a dev-dependency on nightly, which is why CI on main is clean. Where I actually hit it was the astral-sh fork, whose minimal-versions job uses `cargo +nightly build` / `cargo +nightly test`, so it builds rustix on nightly and fails (fixed on the fork in astral-sh/pubgrub#55).

Given that, this PR is not needed for `dev` as the job stands today, and I am happy to close it. The one alternative worth flagging: if you would want the minimal-versions job to also catch this class of nightly-only dev-dep rot, switching its build and test steps to `+nightly` would surface it, and then these floor bumps would be the fix that keeps it green. Your call on whether that is worth it; otherwise I will close.

konstin:
As nightly is unstable, we shouldn't add fixes specifically for nightly and minimal versions.

PR #432, opened 2026-08-07
Avoid materializing complements in term algebra

konstin:
More performance gains from nab, to subtract one version range from another, the solver currently builds the complement of the second range and intersects with it, allocating an intermediate range that is immediately thrown away.

This adds `Ranges::difference`, which computes the subtraction directly.

| bench | instructions |
|---|---|
| backtracking_disjoint_versions | -13.8% |
| backtracking_ranges | -4.7% |
| large_case | -1.4% |
| backtracking_singletons | -0.8% |
| no_backtracking, sudoku | no change |

This is an upstream port of https://github.com/astral-sh/pubgrub/pull/77

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## Unreleased
+
+### Added
+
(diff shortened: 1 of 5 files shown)


notatallshaw:
Thanks for upstreaming my commit faithfully, though I would have preferred to raise this myself.

konstin:
I will up to you the next time!

PR #433, opened 2026-08-16
Use `contains_many` to batch offline version membership checks

notatallshaw:
I have my own implementation of pubgrub in Python, [nab](https://github.com/notatallshaw/nab/tree/main/nab-resolver), and I often check pubgrub-rs optimizations to port over to nab, this is the other way around, this is inspired by an optimization I have in nab that can be applied to pubgrub-rs.

`OfflineDependencyProvider::prioritize` checks each of a package's versions against the range one at a time, but the versions come out in order, so we can use `Ranges::contains_many` to walk them in one pass.

Performance improvements I see locally:

| bench | instructions |
|---|---|
| backtracking_singletons | -6.1% |
| backtracking_disjoint_versions | -5.1% |
| large_case, all 641 roots | -1.0% |
| sudoku, hard board | -0.8% |
| sudoku, easy board | -0.6% |
| backtracking_ranges | -0.4% |

P.S. @konstin this doesn't help uv, because uv bypasses this API by exposing cursor methods directly on the range. I assume there is a history to why such API was never upstreamed to pubgrub-rs.


konstin (approved):
Thank you!

PR #436, opened 2026-09-11
Use a niche-optimized representation for arena IDs

konstin:
## What

`Id<T>` now stores its zero-based arena index as `index + 1` in a `NonZeroU32`. This gives `Option<Id<T>>` a zero niche, keeping it the same four-byte size as `Id<T>`. A compile-time assertion pins that layout.

Arena allocation now converts `usize` indices through a checked constructor instead of truncating them with `as u32`. The largest representable index is `u32::MAX - 1`; attempts to allocate beyond it panic before mutating a normal arena. Debug output and hashing continue to use the logical zero-based index, so the representation change is otherwise invisible.

## Why

The partial solution stores an `Option<IncompId>` for each satisfied package. Removing the separate `Option` discriminant shrinks that value tuple from 16 to 12 bytes and its common inline `(Id, value)` entry from 20 to 16 bytes. It also turns a previously silent index truncation into an explicit overflow failure.

Upstreamed from [https://github.com/astral-sh/pubgrub/pull/58](https://github.com/astral-sh/pubgrub/pull/58).


mpizenberg:
Nice trick!

PR #443, opened 2026-09-11
Fuse version-set relation checks

konstin:
## Summary

PubGrub classifies term relationships by first checking whether the accumulated term is a subset of the incompatibility term and, when that fails, separately checking whether the terms are disjoint. For `Ranges`, those operations each traverse the segment lists, so relationships that need both answers can walk the same inputs twice during unit propagation.

This adds `SetRelation` and a default `VersionSet::relation` implementation, then gives `Ranges` a single traversal that reports subset, disjoint, or overlapping relationships. `Term::relation_with` uses the fused query for the two polarity combinations that need both subset and disjointness, while keeping subset-only combinations on `subset_of`. An empty positive intersection is handled separately to preserve the existing satisfaction-before-contradiction precedence, and custom version-set implementations retain the existing two-operation behavior through the default method.

Because `pubgrub` now re-exports and calls the new API, this also bumps `version-ranges` to 0.1.4 and raises PubGrub's dependency lower bound to that release. `version-ranges` must be published before the next `pubgrub` release.

## Performance

In the original Astral PR, alternating same-process A/B measurements against `0edce90`, followed by post-review checks against Astral `main`, produced:

| PubGrub benchmark | Result |
| --- | ---: |
| `backtracking_singletons` | 13.2%–17.1% faster across three runs |
| `backtracking_disjoint_versions` | no consistent change beyond local noise |
| `backtracking_ranges` | no consistent change beyond local noise |

The full all-features workspace test suite, relation property checks, focused regression tests, Clippy with warnings denied, formatting, and diff checks pass.

Upstreamed from [https://github.com/astral-sh/pubgrub/pull/66](https://github.com/astral-sh/pubgrub/pull/66).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1160,7 +1160,7 @@ checksum = "68ed610a8d5e63d9c0e31300e8fdb55104c5f21e422743a9dc74848fa8317fd2"
 
 [[package]]
 name = "version-ranges"
-version = "0.1.3"
+version = "0.1.4"
 dependencies = [
  "proptest",
  "ron",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ priority-queue = "2.3.1"
 rustc-hash = "^2.1.1"
 serde = { version = "1.0.219", features = ["derive"], optional = true }
 thiserror = "2.0"
-version-ranges = { version = "0.1.0", path = "version-ranges" }
+version-ranges = { version = "0.1.4", path = "version-ranges" }
 
 [dev-dependencies]
 criterion = { version = "4.7.0", package = "codspeed-criterion-compat" }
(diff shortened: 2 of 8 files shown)


Eh2406:
I'm still very confused that there are inputs that do a significant amount of work to check both branches. Something that is subset is pretty clearly not disjoint and vice versa. It is easy to construct inputs where subset takes len time or where disjoint does, but the first value checked in the other loop should lead to termination. The existing implementation has worst case complexity of `N+1` not `2N`. So 🤷 why the performance results are so good. Not that I'm arguing with them.

konstin:
It seems that most terms/ranges are overwhelmingly disjoint and overwhelmingly have a single segment (all in positive terms). With the previous version, we have to call `valid_segment` twice, once for both `subset_of` and once for `is_disjoint`, while the fused version can do it in a single operation. For `backtracking_singletons`, we go from 37 mio to 25 mio `valid_segment` checks. We're almost never running into O(n) cases, due to n=1 for most benchmarked cases, so we get our speedup from the single range case.

